PR #27, opened 2019-07-19
Lock iqsharp-base version of Debian to Stretch.

cgranade:
This PR fixes the version of Debian used in the FROM line for the iqsharp-base image to stretch, rather than allowing iqsharp-base to upgrade implicitly.

### DIFF
--- a/images/iqsharp-base/Dockerfile
+++ b/images/iqsharp-base/Dockerfile
@@ -1,5 +1,5 @@
 # The same image used by mybinder.org
-FROM python:3.7-slim
+FROM python:3.7-slim-stretch
 
 # install the notebook package
 RUN pip install --no-cache --upgrade pip && \


images/iqsharp-base/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # The same image used by mybinder.org
-FROM python:3.7-slim
+FROM python:3.7-slim-stretch
```
anpaz-msft: do we want to a dynamic label, or should we attach to a specific version?

cgranade: From my understanding, this is the almost the most specific tag offered by the Python image repo. We could possibly use 3.7.4-slim-stretch instead, though, to lock down the patch version of Python as well.

anpaz-msft: then, it lgtm

PR #33, opened 2019-08-22
Updates IQ# to use the prerelease feed.

cgranade:
Mirroring microsoft/qsharp-runtime#39 and microsoft/quantumlibraries#151, this PR updates IQ# to use the Quantum Development Kit pre-release feed and updates the CI-built version to use 0.9 as a prefix. This allows, for example, depending on recent Microsoft.Jupyter.Core functionality such as microsoft/jupyter-core#34 to diagnose Jupyter protocol issues.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+<!--
+    This configuration tells the .NET Core SDK to build IQ# using the
+    Quantum Development Kit prerelease feed in addition to any other currently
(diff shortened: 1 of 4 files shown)


NuGet.Config
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+<!--
+    This configuration tells the .NET Core SDK to build IQ# using the
```
anpaz-msft: Technically, it just tells .net core to also look in the `qdk-alpha` feed, not to build against using that feed. Each project then specifies which version of the QDK to build against.

cgranade:
Thanks for the feedback, I've modified accordingly.

PR #37, opened 2019-09-11
Build conda packages for IQ#, qsharp.py

cgranade:
This PR prepares for distributing IQ# and the `qsharp` Python as conda packages, making it easier to install native dependencies at the same time.


build/ci.yml
```diff
@@ -9,8 +9,36 @@ variables:
 
 jobs:
 - job: "iqsharp"
+  pool:
+    vmImage: ubuntu-16.04
   steps:
   - template: steps.yml
+    parameters:
+      onlyConda: "false"
+  - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
```
anpaz-msft: I don't think we need to run these more than once.

build/steps.yml
```diff
@@ -1,35 +1,37 @@
 ##
 # Build and test IQ#.
 ##
+parameters:
```
anpaz-msft: It feels we need different steps file for conda. They can all publish to the same artifact.

conda-recipes/dotnetcore-sdk/bld.bat
```diff
@@ -0,0 +1,64 @@
+set DOTNET_ROOT=%PREFIX%\opt\dotnet
```
anpaz-msft: Why do you need to explicitly install .netcore during the conda build? Can't you assume is already globally installed?

conda-recipes/dotnetcore-sdk/build.sh
```diff
@@ -0,0 +1,41 @@
+DOTNET_ROOT=$PREFIX/opt/dotnet
```
anpaz-msft: even if we do need to install .net core, wouldn't it be better to have a single script?
this is only for someone building the recipe in their machine, right? the conda package itself, during installation, will not do these things, correct? if so, we can definitely be more relax about what we can expect the developer trying to build the conda recipe has installed

conda-recipes/iqsharp/ToolStandalone.csproj
```diff
@@ -0,0 +1,45 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
anpaz-msft: why do we need a different project? can you just target the existing one?

conda-recipes/iqsharp/build.ps1
```diff
@@ -0,0 +1,77 @@
...
+    "Path to .NET Core SDK" = $DotNetPath;
+    "Script root" = $PSScriptRoot;
+    "Prefix" = $Env:PREFIX;
+    "Runtime ID" = $RuntimeID;
+    "Target directory" = $TargetDirectory;
+    "Path to Jupyter" = (Get-Command jupyter -ErrorAction SilentlyContinue).Source;
+    "Repo root" = $RepoRoot;
+} | Format-Table | Out-String | Write-Host
+
+# We need to disable <PackAsTool>true</PackAsTool> when publishing
```
anpaz-msft: Oh, I see. Can you just pass /property:PackAsTool=false from the command line, instead of having two versions of the .csproj?

conda-recipes/pwsh/build.sh
```diff
@@ -0,0 +1,6 @@
+# The user may not have run .NET Core SDK before, so we disable first-time
+# experience to avoid capturing the NuGet cache.
+export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+export NUGET_XMLDOC_MODE=skip
+
+dotnet tool install PowerShell --tool-path $PREFIX/bin --version $PKG_VERSION
```
anpaz-msft: similar, I think we can assume pwsh is installed in the build machine.

build/steps.yml
cgranade: Sounds good, happy to remove that from pack.ps1, then. That's definitely a lot simpler than trying to make pack.ps1 do two different things.

conda-recipes/iqsharp/ToolStandalone.csproj
cgranade: `dotnet publish` has a lot of bugs when running on a project that has `<PackAsTool>true</PackAsTool>`. This was the most reliable workaround that I could find for those issues.

conda-recipes/iqsharp/build.ps1
cgranade: I tried that, and hit a bug with the publish step, such that this turned out to be more reliable. I hope to be able to remove this kludge with future .NET Core SDK versions, though.

conda-recipes/dotnetcore-sdk/bld.bat
cgranade: This keeps with conda packaging principles that the recipe should be independent of the local environment used to build a package. As discussed offline a little while ago, this approach allows us to control precisely what version of the .NET Core SDK is used as a build dependency, with zero runtime impact.

anpaz-msft: But we have full control of the build machine. Adding (and maintaining) this, seems a lot more work than is needed.

cgranade:
> I'm missing something: where do you specify conda that during installation you need to register the Q# kernel?

That happens at [Line 79 of `iqsharp/build.ps1`](https://github.com/microsoft/iqsharp/pull/37/files#diff-3b83d6e4544183c701b5a3e3742fa1c6R79). The configuration files created by that command are then included in the conda package, and installed along with the IQ# binaries themselves.

conda-recipes/iqsharp/build.ps1
```diff
@@ -0,0 +1,87 @@
...
+Pop-Location
+
+Write-Host "## Installing IQ# into Jupyter. ##"
+$BaseName = "Microsoft.Quantum.IQSharp";
+if ($IsWindows) {
+    $BaseName = "${BaseName}.exe";
+}
+Push-Location $TargetDirectory
+    $PathToTool = Resolve-Path "./$BaseName";
+    & $PathToTool install --path-to-tool $PathToTool --sys-prefix
```
anpaz-msft: @cgranade, how does conda know that it needs to pick the kernel.json and add it to the kernels folder in the machine being installed.
also, couldn't the $PathToTool be different in the build machine vs where it actually gets installed?
I think it would be better to run this command on the actual machine during installation. Not sure if that's possible with Conda.

cgranade:
@anpaz-msft @bromeg: Based on offline discussion, it sounds like we have an action plan for moving forward. Namely, we'll drop the dotnetcore-sdk and pwsh packages, making sure to document that any contributors (not users!) who want to build the `iqsharp` and `qsharp` packages locally need both of those requirements installed out-of-band. This will mean that:

- A Python user who wants to use IQ# will _only_ need the `qsharp` conda package to run Q# from a Python host, and all other requirements will be brought in as dependencies automatically.
- A Python user who wants to develop Q# code from within Visual Studio Code will need to install the .NET Core SDK out-of-band to use intelliSense feature, and will automatically be prompted to do so by the Visual Studio Code extension.

PR #43, opened 2019-10-11
Updating to .NET Core 3.0

bettinaheim:
We first need to update the libraries packages to pick up the Range -> QRange change and then the build should succeed.


bettinaheim:
@cgranade  From taking a brief look, it looks like we need updated libraries packages first; iQ# is preloading some of the libraries, and currently that causes a runtime exception since it tries to load libraries that use the old Range type instead of QRange.

PR #44, opened 2019-10-12
First stab at Ready2Run conda package.

cgranade:


cgranade:
Note, this PR will likely fail until microsoft/jupyter-core#37 is completed and until this kernel is updated to use the newest Jupyter.Core version, as KernelApplication currently calls an ASP.NET Core method that was removed with 3.0.

cgranade:
With microsoft/jupyter-core#37 merged in, we should hopefully have everything we need for this build to work.

cgranade:
NB: The pipeline for this PR has not yet been updated to enable building conda packages by default, so the status check is not yet accurate.

cgranade:
This should now be ready, and the CI pipeline should be correctly reporting its status. @bettinaheim and @anpaz-msft, could you please take another look? Thanks! 💕

build/pack.ps1
```diff
@@ -79,12 +79,40 @@ function Pack-Image() {
...
+        [string] $Configuration = $Env:BUILD_CONFIGURATION
+    );
+
+    $OutputPath = Join-Path $Env:BLOBS_OUTDIR $Runtime;
+
+    # Suppress generating pdb files.
+    # See https://github.com/dotnet/cli/issues/2246#issuecomment-320633639.
+    dotnet publish `
+    (Join-Path $PSScriptRoot $Project) `
+        -c $Configuration `
```
anpaz-msft: I'm pretty sure you're missing passing the assembly version.
I think you will also need to pass assembly constants for signed/telemetry.

anpaz-msft: these are all the flags we normally use:
```
        -v $Env:BUILD_VERBOSITY `
        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
        /property:Version=$Env:ASSEMBLY_VERSION `
```

build/steps-conda.yml
```diff
@@ -7,24 +7,35 @@ steps:
 ##
 # Pre-reqs
 ##
+- task: DownloadBuildArtifacts@0
+  inputs:
+    artifactName: iqsharp
+    downloadPath: '$(System.DefaultWorkingDirectory)/artifact'
+
+- pwsh: |
```
anpaz-msft: how is this going to work with the end-to-end/signed build? name of the drops is different. 
Also, should this be inside a script so we dont' have to copy the same commands?

build/pack.ps1
cgranade: Thanks for the catch, somehow I had thought those propagated from the build to the publish step, but that would explain it. I'll fix that now, thanks!

build/steps-conda.yml
cgranade: With respect to copying the same commands, my thinking was that it's a single-line script whose parameters vary between the signed and prerelease builds as per your comment, such that making a new script would not save very much over copying and pasting. Happy to discuss that further, though.

PR #48, opened 2019-10-19
Support for runtime settings

anpaz-msft:
This commit adds a new `%settings` magic that can be used to read/write runtime settings.
Currently, there are some settings we're reading from environment or config files, like the current Workspace folder or the default package versions but it's not possible to see their values nor to modify them, with this commit there is a new `IRuntimeSettings` service that can be used as a storage of key=value pairs, which has an event raised whenever a setting is modified; plus a new %settings` magic that reports all the current settings plus will add/update any settings specified in the cell.


cgranade:
There's existing precedent for [`%config`](https://ipython.org/ipython-doc/3/config/intro.html#the-config-magic) rather than `%settings`, perhaps it might make sense to follow that convention?

anpaz-msft:
hey @cgranade, I agree with the `%config` and made the change.
let me know if you have other concerns. I'd like to use this to start working on removing iq# dependency on other repos. This is needed so:
1. I can specify the default namespaces to open (so we can remove `Standard` from the unittests)
2. I can specify the default packages to load (similar, to remove `Standard)

Thanks.

anpaz-msft:
this was implemented with the `%config` pattern.

Issue #20, opened 2019-02-28
Python package should not find .qs files recursively.


anpaz-msft:
Currently, the Python qsharp module recursively tries to find all `.qs` files in all subdirectories of the current working folder, which is causing some confusion as that's not normal Python behavior.

We should just load the .qs files under the current working folder and find some other mechanism to import/load files from other folders.

ScottCarda-MS:
PR for this issue is #49.

cgranade:
Closing since #49 is now in, and as the rest of the work is tracked as #24.

PR #49, opened 2019-10-21
Workspace Loader Only Check Top Level Directory

ScottCarda-MS:
Changed the behavior of the workspace loader's file search to only check for qs file in the top level folder.

### DIFF
--- a/src/Core/Workspace/Workspace.cs
+++ b/src/Core/Workspace/Workspace.cs
@@ -222,7 +222,7 @@ namespace Microsoft.Quantum.IQSharp
 
                 if (File.Exists(CacheDll)) { File.Delete(CacheDll); }
 
-                files = Directory.EnumerateFiles(Root, "*.qs", SearchOption.AllDirectories).ToArray();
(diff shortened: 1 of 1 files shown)


ScottCarda-MS:
PR for issue #20.

tcNickolas:
Would it be possible to get instructions on how to check that this change doesn't break the katas?

cgranade:
> Would it be possible to get instructions on how to check that this change doesn't break the katas?

This change causes that, when called from IQ# or Python, all Q# code must be in the current working directory or in a referenced package. There should be an alpha package from this PR on the prerelease feed that will allow for testing, just need to make sure you have a NuGet.Config set up to point to that feed (this repo has an example of such a config in its root).

anpaz-msft:
@tcNickolas , I took a quick look at Katas, and all the .qs files are in the same folder as the notebook, so it should not affect them.
There are a couple of other .qs files under the `$/utilities` folder, but should not be affected.

ScottCarda-MS:
@tcNickolas mentioned that this may cause issues with the DeutschJozsaAlgorithm kata. I am at this time unable to test this change against that kata.

Issue #31, opened 2019-08-14
Misc syntax highlighting errors.


cgranade:
**Describe the bug**
Syntax highlighting in IQ# exhibits several small issues:
- Parens before types are colored with the types.
- The `newtype` keyword is not colored.
- The `%` operator is treated as a comment.

**Screenshots**
![image](https://user-images.githubusercontent.com/31516/63063819-127b3a80-beb3-11e9-9fe1-6c57dae96ff5.png)

**System information**
 - OS: Windows 10
 - Browser: Chromium Edge

**Additional context**
![image](https://user-images.githubusercontent.com/31516/63063830-21fa8380-beb3-11e9-9a4c-dc53d6eb2abc.png)

RutujaShirali:
Working on this now

tcNickolas:
Resolved in #53.

PR #53, opened 2019-11-01
Update kernel.js

rvpradeep:
Fix for Misc syntax highlighting errors. #31

image

### DIFF
--- a/src/Jupyter/res/kernel.js
+++ b/src/Jupyter/res/kernel.js
@@ -18,7 +18,7 @@ define(
                         {
                             token: "comment",
                             // include % to support kata special commands
-                            regex: /(\/\/|%kata).*/
+                            regex: /(\/\/|%kata|%version|%simulate).*/
                         },
                         {
                             token: "string",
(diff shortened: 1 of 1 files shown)


src/Jupyter/res/kernel.js
```diff
@@ -18,7 +18,7 @@ define(
                         {
                             token: "comment",
                             // include % to support kata special commands
-                            regex: /(\/\/|%).*/
+                            regex: /(\/\/|%kata).*/
```
cgranade: It might be good to include other magic commands (e.g. `%version`, `%simulate`) here as well.

rvpradeep: Sure! Updated the PR

tcNickolas: We'll need to list all our special commands if we are to support them individually... Off the top of my head I remember `%package`, `%workspace` and `%check_kata`

tcNickolas: Also `%package`, `%workspace` and `%check_kata`

rvpradeep:
I do not have write access to the repo so I will not be able to merge the pull req.

PR #57, opened 2019-11-11
Workspace and compilation unit caching

bettinaheim:
This PR includes the caching that @anpaz-msft set up, and additionally also keeps and updates a compilation unit instead of rebuilding it from scratch every time, and thus specifically only reloads references if it needs to.


bettinaheim:
I'll put up a separate PR for the Compilation Unit caching and in-memory resource processing.

PR #59, opened 2019-11-14
Fixing issues when using dlls without assembly attributes

bettinaheim:


src/Core/Compiler/CompilerService.cs
```diff
@@ -130,7 +137,16 @@ private static AssemblyInfo BuildAssembly(Uri[] fileNames, QsCompiler.SyntaxTree
...
+                    var syntaxTreeFile = Path.Combine(Path.GetDirectoryName(targetDll), Path.GetFileNameWithoutExtension(targetDll) + ".bson");
+                    var resourceDescription = new ResourceDescription
+                    (
+                        resourceName: QsCompiler.ReservedKeywords.DotnetCoreDll.ResourceName,
+                        dataProvider: () => File.OpenRead(syntaxTreeFile),
+                        isPublic: true
+                    );
+
+
+                    EmitResult result = compilation.Emit(ms, manifestResources: new[] { resourceDescription });
```
cgranade:
```suggestion
                    var result = compilation.Emit(ms, manifestResources: new[] { resourceDescription });
```

bettinaheim:
As discussed with @anpaz-msft  and @cgranade, I'll merge this to continue with perf.

PR #60, opened 2019-11-15
Access to compilation unit manager and fast id resolution

bettinaheim:
A compilation unit is cached and updated. 
It is no longer needed that the resource to embed into the C# dll is written to disk first. 
Determining which declarations exist base on lexing, without having to do the full compilation.

### DIFF
--- a/src/Core/Compiler/CompilerService.cs
+++ b/src/Core/Compiler/CompilerService.cs
@@ -23,6 +23,7 @@ using Microsoft.Quantum.QsCompiler.Transformations.BasicTransformations;
 using Newtonsoft.Json.Bson;
 using QsReferences = Microsoft.Quantum.QsCompiler.CompilationBuilder.References;
 
+
 namespace Microsoft.Quantum.IQSharp
 {
     /// <summary>
@@ -35,7 +36,7 @@ namespace Microsoft.Quantum.IQSharp
         private readonly CompilationUnitManager CompilationManager;
         private ImmutableHashSet<Uri> LoadedFileManagers;
         private ImmutableHashSet<NonNullable<string>> LoadedReferences;
-
+        
         public CompilerService()
         {
             this.Logger = new QSharpLogger(null);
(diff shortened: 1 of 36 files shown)


src/Core/Compiler/CompilerService.cs
```diff
@@ -28,98 +31,112 @@ namespace Microsoft.Quantum.IQSharp
...
-        /// </summary>
-        public AssemblyInfo BuildFiles(string[] files, CompilerMetadata metadatas, QSharpLogger logger, string dllName)
+        /// <summary> 
+        /// Compiles the given Q# code and returns the list of elements found in it. 
+        /// Removes all currently tracked source files in the CompilationManager and replaces them with the given ones.  
+        /// The compiler does this on a best effort, so it will return the elements even if the compilation fails. 
+        /// If the given references are not null, reloads the references loaded in by the CompilationManager  
+        /// if the keys of the given references differ from the currently loaded ones. 
+        /// Returns an enumerable of all namespaces, including the content from both source files and references.  
+        /// If generateFunctorSupport is set to true, replaces all auto-generation directives in the built syntax tree with the generated implementation.
```
cgranade: Is `generateFunctorSupport` supposed to be a parameter to the method?

bettinaheim: Ah, sorry, it was but is no longer needed - will update the doc comment. Thanks!

cgranade (dismissed):
I ran this locally and it worked fine, does indeed seem more responsive as well! Thanks for adding this in!

ricardo-espinoza (dismissed):

rmshaffer:
@bettinaheim just checking on the status of this PR. Are these changes still relevant? Any reason this hasn't been merged yet?

rmshaffer:
Closing this stale PR for now. @bettinaheim, please let me know if these changes are still relevant. If so, we should probably merge this branch with the latest code, re-test, and start a new PR.

PR #62, opened 2019-11-17
Loading from attributes & other perf related things

bettinaheim:
For notebooks running on classical hardware, loading references based on attributes is a lot faster. I hence added an override switch in the compiler. 
I also added the in-memory resource generation.


bettinaheim:
@anpaz-msft, @cgranade  Given that this is what is currently in our release, I am merging this to make sure our internal end-to-end build is in working state. I am happy to make a separate PR for any requested changes.

PR #70, opened 2019-12-17
Changes for QDK pipeline

anpaz-msft:
A couple of new build variables to disable some build features:
- Enable.Python
- Enable.Docker
If explicitly disabled (i.e. they are set to false), the corresponding build steps are skipped, instead a warning is created.
Also, the version of the telemetry (Aria) package is now controlled by a build variable.


build/pack.ps1
```diff
@@ -111,14 +111,22 @@ Pack-Nuget '../src/Core/Core.csproj'
...
-Write-Host "##[info]Packing Python wheel..."
-python --version
-Pack-Wheel '../src/Python/'
+if ($Env:ENABLE_PYTHON -eq "false") {
+    Write-Host "##vso[task.logissue type=warning;]Skipping Creating Python packages. Env:ENABLE_PYTHON was set to 'false'."
+} else {
+    Write-Host "##[info]Packing IQ# as self-contained executables."
+    Pack-Exe "../src/Tool/Tool.csproj" -Runtime win10-x64
+    Pack-Exe "../src/Tool/Tool.csproj" -Runtime osx-x64
+    Pack-Exe "../src/Tool/Tool.csproj" -Runtime linux-x64
```
cgranade: These steps are part of the conda feature, and aren't included in Python wheels. I'd suggest moving these to be  guarded by `$Env:ENABLE_CONDA` instead, perhaps?

images/iqsharp-base/Dockerfile
```diff
@@ -93,7 +93,7 @@ ENV PATH=$PATH:${HOME}/dotnet:${HOME}/.dotnet/tools \
 # Install IQ# and the project templates, using the NuGet packages from the
 # build context.
 ARG IQSHARP_VERSION
-RUN dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.10.1910.2802-beta" && \
+RUN dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.10.1912.1705-beta" && \
```
cgranade: Should this also be an `ARG` perhaps, to prevent it getting similarly out of date in the future? I know `bootstrap.ps1` should update it, so perhaps moot anyway.

build/pack.ps1
anpaz-msft: Good point.

PR #81, opened 2020-01-15
Add new result encoders to output state vectors as HTML.

cgranade:
This adds new functionality to expose state dumps from `DumpMachine` and `DumpRegister` as HTML tables:

image


cgranade:
As per feedback from @bettinaheim and using suggestions from @anpaz-msft, this PR now includes a new `%config` magic that allows for configuring the behaviour of the state dumper:

image

cgranade:
Looks like the build is failing because I forgot to adapt a unit test to include the new configuration source. I will do that shortly and update the PR accordingly.

src/Jupyter/Extensions.cs
```diff
@@ -23,6 +23,21 @@ public static void AddIQSharpKernel(this IServiceCollection services)
         {
             services.AddSingleton<ISymbolResolver, Jupyter.SymbolResolver>();
             services.AddSingleton<IExecutionEngine, Jupyter.IQSharpEngine>();
+            services.AddSingleton<IConfigurationSource, Jupyter.IQSharpEngine>(
+                provider => provider.GetService<IExecutionEngine>() as IQSharpEngine
```
anpaz-msft: I'm not sure we should have the `IQSharpEngine` double as both, the `IExecutionEngine` engine and the `IConfigurationSource`.
I recommend just create a simple configuration class.

src/Jupyter/IQSharpEngine.cs
```diff
@@ -35,8 +43,11 @@ IServiceProvider services
 
             RegisterDisplayEncoder(new IQSharpSymbolToHtmlResultEncoder());
             RegisterDisplayEncoder(new IQSharpSymbolToTextResultEncoder());
+            RegisterDisplayEncoder(new StateVectorToHtmlResultEncoder());
+            RegisterDisplayEncoder(new StateVectorToTextResultEncoder());
             RegisterJsonEncoder(TupleConverters.Converters);
-            
+            RegisterJsonEncoder(new StringEnumConverter());
```
anpaz-msft: What is this about? Would it make sense to instead add it to the list in `TupleConverters.Converters`?

src/Jupyter/Magic/ConfigMagic.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (parts.Length != 2)
+                {
+                    return "Expected config option in the form key = value."
+                           .ToExecutionResult(ExecuteStatus.Error);
+                }
+                var key = parts[0].Trim();
+                var value = JToken.Parse(parts[1]);
+                ConfigurationSource.Configuration[key] = value;
+                // Serialize back to a string for reporting to the user.
+                return JsonConvert.SerializeObject(value).ToExecutionResult();
```
anpaz-msft: Would it be better to report back all the settings and not just the last one? I think it's convenient to see all values.

src/Jupyter/Visualization/JupyterSimulator.cs
```diff
@@ -0,0 +1,119 @@
...
+        : base(throwOnReleasingQubitsNotInZeroState, randomNumberGeneratorSeed, disableBorrowing)
+        {
+            Channel = channel;
+            ConfigurationSource = configurationSource;
+            // Default to setting the channel to be the handler for Message.
+            DisableLogToConsole();
+            OnLog += Channel.Stdout;
+        }
+
+        protected override QVoid Dump<T>(T target, IQArray<Qubit>? qubits = null)
```
anpaz-msft: In order to make it a bit more flexible, what we can do is instead of extending the QuantumSimulator and override the `Dump` method, create two native implementations of `Dump` and `DumpMachine` (similar to [the ootb implementation](https://github.com/microsoft/qsharp-runtime/blob/4109c3bdd8f8048d1309b85bb81784ea61a2d7ff/src/Simulation/Simulators/QuantumSimulator/Dump.cs#L66)), and then instead of using a `JupyterSimulator`, call the `Register` method of the `QuantumSimulator` to override the default `DumpMachine` with this new implementation.
Doing it this way, would allow us to easily incorporate it on the Katas simulator (all it will need to do is do the same registration).
Let's talk tomorrow if you have questions...

src/Jupyter/Extensions.cs
```diff
@@ -23,6 +23,21 @@ public static void AddIQSharpKernel(this IServiceCollection services)
...
+            );
```
anpaz-msft: It would be nice if we could provide default values of these configuration parameters from files or env variables.

src/Jupyter/Extensions.cs
cgranade: Good point, I'll split that out. Thankfully shouldn't require any changes to the consumers thanks to dependency injection.

src/Jupyter/IQSharpEngine.cs
cgranade: This is needed for enum-valued configuration options. I'll move it into `TupleConverters`, then.

src/Jupyter/Magic/ConfigMagic.cs
cgranade: My thought was that running `%config` on its own reports everything, but that keeping it minimal as a response to setting a single option helps keep things readable as the number of config options increases.

src/Jupyter/Extensions.cs
cgranade: Agreed, my thought was to put that in a future PR, though, as there's a few more details there that I'd like to work out first if that's OK?

src/Jupyter/Visualization/JupyterSimulator.cs
cgranade: Good idea, I'll go on and do that, then.

src/Jupyter/ConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+            ? token.ToObject<T>()
+            : defaultValue;
+
+        public BasisStateLabelingConvention BasisStateLabelingConvention =>
+            GetOptionOrDefault("dump.basisStateLabelingConvention", BasisStateLabelingConvention.LittleEndian);
+
+        public bool TruncateSmallAmplitudes =>
+            GetOptionOrDefault("dump.truncateSmallAmplitudes", false);
+
+        public double TruncationThreshold =>
```
anpaz-msft: so they allow implementation in interfaces now?

src/Jupyter/Extensions.cs
```diff
@@ -23,6 +25,53 @@ public static void AddIQSharpKernel(this IServiceCollection services)
...
+        {
+            simulator.DisableLogToConsole();
+            simulator.OnLog += channel.Stdout;
+
+            simulator.Register(
+                typeof(Diagnostics.DumpMachine<>), typeof(JupyterDumpMachine<>),
+                signature: typeof(ICallable)
+            );
+
+            var op = ((GenericCallable)simulator.GetInstance(typeof(Microsoft.Quantum.Diagnostics.DumpMachine<>)));
```
anpaz-msft: oh, wow. it is indeed ugly as hell :(

src/Jupyter/Magic/Simulate.cs
```diff
@@ -33,15 +36,9 @@ public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
...
-            using (var qsim = new QuantumSimulator())
-            {
-                qsim.DisableLogToConsole();
-                qsim.OnLog += channel.Stdout;
-
-                var value = await symbol.Operation.RunAsync(qsim, args);
-
-                return value.ToExecutionResult();
-            }
+            using var qsim = new QuantumSimulator().WithJupyterDisplay(channel, ConfigurationSource);
```
anpaz-msft: is this new syntax for `using`? you're way more up to date than me.

src/Jupyter/ConfigurationSource.cs
cgranade: That's a new C# 8 feature, default properties and methods... as far as I understand, intended to provide traits-style programming. I'm still wrapping my head around the concept of a _private_ method in an interface, that takes some getting used to.

src/Jupyter/Extensions.cs
cgranade: Yeah, sorry about that. I think we just haven't quite exercised this particular use case for `Register` before. As per your offline suggestion, we may be able to remove this ugliness by adding a new event to IQuantumProcessor instead.

src/Jupyter/Magic/Simulate.cs
cgranade: The semantics of `using var` are identical to the old `using`, but implicitly places the closing `}` for the `using` statement at the end of the current scope. I find it pretty useful for when I want to have a `using` for most or all of a method scope.

cgranade:
Looks like moving the enum converter from `TupleConverter` to the specific configuration preference being converted fixed the serialization tests. If you could take another look, @anpaz-msft, that fix dismissed your review. Thank you!

Issue #79, opened 2019-12-20
Katas taking 4+ mins to execute on binder


anpaz-msft:
For some reason, after updating to the latest QDK and launching the Katas in Binder, trying to run a %kata cell takes 4+ minutes.

We're currently investigating this.

cgranade:
As per @anpaz-msft's report, it appears as though the affected image is `mcr.microsoft.com/quantum/iqsharp-base:0.10.1912.0501`, and that `mcr.microsoft.com/quantum/iqsharp-base:0.10.1911.1607` works well on mybinder.org. Both images seem to work well when running the images locally.

rmshaffer:
Two fixes have been made here:
- in the base IQ# image: https://github.com/microsoft/iqsharp/commit/84bfc6b1da501e31fff93b6a544256eb4c573ab3
- in the Katas image, by @vxfield: https://github.com/microsoft/QuantumKatas/commit/6ddf60485ed3ad5bffe463cc85548ed1915e8c94

Performance on MyBinder is much improved with these two fixes. First cell execution times are typically only a couple of seconds or less.

PR #85, opened 2020-01-21
Add logging to file, config from environment variables, and performance monitoring.

cgranade:
This PR adds new functionality to IQ# to help diagnose #79, including logging to files, configuration of logging behavior from environment variables, and a performance monitoring service that reports on managed and unmanaged RAM usage.


cgranade:
As a quick update, I added a magic command that exposes the same information as a table instead for convenience in hosted contexts:
image

cgranade:
> Had to adapt the place we read the `IQSHARP_TELEMETRY_OPT_OUT`, except for that everything is working well.

Thanks for catching that, and thanks for the review. I really appreciate it!

PR #92, opened 2020-02-05
Minor improvements to Jupyter displays.

cgranade:
This PR fixes an issue with sorting basis state labels for more than 3 qubits, and formats resources estimates into tables: 
image


src/Jupyter/Magic/EstimateMagic.cs
```diff
@@ -36,10 +38,10 @@ public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
 
             var qsim = new ResourcesEstimator();
             qsim.DisableLogToConsole();
-            
+
             await symbol.Operation.RunAsync(qsim, args);
 
-            return qsim.AsDictionary().ToExecutionResult();
+            return qsim.ToExecutionResult();
```
anpaz-msft: you're returning the simulator.... curious, what happens on clients that do not support HTML, what gets generated?
Would it be better to return the `qsim.Data`, which is a [DataTable](https://docs.microsoft.com/en-us/dotnet/api/system.data.datatable?view=netstandard-2.1) and write an encoder for `DataTable`? More generic and could be used in more places. Maybe write an `DataTable` encoder for json/text too?

cgranade:
> I'm marking this as "Request Changes" for the DataTable think I'm suggesting. If you feel strongly about the current approach, stop by and we can discuss it.

Makes sense! I did strongly consider the `DataTable` approach before taking this, but the main trouble I ran into is that `DataTable` alone doesn't tell us that it's the _particular_ data table returned by `ResourcesEstimator`. We could perhaps have the encoder check if it's a datatable with the right column headers, but then that would make writing the JsonConverter a bit harder. I also tried making a class that just wraps an `IDictionary` with extra metadata that could be used to format the table, but that wound up being a lot heavier-weight than what seemed reasonable. Happy to go back to `DataTable`, though, and look at how to make the JsonConverter work.

src/Jupyter/Magic/EstimateMagic.cs
cgranade: Ah, good point. I checked HTML and JSON output, but forgot text/plain...

anpaz-msft:
Why do you need to check if the `DataTable` is one generated by the `ResourceEstimator`... there is nothing special about the `ResourcesEstimator` data table. You can figure out the name of the columns from the `.Columns` field and then you can iterate on the Rows.  
See how we do it on the [`ToTSV` method](https://github.com/microsoft/qsharp-runtime/blob/28934371f6d0f19daa3508a969c393d7809ffafb/src/Simulation/Simulators/ResourcesEstimator/ResourcesEstimator.cs#L161) if you need some inspiration.

cgranade:
Sorry for the delay. I came up with what I think is a fairly elegant solution: convert `DataTable` instances to instances of `Table<DataRow>` and use the existing table formatting logic for arbitrary data tables. This should now be good to go, hopefully.

PR #94, opened 2020-02-12
Improve package loading by using async enumerables, reporting status.

cgranade:
This PR improves package loading times by using async enumeration over NuGet search results to eliminate extraneous HTTP requests to nuget.org, and uses microsoft/jupyter-core#44 to report package download progress back to the user. 

pkgs


src/Core/References/NugetPackages.cs
```diff
@@ -175,32 +177,34 @@ public async Task<PackageIdentity> ParsePackageId(string package)
         /// <summary>
         /// Adds a new package given the name and version as strings.
         /// </summary>
-        public async Task<PackageIdentity> Add(string package)
+        public async Task<PackageIdentity> Add(string package, Action<string>? statusAction = null)
```
anpaz-msft: Maybe call this argument `statusCallback`?

cgranade: Good idea, I'll go on and change accordingly.

PR #98, opened 2020-03-02
Document all public APIs in Jupyter.csproj.

cgranade:
This PR adds API documentation for all public APIs in the Jupyter project for IQ#.


src/Jupyter/ConfigurationSource.cs
```diff
@@ -13,35 +13,76 @@
...
         void Persist();
 
         private T GetOptionOrDefault<T>(string optionName, T defaultValue) =>
             Configuration.TryGetValue(optionName, out var token)
             ? token.ToObject<T>() ?? defaultValue
             : defaultValue;
 
+        /// <summary>
+        ///     The labeling convention to be used when labeling computational
+        ///     basis states.
```
tcNickolas:
```suggestion
        ///     basis states (bit string, little-endian or big-endian).
```

src/Jupyter/KernelProperties.cs
```diff
@@ -18,7 +22,7 @@ public static class Constants
             DisplayName = "Q#",
 
             LanguageName = "qsharp",
-            LanguageVersion = "0.4",
+            LanguageVersion = "0.10",
```
tcNickolas: Will this change update the `"version": "0.4"` field in the notebooks the next time they are edited and saved? Not objecting, just want to be prepared :-)

cgranade: I believe so, yes. Ideally, we'd keep this in sync so that the metadata in notebooks is accurate, but as you can tell we haven't been great about that...

cgranade:
Thanks for the review, @tcNickolas! I've addressed your comment, which as a side effect dismissed your review. If you could please approve again, I'd appreciate it. Thanks!

PR #99, opened 2020-03-04
Use OnException event to display stack traces for exceptions.

cgranade:
This PR uses the `OnException` event introduced by https://github.com/microsoft/qsharp-runtime/pull/157 to format exception stack traces as an HTML details element, making it easier to hide extraneous information while still allowing interested users to drill down through stack traces.

Collapsed:
image

Expanded:
image

When stack traces contain more useful information (e.g.: source locations for library functions and operations), that information is formatted as a hyperlink using @vadym-kl's logic for finding best source locations added in https://github.com/microsoft/qsharp-runtime/pull/79.

This PR is marked as draft in lieu of API docs (#98) and feedback on the UI design, as well as the completion of https://github.com/microsoft/qsharp-runtime/pull/157.

### DIFF
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -34,9 +34,9 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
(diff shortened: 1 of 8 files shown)


cgranade:
@anpaz-msft: Thanks for the  feedback! I've confirmed that the Python output is unaffected, since the changes here set up result encoders for both `text/plain` and `text/html` (see screenshot below). I've also added the same logic to `%toffoli` and `%estimate`. Once I've finished API docs, I'll mark this as being ready. Thanks!

image

cgranade:
Thanks for the review, @vxfield, and for the suggestions for future improvements! I think that could work great for improving how result encoders work going forward.

PR #100, opened 2020-03-05
Update iqsharp-base to .NET Core SDK 3.1

cgranade:
Now that .NET Core SDK 3.0 is EOL, this PR updates iqsharp-base to use .NET Core SDK 3.1 instead.


cgranade:
> Very straightforward, and I'm assuming there is not breaking changes with 3.1 that would impact us.

We already test on 3.1 on all of our CI builds, I think updating the Dockerfile just got missed in the update. Thanks for the review!

Issue #102, opened 2020-03-09
Installing MS QDK to run Q# with python on Ubuntu 18 — “dotnet iqsharp install” step fails


swernli:
Copied from a [StackOverflow question](https://stackoverflow.com/questions/60593638/installing-ms-qdk-to-run-q-with-python-on-ubuntu-18-dotnet-iqsharp-install?newreg=9c6c33dc21f541e98aac2280569562dd).

Here's the original question and repro steps:

Using the instructions in https://docs.microsoft.com/en-us/quantum/install-guide/pyinstall

I get following:

dotnet iqsharp install

It was not possible to find any compatible framework version The framework 'Microsoft.AspNetCore.App', version '3.0.0' was not found. - No frameworks were found.

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at: - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.AspNetCore.App&framework_version=3.0.0&arch=x64&rid=ubuntu.18.04-x64

It seems you cannot get AspNetCore.App 3.0.0 anymore but I have 3.0.3 installed - See below here for my setup .....
dotnet --info

.NET Core SDK (reflecting any global.json): Version: 3.1.102 Commit: 573d158fea

Runtime Environment: OS Name: ubuntu OS Version: 18.04 OS Platform: Linux RID: ubuntu.18.04-x64 Base Path: /usr/share/dotnet/sdk/3.1.102/

Host (useful for support): Version: 3.1.2 Commit: 916b5cba26

.NET Core SDKs installed:

3.0.103 [/usr/share/dotnet/sdk]

3.1.102 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:

Microsoft.AspNetCore.App 3.0.3 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App] Microsoft.AspNetCore.App 3.1.2 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App] Microsoft.NETCore.App 3.0.3 [/usr/share/dotnet/shared/Microsoft.NETCore.App] Microsoft.NETCore.App 3.1.2 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs: https://aka.ms/dotnet-download

[...]

PR #103, opened 2020-03-09
Updating .NET references

swernli:
Updating the .NET references used in project files. Follows the recommendation of skipping explicit version on AspNetCore.App references found [here](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#recommendation).

See related discussion in #102


cgranade:
As noted at #102, I absolutely agree with this change. It is strictly speaking a breaking change for IQ# users, though, such that there is release note impact for the March release. +@bromeg and @cpalmer2020 for release notes impact.

PR #104, opened 2020-03-10
Updating some missed files from netcoreapp3.0 to netcoreapp3.1

bettinaheim:
Looks like e.g. the launch configuration was missed.

### DIFF
--- a/src/Tool/Tool.csproj.orig
+++ b/src/Tool/Tool.csproj.orig
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <PlatformTarget>x64</PlatformTarget>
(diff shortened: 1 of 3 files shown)


src/Tool/Tool.csproj.orig
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <PlatformTarget>x64</PlatformTarget>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
```
swernli: It looks like this is an old file kept for reference, so it doesn’t technically need to be updated. That said, it shouldn’t hurt.

bettinaheim: I don't know why it's kept around - I reverted the change.

PR #114, opened 2020-04-07
Update jupyter core 1.3, begin adding magic command docs.

cgranade:
This PR updates to Microsoft.Jupyter.Core 1.3, including async and improved symbol documentation, and begins to use 1.3 to improve magic symbol documentation using `%config` as an example.

### DIFF
--- a/src/Jupyter/Extensions.cs
+++ b/src/Jupyter/Extensions.cs
@@ -170,8 +170,31 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
             return simulator;
         }
 
-        internal static string TrimLeadingWhitespace(this string s) =>
-            new Regex(@"^\s+", RegexOptions.Multiline)
-                .Replace(s, string.Empty);
+        /// <summary>
+        ///      Removes common indents from each line in a string,
+        ///      similarly to Python's <c>textwrap.dedent()</c> function.
+        /// </summary>
+        internal static string Dedent(this string text)
+        {
+            // First, start by finding the length of common indents,
+            // disregarding lines that are only whitespace.
+            var leadingWhitespaceRegex = new Regex(@"^[ \t]*");
+            var minWhitespace = int.MaxValue;
+            foreach (var line in text.Split("\n"))
+            {
+                if (!string.IsNullOrWhiteSpace(line))
+                {
+                    var match = leadingWhitespaceRegex.Match(line);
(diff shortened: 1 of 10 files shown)


src/Jupyter/Extensions.cs
```diff
@@ -169,5 +169,9 @@ public static T WithStackTraceDisplay<T>(this T simulator, IChannel channel)
             };
             return simulator;
         }
+
+        internal static string TrimLeadingWhitespace(this string s) =>
+            new Regex(@"^\s+", RegexOptions.Multiline)
+                .Replace(s, string.Empty);
```
rmshaffer, line 175: can this just be string.TrimStart()?

src/Jupyter/Magic/ConfigMagic.cs
```diff
@@ -25,7 +25,48 @@ public class ConfigMagic : AbstractMagic
...
+                    @"
+                        Print a list of all currently set configuration options:
+                        ```
+                        In []: %config
+                        Out[]: Configuration key                 Value
+                               --------------------------------- -----------
+                               dump.basisStateLabelingConvention ""BigEndian""
+                               dump.truncateSmallAmplitudes      true
+                        ```
+                    ",
```
rmshaffer, line 47: .TrimLeadingWhitespace()? (or .TrimStart()?)

src/Jupyter/Extensions.cs
cgranade: The main difference is that it trims whitespace from the start of each individual line, rather than from the string as a whole. Useful for formatting `@""` strings that are indented along with code, such as at https://github.com/microsoft/iqsharp/pull/114/files#diff-19b4af072f33fc9bfc7bda1e8a134d74R28. I'll add a comment to that effect, thanks!

src/Jupyter/Magic/ConfigMagic.cs
cgranade: Good point, neither does quite the right thing here. I'm looking for something like Python's [`dedent()`](https://docs.python.org/3.8/library/textwrap.html#textwrap.dedent) function... will take another stab.

src/Jupyter/Extensions.cs
rmshaffer: ah, got it! I see, that's what `RegexOptions.Multiline` means 👍

PR #115, opened 2020-04-08
Fix typo in pack.ps1.

cgranade:
This PR fixes a one character typo in `pack.ps1` that causes PowerShell to exit when either Docker or Python build steps are disabled, as the `\` command is not defined.


cgranade:
Thanks for the review, @rmshaffer!

PR #117, opened 2020-04-15
Separate doc building into new script for QDK build.

cgranade:
This PR separates out the docs build script so that it is easier to call in all-QDK builds, and to disable in Docker configurations that don't support building Linux images (e.g.: when Docker for Windows is configured to use Windows containers).

### DIFF
--- a/build/pack.ps1
+++ b/build/pack.ps1
@@ -50,11 +50,6 @@ function Pack-Image() {
         [string] $Dockerfile
     );
 
-    if (($Env:AGENT_OS -ne $null) -and ($Env:AGENT_OS.StartsWith("Win"))) {
-        Write-Host "--> Cannot create docker image on Windows."
-        return
-    }
-
     Try {
         docker version
     } Catch {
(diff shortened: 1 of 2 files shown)


build/pack.ps1
```diff
@@ -143,47 +138,26 @@ if ($Env:ENABLE_PYTHON -eq "false") {
...
+    $runDocker = $true;
+}
+
+if ($runDocker) {
     Write-Host "##[info]Packing Docker image..."
     Pack-Image -RepoName "iqsharp-base" -Dockerfile '../images/iqsharp-base/Dockerfile'
 }
 
-if (($Env:ENABLE_DOCKER -eq "false") -or ($Env:ENABLE_PYTHON -eq "false")) {
+if (-not $runDocker -or ($Env:ENABLE_PYTHON -eq "false")) {
```
rmshaffer, line 158: for consistency, do you want to also create a variable like `$enablePython`? for use both here and above on line 133.

cgranade: My thinking here was that `$runDocker` captures both `$Env:ENABLE_DOCKER` and the requirement of a Linux host, where as the Python condition is simpler. From the perspective of minimizing changes, I wanted keep the existing logic for enabling Python given that. Happy to change for consistency, though, if you would prefer.

rmshaffer: No, makes sense, minimizing changes is a good reason to leave it alone.

cgranade: Sounds good, thanks!

PR #118, opened 2020-04-15
Improve visual formatting for the output of %lsmagic.

cgranade:
image

### DIFF
--- a/src/Jupyter/Magic/LsMagicMagic.cs
+++ b/src/Jupyter/Magic/LsMagicMagic.cs
@@ -10,6 +10,17 @@ using Microsoft.Quantum.IQSharp.Jupyter;
 namespace Microsoft.Quantum.IQSharp.Jupyter
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


tcNickolas:
Is there a way to left-justify this? Right-justify looks weird for this kind of output to me...

cgranade:
> Is there a way to left-justify this? Right-justify looks weird for this kind of output to me...

It would be possible, but would take a bit more effort since the right-alignment is configured by the Jupyter Notebook client. I think that's a good idea for a future enhancement, but probably beyond the scope of this PR.

cgranade:
Looks like the right-align styling is coming from https://github.com/jupyter/notebook/blob/b8b66332e2023e83d2ee04f83d8814f567e01a4e/notebook/static/notebook/less/renderedhtml.less#L92. We could override by setting our own CSS properties in the generated output, but doing so would require improvements to jupyter-core where the HTML is generated. I agree, this could be a good thing to address in the future, though.

PR #132, opened 2020-04-30
IQ# magic commands for Azure integration

rmshaffer:
This PR is targeted at the feature branch `feature/azure-client` and contains initial implementations of several magic commands intended for eventual Azure integration.

Right now, looking for feedback on the overall structure and the implementation of the magic commands themselves, which should be pretty stable at this point.

This PR is currently a draft, with the following major changes to come before actually merging into the feature branch:

- [ ] Integration with published packages for integration with DataPlane and ResourceManager. (Currently I'm just this running against locally-generated AutoRest code, which is not included in this PR.)
- [ ] The Azure integration (in `AzureClient.cs`) is purely prototype code at this point. Need to revamp this after integrating with the published packages.
- [x] Microsoft.Jupyter.Core package reference needs to be updated to a public version, after a new version of that NuGet package has been released.
- [ ] Mocks for DataPlane and ResourceManager clients (in `AzureClientTests.cs`) need to be extended significantly and corresponding unit tests need to be written.


rmshaffer:
Abandoning this for now, since the plan for integrating with Azure client packages has changed significantly. Will start a new branch and open a new PR for these changes later.

PR #133, opened 2020-05-01
Make Jupyter visualizations available in published library package

### LINKED ISSUE #130: Make Jupyter visualizations available in published library package
Consumers of this project, such as [QuantumKatas](https://github.com/microsoft/quantumkatas), want the ability to use the custom visualizations (`JupyterDumpMachine`, `JupyterDumpRegister`, etc.) currently inside the `Jupyter` project. But this visualization functionality is not contained in the published `Microsoft.Quantum.IQSharp.Core` library package.

Ideally, common functionality for quantum-related visualizations should be accessible through `Microsoft.Quantum.IQSharp.Core` or a similar package so that downstream projects like QuantumKatas can consume it.

**Describe alternatives you've considered**
Duplicating the code in QuantumKatas is one option, but this is very unappealing. The functionality is actually contained in the larger `Microsoft.Quantum.IQSharp` package, but this contains a netcore3.1 tool and cannot be consumed from projects like QuantumKatas.

**Additional context**
Thanks to @vivanwin for reporting this issue.

rmshaffer:
This PR refactors the `Jupyter` project into `Jupyter` and `Kernel` projects. The idea is that there is Jupyter-specific functionality in IQ# (notably visualizations) that should be accessible from other packages without having to import the full tool package.

The `Jupyter` project now contains a NuGet package configuration and will be available as a library.

Remaining tasks before merging to `master`:
- [x] Reference a new public `Microsoft.Jupyter.Core` package, rather than a CI version.
- [x] Validate end-to-end signing works correctly for the new assembly.

Some details on the two projects:
- `Jupyter` now depends only on the `Core` project and on the `Microsoft.Jupyter.Core` package.
   - Code related to `ConfigurationSource` and most of the Jupyter visualization code has been left in this `Jupyter` project.
- `Kernel` depends on the `Core` project, the `Jupyter` project, and several additional packages.
   - The IQ# engine, custom shell, magic commands, etc. have been moved into this project.
   - The `Kernel` project will not be published as a separate package, such that this functionality remains available only through the `Microsoft.Quantum.IQSharp` tool package.

Resolves #130 and allows downstream projects (such as [QuantumKatas](https://github.com/microsoft/quantumkatas)) to consume the Jupyter visualizations from IQ# by importing the new library package.


src/Core/Properties/AssemblyInfo.cs
```diff
@@ -4,4 +4,4 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Tests.IQsharp" + SigningConstants.PUBLIC_KEY)]
-[assembly: InternalsVisibleTo("Microsoft.Quantum.IQSharp.Jupyter" + SigningConstants.PUBLIC_KEY)]
\ No newline at end of file
+[assembly: InternalsVisibleTo("Microsoft.Quantum.IQSharp.Kernel" + SigningConstants.PUBLIC_KEY)]
```
cgranade, line 7: This name change makes sense, but will affect signing process such that we'll have to adapt slightly. Please feel free to sync with me offline if I can help with that adaptation; I'd suggest an end-to-end build at the least to make sure that this renaming doesn't break anything across QDK repos.

src/Jupyter/Jupyter.csproj
```diff
@@ -8,39 +8,40 @@
...
-    <PackageReference Include="System.Reactive" Version="4.3.2" />
+    <PackageReference Include="Microsoft.Jupyter.Core" Version="1.3.9294-CI-20200417-004906" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\Core\Core.csproj" />
   </ItemGroup>
 
   <!-- TypeScript compilation -->
-  <ItemGroup>
```
cgranade, line 34: This is a good change, it was always a bit weird that the TypeScript had to be built from a different project than the kernel itself.

rmshaffer: Now that you point it out, I should also remove the comment saying that the TypeScript compilation is here.

src/Core/Properties/AssemblyInfo.cs
rmshaffer: End-to-end build completed successfully with latest Jupyter Core release.

PR #138, opened 2020-05-05
Improve handling of @EntryPoint() in IQ#

rmshaffer:
This PR updates the IQ# compilation strategy as follows:
- Compile workspace as executable. This allows successful compilation of Q# code files even if they contain an `@EntryPoint()`.
- Continue to compile snippets as library, and ignore the error code `QS6241` (EntryPointInLibrary) when compiling snippets. This means that the `@EntryPoint()` attribute is effectively ignored in the context of Q# notebooks (or compilation from a Python string literal).

In the future, when compiling for job submission, an entry point operation will be auto-generated and compiled into a separate executable which references both the snippets library and the workspace executable.


src/Core/Snippets/Snippets.cs
```diff
@@ -190,11 +190,18 @@ public Snippet Compile(string code)
         /// - either because they have the same id, or because they previously defined an operation
         /// which is in the new Snippet - and replaces them with `newSnippet` itself.
         /// </summary>
-        private IEnumerable<Snippet> SelectSnippetsToCompile(string code)
+        private IEnumerable<Snippet> SelectSnippetsToCompile(string code, QSharpLogger logger)
         {
             var ops = Compiler.IdentifyElements(code).Select(Extensions.ToFullName).ToArray();
             var snippetsWithNoOverlap = Items.Where(s => !s.Elements.Select(Extensions.ToFullName).Intersect(ops).Any());
 
+            var entryPointPattern = @"@\s*EntryPoint\s*\(\s*\)";
```
rmshaffer, line 198: The only way I can see to do this is to replace via regex prior to compilation, since the compiler may throw an error. This isn't ideal, since the regex doesn't know whether what it's removing is truly an EntryPoint attribute. If there is a better suggestion for how to do this, please let me know.

src/Core/Snippets/Snippets.cs
```diff
@@ -149,6 +149,14 @@ public Snippet Compile(string code)
 
             try
             {
+                // strip any @EntryPoint() attributes from the snippet
+                var entryPointPattern = @"@\s*EntryPoint\s*\(\s*\)";
+                if (Regex.IsMatch(code, entryPointPattern))
+                {
+                    code = Regex.Replace(code, entryPointPattern, string.Empty);
+                    logger.LogWarning("@EntryPoint()", "@EntryPoint() attributes are ignored when compiling Q# code in this workspace.");
+                }
```
rmshaffer, line 158: The only way I can see to do this is to replace via regex prior to compilation, since the compiler may throw an error. This isn't ideal, since the regex doesn't know whether what it's removing is truly an EntryPoint attribute. If there is a better suggestion for how to do this, please let me know.

rmshaffer:
Friendly nudge to reviewers @cgranade @bettinaheim @anpaz-msft @efratshabtai - it'd be great if someone (particularly @bettinaheim) could review and/or provide feedback on this change, since I may be missing a better way to do this. Thank you!

src/Core/Snippets/Snippets.cs
```diff
@@ -149,6 +149,14 @@ public Snippet Compile(string code)
 
             try
             {
+                // strip any @EntryPoint() attributes from the snippet
+                var entryPointPattern = @"@\s*EntryPoint\s*\(\s*\)";
```
efratshabtai, line 153: The compiler might already have code to identify EntryPoint() in a code file, we should use the same regex (or other method) here

src/Core/Snippets/Snippets.cs
```diff
@@ -149,6 +149,14 @@ public Snippet Compile(string code)
...
+                if (Regex.IsMatch(code, entryPointPattern))
+                {
+                    code = Regex.Replace(code, entryPointPattern, string.Empty);
+                    logger.LogWarning("@EntryPoint()", "@EntryPoint() attributes are ignored when compiling Q# code in this workspace.");
```
efratshabtai, line 157: Will the user see this warning in the notebook?
If so, can we make it more user friendly, maybe something like: "There is no need to define an @EntryPoint() attribute in a notebook, it will be ignored." (or similar)
'workspace' might be confused with Azure workspace and I don't think the user cares about compilation.

src/Tests/SNIPPETS.cs
```diff
@@ -54,6 +54,23 @@ public static class SNIPPETS
     }
 ";
 
+        public static string Op3_Op4_EntryPoints =
+@"
+    /// # Summary
+    ///     This to show that @EntryPoint() is ignored when compiling from snippets:
```
efratshabtai, line 60: I can't really see but did you try white spaces before and after?

src/Core/Snippets/Snippets.cs
rmshaffer: This warning only goes to the logger, which won't be visible in the notebook itself. I agree though, "workspace" isn't the right word to use here, since it's overloaded both in the Azure sense and also in the [IQ# sense](https://docs.microsoft.com/en-us/qsharp/api/iqsharp-magic/workspace). I also don't want to say only "notebook", because this also applies to compiling code snippets from string literals in Python. Perhaps "...when compiling Q# code defined in notebook cells or in string literals".

src/Tests/SNIPPETS.cs
rmshaffer: I will update this to include a more thorough test of the regex, e.g. `@ EntryPoint (  )` in addition to `@EntryPoint()`.

src/Core/Snippets/Snippets.cs
rmshaffer: Yes, ideally we could just tell the compiler to ignore these attributes, since only the compiler understands the semantics of the language and can truly do this correctly. I believe, however, that's not something the compiler exposes at the moment. @bettinaheim, can you confirm?

bettinaheim: I mean, you could also just discard the error that the compiler generates? The syntax tree (minus the entry point attribute) will be built regardless. 

Well, given that the EntryPoint attribute is in Core and we auto-open Core, you can actually be sure that something called @EntryPoint() is indeed the entry point attribute (and not another attribute with the same unqualified name). 
However, the compiler does some preprocessing like tracking all strings and comments and tagging all scopes before validating the tokens, and that is rather painful to reproduce as a regex. Hence my suggestion above.

rmshaffer: Yes, we could just compile and discard the error generated when multiple entry points exist (which is error QS6236). We don't currently discard/ignore any compiler errors in IQ#, but it looks like it shouldn't be that hard.

However, in this case, it doesn't really produce the desired behavior. It would still compile one of the operations (the first one, it appears) as _the_ entry point for the executable. So if the user has run any cells in the notebook with `@EntryPoint()` — or worse, if they had run one earlier and then deleted it — it puts them in a weird state where all of the non-`@EntryPoint()` operations become invisible and can no longer be executed. Only the first one that they happened to compile would be executable.

And I believe ignoring any `@EntryPoint()` attributes within the notebook cells is actually the functionally correct thing to do here, because we'll eventually be auto-generating a new `@EntryPoint` for job execution, and it shouldn't conflict with anything else that happens to be defined in the notebook.

So I guess, to state the core question: is there a way for IQ# to tell the compiler to simply ignore the existence of these attributes? I stepped all the way down to where we call the C# codegen classes, but I didn't see a way to do this (please let me know if there's something I'm missing!). Otherwise I think this might be the best we can do for now.

(And yes, as written, the regex would strip the string `"@EntryPoint()"` out of string literals or comments or whatever. I can't imagine that's common, so it may be ok to do this, but it's obviously not the ideal way.)

bettinaheim: I was thinking this would be compiled as library, and the error to ignore would be EntryPointInLibrary. Would that make sense?

rmshaffer: Ok, yes. I was stuck in the mindset of compiling as an executable because (a) we want to compile the workspace (.qs files) as an executable, and (b) we will also need to compile the generated entry point into an executable for job submission. But there is another option: we could the snippets into a library which is then *referenced* by that generated entry point.

I'll try to rework these changes as follows:
- Compile workspace as executable and snippets as library.
- Ignore the error EntryPointInLibrary when compiling snippets.

And when we add the generated entry point, we'll compile it into a third assembly that references both the workspace and the snippets. And this eliminates the need for something hacky like the regex. Does that sound good? Also looping in @anpaz-msft for another opinion (sorry that we're designing this in PR comments).

src/Core/Snippets/Snippets.cs
```diff
@@ -145,7 +144,11 @@ public Snippet Compile(string code)
             if (string.IsNullOrWhiteSpace(code)) throw new ArgumentNullException(nameof(code));
 
             var duration = Stopwatch.StartNew();
-            var logger = new QSharpLogger(Logger);
+            var errorCodesToIgnore = new List<string>()
+            {
+                "QS6241",   // Invalid entry point. Only executable Q# projects can have entry points.
```
efratshabtai, line 149: Is this a short term solution (ignoring the error) and if so, what's the long term?

rmshaffer: This is intended to be the long-term solution. Having an `@EntryPoint()` attribute in a notebook cell is somewhat meaningless (since we will automatically generate an entry point whenever trying to execute one of these operations), and we don't want to bother the user with an error related to this when we can just as easily ignore it.

PR #140, opened 2020-05-11
Switch to latest images for build agents

anjbur:
Switching QDK pipeline to use latest images for build agents. There are corresponding PRs in other repos as well.


anjbur:
Looping in @anpaz-msft. I used `*-latest` based on an already-existing PR for this work item that Andres had started. I'm not sure if we can fix on a specific version or if using `*-latest` is necessary for compliance.

anjbur:
@cgranade I talked to Andres offline about the images. He recommended switching to `windows-2019` but staying on `*-latest` for linux and mac as they're primarily used for validation in the build pipeline.

PR #141, opened 2020-05-13
Add AzureClient assembly to manifest

rmshaffer:
Adding `Microsoft.Quantum.IQSharp.AzureClient.dll` to the `manifest.ps1` script. Also adds `manifest.ps1` to the CI script, and fixes the case of the file paths so that the CI passes (`tool` -> `Tool`).

This PR is targeting the `feature/azure-quantum-preview` branch, since that's the only place this assembly exists at the moment.


build/manifest.ps1
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env pwsh
-#Requires -PSEdition Core
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1: Good idea to add this header as well; can we keep the old shebang along with this header?

rmshaffer: Yes, thanks for explaining this offline! Restoring both of these lines so that we'll continue to use PowerShell Core for this script.

PR #144, opened 2020-05-18
Initial IQ# magic command implementation for Azure Quantum

rmshaffer:
This PR targeting `feature/azure-quantum-preview` contains the initial implementation of `%connect`, `%status`, `%submit`, and `%target` magic commands for Azure Quantum integration from IQ#, along with associated test code.

This is a rather large PR, but since everything is so interconnected, it made sense to allow someone to review all of the pieces at once to get the big picture.

There are a number of aspects about the design, user experience, argument parsing, etc. that are not yet final or not yet implemented, but all of these should be relatively minor changes and will be done incrementally. The overall structure here should be stable.

There are additional magic commands planned, which will follow in later PRs. This one is already big enough. Subsequent PRs will be much smaller and more targeted.


src/AzureClient/IAzureClient.cs
```diff
@@ -8,14 +8,119 @@
...
+        /// Sets the specified target for job submission.
+        /// </summary>
+        public Task<ExecutionResult> SetActiveTargetAsync(
+            IChannel channel,
+            string targetName);
+        
+        /// <summary>
+        /// Prints the specified target for job submission.
+        /// </summary>
+        public Task<ExecutionResult> PrintActiveTargetAsync(
```
efratshabtai, line 104: wonder why we need an async version of PrintActiveTarget - do we expect this to be blocking?

efratshabtai: To clarify - Isn't the active target saved locally so this doesn't require a call to the server?

src/AzureClient/Magic/AzureClientMagicBase.cs
```diff
@@ -0,0 +1,42 @@
...
+    public abstract class AzureClientMagicBase : AbstractMagic
+    {
+        /// <summary>
+        ///     The object used by this magic command to interact with Azure.
+        /// </summary>
+        public IAzureClient AzureClient { get; }
+
+        /// <summary>
+        ///     Constructs the Azure Client magic command with the specified keyword
+        ///     and documentation.
```
efratshabtai, line 25: I don't understand from this description what is the keyword here - can you please elaborate? 
same for documentation.

src/AzureClient/Magic/AzureClientMagicBase.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        public AzureClientMagicBase(IAzureClient azureClient, string keyword, Documentation docs):
+            base(keyword, docs)
+        {
+            this.AzureClient = azureClient;
```
efratshabtai, line 30: could be null

src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Microsoft.Quantum.IQSharp.AzureClient
+{
+    /// <summary>
+    ///     A magic command that can be used to connect to an Azure workspace.
+    /// </summary>
+    public class ConnectMagic : AzureClientMagicBase
+    {
+        private const string
+            ParamName_Login = "login",
```
efratshabtai, line 21: Not sure if this is part of the C# code conventions:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#naming-conventions
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constants

src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -0,0 +1,107 @@
...
+            ParamName_SubscriptionId = "subscriptionId",
+            ParamName_ResourceGroupName = "resourceGroupName",
+            ParamName_WorkspaceName = "workspaceName";
+
+        /// <summary>
+        ///     Constructs a new magic command given an IAzureClient object.
+        /// </summary>
+        public ConnectMagic(IAzureClient azureClient) :
+            base(azureClient,
+                "connect",
```
efratshabtai, line 32: Is there a disconnect? or do you call connect again if you want to connect to another workspace?

efratshabtai: Does connect return anything wen successful?
What errors can it return?
worth documenting these.

src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -0,0 +1,107 @@
...
+            var storageAccountConnectionString = inputParameters.DecodeParameter<string>(ParamName_StorageAccountConnectionString);
+            if (string.IsNullOrEmpty(storageAccountConnectionString))
+            {
+                return await AzureClient.PrintConnectionStatusAsync(channel);
+            }
+
+            var subscriptionId = inputParameters.DecodeParameter<string>(ParamName_SubscriptionId);
+            var resourceGroupName = inputParameters.DecodeParameter<string>(ParamName_ResourceGroupName);
+            var workspaceName = inputParameters.DecodeParameter<string>(ParamName_WorkspaceName);
+            var forceLogin = inputParameters.DecodeParameter<bool>(ParamName_Login);
```
efratshabtai, line 97: That's an optional parameter, right?
What happens in this line when it's not found - should we assign a default (false)?

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+        private AuthenticationResult? AuthenticationResult { get; set; }
+        private IQuantumClient? QuantumClient { get; set; }
+        private Azure.Quantum.Workspace? ActiveWorkspace { get; set; }
+
+        /// <summary>
+        /// Creates an AzureClient object.
+        /// </summary>
+        public AzureClient()
+        {
+            ConnectionString = string.Empty;
```
cgranade, line 33: Would it make sense to use an initializer on each of L22 and L23 instead to avoid needing an explicit constructor here? I left a suggestion above to that effect.

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
-using Microsoft.Quantum.Runtime;
+using Microsoft.Jupyter.Core;
+using Microsoft.Quantum.Simulation.Core;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
     /// <inheritdoc/>
     public class AzureClient : IAzureClient
     {
+        private string ConnectionString { get; set; }
```
cgranade, line 22:
```suggestion
        private string ConnectionString { get; set; } = string.Empty;
```

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+            QuantumClient.WorkspaceName = workspaceName;
+
+            ActiveWorkspace = new Azure.Quantum.Workspace(
+                QuantumClient.SubscriptionId, QuantumClient.ResourceGroupName,
+                QuantumClient.WorkspaceName, AuthenticationResult?.AccessToken);
+
+            try
+            {
+                var jobsList = await QuantumClient.Jobs.ListAsync();
+                channel.Stdout($"Found {jobsList.Count()} jobs in Azure Quantum workspace {workspaceName}");
```
cgranade, line 111:
```suggestion
                channel.Stdout($"Found {jobsList.Count()} jobs in Azure Quantum workspace {workspaceName}.");
```

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+            }
+
+            if (AuthenticationResult == null)
+            {
+                return AzureClientError.AuthenticationFailed.ToExecutionResult();
+            }
+
+            var credentials = new Rest.TokenCredentials(AuthenticationResult.AccessToken);
+
+            QuantumClient = new QuantumClient(credentials);
```
cgranade, line 99:
```suggestion
            QuantumClient = new QuantumClient(credentials)
            {
                SubscriptionId = subscriptionId,
                ResourceGroupName = resourceGroupName,
                WorkspaceName = workspaceName
            };
```
In general, I suggest using initialization syntax following constructors to help prepare for [C# 9 init-only properties](https://github.com/dotnet/csharplang/blob/master/proposals/init.md).

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+            {
+                channel.Stderr(e.ToString());
+                return AzureClientError.WorkspaceNotFound.ToExecutionResult();
+            }
+
+            return QuantumClient.ToJupyterTable().ToExecutionResult();
+        }
+
+        /// <inheritdoc/>
+        public async Task<ExecutionResult> PrintConnectionStatusAsync(IChannel channel)
```
cgranade, line 123: This could probably be an expression body method, using something like `QuantumClient == null ? AzureClientError.NotConnected.ToExecutionResult() :  QuantumClient.ToJupyterTable().ToExecutionResult()`.

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+            IChannel channel,
+            string subscriptionId,
+            string resourceGroupName,
+            string workspaceName,
+            string storageAccountConnectionString,
+            bool forceLogin = false)
+        {
+            ConnectionString = storageAccountConnectionString;
+
+            var clientId = "84ba0947-6c53-4dd2-9ca9-b3694761521b"; // Microsoft Quantum Development Kit
```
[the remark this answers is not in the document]
rmshaffer: Yes, this is a public app ID.

---
In reply to: [426982909](https://github.com/microsoft/iqsharp/pull/144#discussion_r426982909) [](ancestors = 426982909)

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+            }
+
+            var operationInfo = operationResolver.Resolve(operationName);
+            var entryPointInfo = new EntryPointInfo<QVoid, Result>(operationInfo.RoslynType);
+            var entryPointInput = QVoid.Instance;
+            var machine = Azure.Quantum.QuantumMachineFactory.CreateMachine(ActiveWorkspace, ActiveTargetName, ConnectionString);
+            if (machine == null)
+            {
+                channel.Stderr($"Could not find an execution target for target {ActiveTargetName}.");
+                return AzureClientError.NoTarget.ToExecutionResult();
```
cgranade, line 164: Just to double-check, should these calls to `ToExecutionResult` have an `ExecuteStatus.Error` parameter so as to raise the appropriate error in the client?
```suggestion
                return AzureClientError.NoTarget.ToExecutionResult(ExecuteStatus.Error);
```

src/AzureClient/Extensions.cs
```diff
@@ -22,5 +30,96 @@ public static void AddAzureClient(this IServiceCollection services)
         {
             services.AddSingleton<IAzureClient, AzureClient>();
         }
+
+        /// <summary>
+        ///      Encapsulates a given <c>AzureClientError</c> as the result of an execution.
```
cgranade, line 35: Minor, but `<see cref="AzureClientError" />` can be nicer than `<c></c>` alone for generating links.

src/AzureClient/Extensions.cs
```diff
@@ -22,5 +30,96 @@ public static void AddAzureClient(this IServiceCollection services)
...
+            return attributes?.Length > 0 ? attributes[0].Description : string.Empty;
+        }
+
+        /// <summary>
+        ///      Encapsulates a given <c>AzureClientError</c> as the result of an execution.
+        /// </summary>
+        /// <param name="task">
+        ///      A task which will return the result of an IAzureClient API call.
+        /// </param>
+        public static async Task<ExecutionResult> ToExecutionResult(this Task<AzureClientError> task) =>
```
cgranade, line 68: It may be good to add the optional `ExecuteStatus` parameter here as well.

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+                cacheDirectory = Path.Join(System.Environment.GetFolderPath(System.Environment.SpecialFolder.UserProfile), ".azure-quantum");
+            }
+
+            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, clientId).Build();
+            var cacheHelper = await MsalCacheHelper.CreateAsync(storageCreationProperties);
+            cacheHelper.RegisterCache(msalApp.UserTokenCache);
+
+            var scopes = new List<string>() { "https://quantum.microsoft.com/Jobs.ReadWrite" };
+
+            bool promptForLogin = forceLogin;
```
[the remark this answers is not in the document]
rmshaffer: I think they mean slightly different things. `forceLogin` indicates whether the user has explicitly asked for a login prompt. `promptForLogin` determines whether we actually show the prompt, i.e., `forceLogin` is set *or* silent authentication failed.

But yes, this naming could be better. I'll change to `forceLoginPrompt` and `shouldShowLoginPrompt`.

---
In reply to: [426983930](https://github.com/microsoft/iqsharp/pull/144#discussion_r426983930) [](ancestors = 426983930)

src/AzureClient/Extensions.cs
```diff
@@ -22,5 +30,96 @@ public static void AddAzureClient(this IServiceCollection services)
...
+            (await task).ToExecutionResult();
+
+        internal static Table<JobDetails> ToJupyterTable(this JobDetails jobDetails) =>
+            new List<JobDetails> { jobDetails }.ToJupyterTable();
+
+        internal static Table<JobDetails> ToJupyterTable(this IEnumerable<JobDetails> jobsList) =>
+            new Table<JobDetails>
```
cgranade, line 75: This may be difficult for a Python client to pick up and deserialize, since it looks like a list of the form `[{"JobId": jobId}, {"JobName", jobName}, ...]`. Given how frequently you're using object-as-a-table patterns here, it may be worth making a custom result encoder that adapts the JSON serialization output to your usecase.

src/AzureClient/AzureClient.cs
```diff
@@ -1,24 +1,249 @@
...
+
+        /// <inheritdoc/>
+        public async Task<ExecutionResult> SubmitJobAsync(
+            IChannel channel,
+            IOperationResolver operationResolver,
+            string operationName)
+        {
+            if (ActiveWorkspace == null)
+            {
+                channel.Stderr("Must call %connect before submitting a job.");
```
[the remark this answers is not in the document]
rmshaffer: Yes, that's nicer. All of these strings need to be reviewed with PM anyway, I've just put things in as placeholders. And all of the user-visible strings in this file should be moved to Resources.cs.

---
In reply to: [426986653](https://github.com/microsoft/iqsharp/pull/144#discussion_r426986653) [](ancestors = 426986653)

src/Jupyter/Extensions.cs
```diff
@@ -174,5 +175,17 @@ public static string Dedent(this string text)
...
         }
+
+        /// <summary>
+        ///      Retrieves and JSON-decodes the value for the given parameter name.
+        /// </summary>
+        public static T DecodeParameter<T>(this Dictionary<string, string> parameters, string parameterName)
+        {
+            if (!parameters.TryGetValue(parameterName, out string parameterValue))
+            {
+                return default(T);
```
cgranade, line 186: As per @efratshabtai's comment above, maybe an optional parameter here specifying a default other than `default(T)` would be helpful?

src/Jupyter/Extensions.cs
```diff
@@ -174,5 +175,17 @@ public static string Dedent(this string text)
...
+            }
+            return (T)(JsonConvert.DeserializeObject(parameterValue) ?? default(T));
```
cgranade, line 188: I'd suggest adding `#nullable enable` to this file to make sure that you've covered all possible paths for nullable references here.

src/AzureClient/IAzureClient.cs
rmshaffer: Yes, it would be saved locally. I could imagine that maybe we'd also want to do something like retrieve the target's current status (e.g. availability, avg. queue time, etc.).

But in fact, this function isn't even used anywhere at the moment, and there's no scenario in the spec that requires it. I'll just delete it.

---
In reply to: [426993539](https://github.com/microsoft/iqsharp/pull/144#discussion_r426993539) [](ancestors = 426993539)

rmshaffer: See related comment - deleting this function.

---
In reply to: [426993312](https://github.com/microsoft/iqsharp/pull/144#discussion_r426993312) [](ancestors = 426993312)

src/Python/qsharp/azure.py
```diff
@@ -0,0 +1,47 @@
...
+__all__ = [
+    'connect',
+    'target',
+    'submit',
+    'status'
+]
+
+## FUNCTIONS ##
+
+def connect(**params) -> Any:
```
cgranade, line 37: I think this makes sense for now, but we will likely want to narrow this from `Any` before release.

src/Tests/AzureClientMagicTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                   resourceGroupName={resourceGroupName}
+                   workspaceName={workspaceName}
+                   storageAccountConnectionString={storageAccountConnectionString}");
+            Assert.AreEqual(azureClient.LastAction, AzureClientAction.Connect);
+            Assert.IsFalse(azureClient.ForceLogin);
+            Assert.AreEqual(azureClient.ConnectionString, storageAccountConnectionString);
+
+            // valid input with forced login
+            connectMagic.Test(
+                @$"login
```
cgranade, line 58: This line is getting highlighted weird by GitHub; there may be an invisible special character here?

src/AzureClient/Magic/AzureClientMagicBase.cs
rmshaffer: Yes, the keyword is used to invoke the magic command (e.g. 'connect'), and the documentation describes how to use the magic command. Sure, I'll elaborate a bit here.

---
In reply to: [426993988](https://github.com/microsoft/iqsharp/pull/144#discussion_r426993988) [](ancestors = 426993988)

src/AzureClient/Magic/AzureClientMagicBase.cs
rmshaffer: In general, the magic command classes receive their arguments from the `DependencyInjection` infrastructure (see `Startup.ConfigureServices()` in the `Tool` project), which keeps singletons of many objects, including `AzureClient`, and automatically provides references to those singletons when constructing the magic commands.

All that said, in general if this is null then something very unexpected went wrong at startup, so throwing an exception in that case would be fine. In general, IQ# code doesn't null-check the parameters coming from the `DependencyInjection` framework, because they should never be null.

---
In reply to: [426994603](https://github.com/microsoft/iqsharp/pull/144#discussion_r426994603) [](ancestors = 426994603)

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: Yes, you're right, underscore is frowned upon, I'll remove it.

---
In reply to: [426996582](https://github.com/microsoft/iqsharp/pull/144#discussion_r426996582) [](ancestors = 426996582)

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: Yes, you would just call connect again. We discussed disconnect but decided there wasn't really a scenario where it was needed.

---
In reply to: [426996900](https://github.com/microsoft/iqsharp/pull/144#discussion_r426996900) [](ancestors = 426996900)

rmshaffer: The return value of these magic commands is always an `ExecutionResult`, which indicates whether it succeeded and also (optionally) contains some information, either about the information that was retrieved, or about the error that occurred.

Yes, when switching these to all be subcommands of `%aq`, I'll need to revisit the return behavior and documentation for all of the commands to make sure they're consistent.

---
In reply to: [426997290](https://github.com/microsoft/iqsharp/pull/144#discussion_r426997290) [](ancestors = 426997290)

src/Jupyter/Extensions.cs
rmshaffer: Agreed, I'll add this.

---
In reply to: [427476716](https://github.com/microsoft/iqsharp/pull/144#discussion_r427476716) [](ancestors = 427476716)

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: Yes, `DecodeParameter` implicitly assigns a default value of false here if it's not found, but as Chris also suggested, I'll add an explicit parameter here and explicitly specify that the default value of this one is false.

---
In reply to: [426997830](https://github.com/microsoft/iqsharp/pull/144#discussion_r426997830) [](ancestors = 426997830)

rmshaffer:
I imagine that users who aren't part of the preview won't be able to create Azure Quantum workspaces, and therefore they won't be able to pass a valid workspace name to `%connect` (or, later, `%aq connect`). We do make clear in the resulting error message that the specified workspace must be an Azure Quantum workspace.

---
In reply to: [414045488](https://github.com/microsoft/iqsharp/pull/144#issuecomment-414045488) [](ancestors = 414045488)

src/AzureClient/AzureClient.cs
rmshaffer: I learn something new about C# every day...

---
In reply to: [427460618](https://github.com/microsoft/iqsharp/pull/144#discussion_r427460618) [](ancestors = 427460618)

src/AzureClient/AzureClient.cs
cgranade: It looks like the `#Pending` comment got edited into the suggestion somehow?

src/Python/qsharp/azure.py
rmshaffer: Yes, as Efrat also pointed out, we need to figure out what information these magic commands should return as part of the `ExecutionResult`.

---
In reply to: [427478579](https://github.com/microsoft/iqsharp/pull/144#discussion_r427478579) [](ancestors = 427478579)

src/AzureClient/AzureClient.cs
rmshaffer: ah. I'm using CodeFlow to read through these comments and it has no idea how to deal with suggestions.

---
In reply to: [427496055](https://github.com/microsoft/iqsharp/pull/144#discussion_r427496055) [](ancestors = 427496055)

cgranade: Ah, makes sense! I've always just used the built-in GitHub UI myself, but that makes a lot of sense.

src/AzureClient/Extensions.cs
rmshaffer: Yes, this is a very good point. This will need to change.

---
In reply to: [427472915](https://github.com/microsoft/iqsharp/pull/144#discussion_r427472915) [](ancestors = 427472915)

src/AzureClient/AzureClient.cs
rmshaffer: It does, but also, it's a good reason to stick with the GitHub UI. This isn't the only weird thing. CodeFlow is nice for reading code, but I may abandon it for commenting.

---
In reply to: [427498865](https://github.com/microsoft/iqsharp/pull/144#discussion_r427498865) [](ancestors = 427498865)

src/AzureClient/AzureClient.cs
rmshaffer: The `AzureClientError.ToExecutionResult` extension function automatically sets `ExecuteStatus.Error` in the `ExecutionResult` object. I think that makes sense, since an `AzureClientError` is always an error?

---
In reply to: [427470175](https://github.com/microsoft/iqsharp/pull/144#discussion_r427470175) [](ancestors = 427470175)

src/AzureClient/Extensions.cs
rmshaffer: This ends up calling the `AzureClientError.ToExecutionResult` extension, which always sets `ExecuteStatus.Error`. I'm not sure it makes sense to allow a different status for an `AzureClientError`?

---
In reply to: [427471505](https://github.com/microsoft/iqsharp/pull/144#discussion_r427471505) [](ancestors = 427471505)

src/AzureClient/Extensions.cs
rmshaffer: Thanks, I forgot about that syntax.

---
In reply to: [427470952](https://github.com/microsoft/iqsharp/pull/144#discussion_r427470952) [](ancestors = 427470952)

src/AzureClient/AzureClient.cs
cgranade: That make sense, I had just missed that overriding of `ToExecutionResult()this object)`. Thanks for the explanation!

src/AzureClient/AzureClient.cs
rmshaffer: Yes, good catch, but actually there's no reason to print the count of jobs here. We should just print a message saying successfully connected. (Getting the jobs list was the recommended way to verify that we can connect to the workspace.)

---
In reply to: [427463264](https://github.com/microsoft/iqsharp/pull/144#discussion_r427463264) [](ancestors = 427463264)

rmshaffer:
@efratshabtai @cgranade Apologies for all of the messy comments and replies, I was using CodeFlow to respond to these comments and it apparently adds hashtags to *your* comments when I change the status of them, creates orphaned replies, doesn't handle suggestions appropriately, etc... I'll be sticking to GitHub for PR comments from now on...

src/AzureClient/AzureClient.cs
rmshaffer: Thanks! I applied your suggestion, but CodeFlow wasn't showing me that this was actually a suggestion, so I didn't use the commit-suggestion feature. I'll give you credit through a mention here: @cgranade   :-)

src/Tests/AzureClientMagicTests.cs
rmshaffer: VS Code is doing this too. Apparently it thinks it's an unclosed string literal (same with the one a few lines down). If I put `subscriptionId` on the same line then it goes away. I'll do that just to avoid weirdness.

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: I've opened a task to track this work.

src/AzureClient/Extensions.cs
rmshaffer: I've opened a task to track this work.

rmshaffer:
> Something to think about is what would the experience be for users that are not part of the private preview.

I imagine that users who aren't part of the preview won't be able to create Azure Quantum workspaces, and therefore they won't be able to pass a valid workspace name to %connect (or, later, %aq connect). We do make clear in the resulting error message that the specified workspace must be an Azure Quantum workspace.

PR #150, opened 2020-05-27
Add remaining AzureClient magic commands

rmshaffer:
This PR primarily includes the code to add the remaining AzureClient magic commands: `execute`, `jobs`, and `output`.

It also includes a few related/supporting changes:
- Rename all AzureClient magic commands to `%azure.*`
- Rename `login` parameter to `refresh` for magic `%azure.connect` 
- Add `AzureEnvironment` class with support for Dogfood, Canary, or Production environments
- Retrieve and cache provisioned provider list when connecting to the workspace
- Add a number of TODO comments to indicate work that will be tracked and completed before merging to master

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -15,6 +15,8 @@ using Microsoft.Jupyter.Core;
 using Microsoft.Quantum.Simulation.Core;
 using Microsoft.Rest.Azure;
 using Microsoft.Azure.Quantum.Client.Models;
+using Microsoft.Azure.Quantum.Storage;
+using Microsoft.Azure.Quantum;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
@@ -27,6 +29,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         private IQuantumClient? QuantumClient { get; set; }
         private IPage<ProviderStatus>? ProviderStatusList { get; set; }
         private Azure.Quantum.IWorkspace? ActiveWorkspace { get; set; }
+        private string MostRecentJobId { get; set; } = string.Empty;
 
         /// <inheritdoc/>
         public async Task<ExecutionResult> ConnectAsync(
@@ -35,13 +38,18 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             string resourceGroupName,
             string workspaceName,
             string storageAccountConnectionString,
-            bool forceLoginPrompt = false)
+            bool refreshCredentials = false)
         {
             ConnectionString = storageAccountConnectionString;
 
-            var clientId = "84ba0947-6c53-4dd2-9ca9-b3694761521b"; // Microsoft Quantum Development Kit
-            var authority = "https://login.microsoftonline.com/common";
-            var msalApp = PublicClientApplicationBuilder.Create(clientId).WithAuthority(authority).Build();
+            var azureEnvironmentEnvVarName = "AZURE_QUANTUM_ENV";
+            var azureEnvironmentName = System.Environment.GetEnvironmentVariable(azureEnvironmentEnvVarName);
+            var azureEnvironment = AzureEnvironment.Create(azureEnvironmentName, subscriptionId);
+
+            var msalApp = PublicClientApplicationBuilder
+                .Create(azureEnvironment.ClientId)
+                .WithAuthority(azureEnvironment.Authority)
+                .Build();
 
             // Register the token cache for serialization
             var cacheFileName = "aad.bin";
@@ -52,20 +60,18 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 cacheDirectory = Path.Join(System.Environment.GetFolderPath(System.Environment.SpecialFolder.UserProfile), ".azure-quantum");
             }
 
-            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, clientId).Build();
+            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, azureEnvironment.ClientId).Build();
             var cacheHelper = await MsalCacheHelper.CreateAsync(storageCreationProperties);
             cacheHelper.RegisterCache(msalApp.UserTokenCache);
 
-            var scopes = new List<string>() { "https://quantum.microsoft.com/Jobs.ReadWrite" };
-
-            bool shouldShowLoginPrompt = forceLoginPrompt;
+            bool shouldShowLoginPrompt = refreshCredentials;
             if (!shouldShowLoginPrompt)
-            { 
+            {
                 try
                 {
                     var accounts = await msalApp.GetAccountsAsync();
                     AuthenticationResult = await msalApp.AcquireTokenSilent(
-                        scopes, accounts.FirstOrDefault()).WithAuthority(msalApp.Authority).ExecuteAsync();
+                        azureEnvironment.Scopes, accounts.FirstOrDefault()).WithAuthority(msalApp.Authority).ExecuteAsync();
                 }
                 catch (MsalUiRequiredException)
                 {
@@ -75,7 +81,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
 
             if (shouldShowLoginPrompt)
             {
-                AuthenticationResult = await msalApp.AcquireTokenWithDeviceCode(scopes,
+                AuthenticationResult = await msalApp.AcquireTokenWithDeviceCode(
+                    azureEnvironment.Scopes,
                     deviceCodeResult =>
                     {
                         channel.Stdout(deviceCodeResult.Message);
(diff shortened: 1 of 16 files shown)


src/AzureClient/AzureEnvironment.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                ClientId = "46a998aa-43d0-4281-9cbb-5709a507ac36",      // QDK dogfood client ID
+                Authority = GetDogfoodAuthority(subscriptionId),
+                Scopes = new List<string>() { "api://dogfood.azure-quantum/Jobs.ReadWrite" },
+                BaseUri = new Uri("https://app-jobscheduler-test.azurewebsites.net/"),
+            };
+
+        private static AzureEnvironment Canary()
+        {
+            var canary = Production();
```
cgranade, line 52: This makes me really look forward to C# 9's `with` keyword.

src/AzureClient/AzureEnvironment.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                var authHeader = response.Headers["WWW-Authenticate"];
+                var headerParts = authHeader.Substring("Bearer ".Length).Split(',');
+                foreach (var headerPart in headerParts)
+                {
+                    var parts = headerPart.Split("=", 2);
+                    if (parts[0] == "authorization_uri")
+                    {
+                        var quotedAuthority = parts[1];
+                        return quotedAuthority[1..^1];
```
cgranade, line 82: Oh, nice, Index!

src/AzureClient/AzureEnvironment.cs
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                }
+            }
+            catch (Exception)
```
cgranade, line 86: It looks like this discards any of the diagnostics included by the caught exception; would it be possible to at least log it as a warning?

src/AzureClient/AzureEnvironment.cs
```diff
@@ -0,0 +1,93 @@
...
+                    response = WebRequest.Create(requestUrl).GetResponse();
+                }
+                catch (WebException webException)
+                {
+                    response = webException.Response;
+                }
+
+                var authHeader = response.Headers["WWW-Authenticate"];
+                var headerParts = authHeader.Substring("Bearer ".Length).Split(',');
+                foreach (var headerPart in headerParts)
```
cgranade, line 76: It looks like if this loop completes without returning, an exception is thrown; would it be possible to make the message more specific to indicate that the right header wasn't found?

src/AzureClient/IAzureClient.cs
```diff
@@ -65,19 +65,21 @@ public interface IAzureClient
     {
         /// <summary>
         /// Connects to the specified Azure Quantum workspace, first logging into Azure if necessary.
+        /// Returns the list of execution targets available in the Azure Quantum workspace.
```
cgranade, line 68: Very minor, but this may be good to include in a `<returns>` element.

src/AzureClient/Magic/ExecuteMagic.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        ///     Executes a new job to an Azure Quantum workspace given a Q# operation
+        ///     name that is present in the current Q# Jupyter workspace, and
+        ///     waits for the job to complete before returning.
+        /// </summary>
+        public override async Task<ExecutionResult> RunAsync(string input, IChannel channel)
+        {
+            Dictionary<string, string> keyValuePairs = ParseInputParameters(input);
+            var operationName = keyValuePairs.Keys.FirstOrDefault();
```
cgranade, line 71: Does this depend on the `keyValuePairs` dictionary having exactly one pair? If so, I'd suggest `Single()` instead of `FirstOrDefault()`.

src/AzureClient/Magic/JobsMagic.cs
```diff
@@ -0,0 +1,60 @@
...
+                            Out[]: <list of jobs in the workspace>
+                            ```
+                        ".Dedent(),
+                    },
+                }) {}
+
+        /// <summary>
+        ///     Lists all jobs in the active workspace.
+        /// </summary>
+        public override async Task<ExecutionResult> RunAsync(string input, IChannel channel)
```
cgranade, line 55: Very minor, but this could be a good example of an expression body method.

src/AzureClient/AzureClient.cs
```diff
@@ -32,13 +38,18 @@ public class AzureClient : IAzureClient
...
             string storageAccountConnectionString,
-            bool forceLoginPrompt = false)
+            bool refreshCredentials = false)
         {
             ConnectionString = storageAccountConnectionString;
 
-            var clientId = "84ba0947-6c53-4dd2-9ca9-b3694761521b"; // Microsoft Quantum Development Kit
-            var authority = "https://login.microsoftonline.com/common";
-            var msalApp = PublicClientApplicationBuilder.Create(clientId).WithAuthority(authority).Build();
+            var azureEnvironmentEnvVarName = "AZURE_QUANTUM_ENV";
```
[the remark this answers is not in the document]
rmshaffer: No, this is purely optional. Default is Production.

src/AzureClient/AzureEnvironment.cs
rmshaffer: I also had that thought while I was typing this. 👍

src/AzureClient/AzureEnvironment.cs
rmshaffer: I just learned this syntax from a code action.

src/AzureClient/AzureEnvironment.cs
```diff
@@ -0,0 +1,93 @@
...
+        public string ClientId { get; private set; } = string.Empty;
+        public string Authority { get; private set; } = string.Empty;
+        public List<string> Scopes { get; private set; } = new List<string>();
+        public Uri? BaseUri { get; private set; }
+
+        private AzureEnvironment()
+        {
+        }
+
+        public static AzureEnvironment Create(string environment, string subscriptionId) =>
```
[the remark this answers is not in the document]
rmshaffer: In principle, yes, although the caller is getting this value from an environment variable, so at some point we have to convert the string to the enum value. And I think we would want that logic also to be encapsulated inside this class. So I'm not sure it adds much other than a layer of indirection inside this file.

src/AzureClient/Magic/ExecuteMagic.cs
rmshaffer: Thanks, I didn't know about `Single()`! Actually, I will soon be adding support for additional parameters to `%azure.submit` and `%azure.execute`, so `keyValuePairs` may have multiple entries.

But you've made me realize that this is incorrect, and I shouldn't be relying on the key name here; instead I should be using the `firstParameterInferredName` option that I had previously implemented in `src/Kernel/Magic/Simulate.cs` for cases like this. And then the full set of parameters should be passed into the `IAzureClient` call. I'll fix that.

src/AzureClient/AzureEnvironment.cs
rmshaffer: Yes, thanks for catching my laziness. I'll log something here and/or include the details in the exception that is thrown later.

src/AzureClient/AzureEnvironment.cs
rmshaffer: Yes, I'll do that.

src/AzureClient/IAzureClient.cs
rmshaffer: Thanks, I'll update all the comments in this file.

src/AzureClient/AzureClient.cs
```diff
@@ -160,63 +220,109 @@ public class AzureClient : IAzureClient
...
+                // TODO: Add encoder for CloudJob rather than calling ToJupyterTable() here directly.
+                return job.Details.ToJupyterTable().ToExecutionResult();
+            }
+
+            var stream = new MemoryStream();
+            var protocol = await new JobStorageHelper(ConnectionString).DownloadJobOutputAsync(jobId, stream);
+            stream.Seek(0, SeekOrigin.Begin);
+            var outputJson = new StreamReader(stream).ReadToEnd();
+
+            // TODO: Deserialize this once we have a way of getting the output type
```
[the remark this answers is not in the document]
rmshaffer: It's not worth it at the moment; displaying the output correctly is one of the next changes I'll be making.

rmshaffer:
> I think this looks really good, just a few minor comments here and there.

@cgranade Thanks for the thorough review!

src/AzureClient/AzureClient.cs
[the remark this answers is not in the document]
cgranade: Makes sense!

src/AzureClient/AzureEnvironment.cs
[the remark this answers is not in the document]
cgranade: Makes sense. As one last thought, then, would Enum.Parse be reasonable?

cgranade:
> > I think this looks really good, just a few minor comments here and there.
> 
> @cgranade Thanks for the thorough review!

No worries, happy to take a look!

src/AzureClient/AzureEnvironment.cs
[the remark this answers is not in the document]
rmshaffer: Yes, that's a good idea and is very simple.

PR #151, opened 2020-05-29
Improvements to CodeMirror syntax highlighting rules

### LINKED ISSUE #77: IQ# syntax highlighting errors
**Describe the bug**
Syntax highlighting in IQ# exhibits several issues:
* Type names at the end of operation names are highlighted as types
* Characters before reserved words are highlighted as part of those words
* Reserved words at the beginning of the line (without whitespace before them) are not highlighted 

**To Reproduce**
Add the following Q# code to a Q# notebook:

    operation SyntaxHighlight () : Unit {
    
    // Type names at the end of operation names are highlighted
    ControlledOnInt();
    AssertQubit();
    ApplyPauliFromBitString(PauliX, true, negate, inputs);
    
    // Characters before reserved words are highlighted as part of those words
    {let b =Length(array);}
    ApplyToEach(X, array);
    ApplyPauliFromBitString(PauliX,true, negate, inputs);
    let array = [false, false];
    
    // Reserved words at the beginning of the line (without whitespace before them) are not highlighted
    let a = Length();
    Adjoint Z(qubit);
    Message("Hi");
    }

**Screenshots**

image

**System information**
 - OS: Windows 10
 - Browser: Firefox 

**Additional context**
image

rmshaffer:
This PR partially addresses #77 by updating CodeMirror syntax highlighting rules:
- The main improvement here is to use the `sol` property for [CodeMirror simple mode](https://codemirror.net/demo/simplemode.html) rules, which works around a limitation that prevents the standard `^` start-of-line regex marker from working as expected.
- This PR also changes the syntax highlighting for magic commands such that only the magic command token itself is highlighted as a "builtin" token, and not as a "comment" (which highlights the full line). This is in response to recent behavioral changes such that magic command arguments can span multiple lines, and the "comment" style doesn't play nicely with multi-line arguments.

The remaining bug from #77 is that a single non-word character immediately preceding a keyword still gets highlighted as part of the keyword. This is existing behavior, and I haven't found a way to fix this within the confines of CodeMirror simple mode. We would have to implement a custom mode, which would give us a lot more flexibility but would also be a much larger task.

But despite this remaining bug, the changes here fix the more glaring issues such as portions of non-keywords being highlighted, or failure to highlight keywords that are at the beginning of a line, as reported in the issue.

### DIFF
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -70,12 +70,6 @@ function defineQSharpMode() {
             beginWord: true,
         },
         {
-            // Azure magic commands
-            token: "builtin",
-            regex: String.raw`(%azure\.(connect|execute|jobs|output|status|submit|target))\b`,
-            beginWord: true,
-        },
-        {
             // chemistry magic commands
             token: "builtin",
             regex: String.raw`(%chemistry\.(broombridge|encode|fh\.add_terms|fh\.load|inputstate\.load))\b`,
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -11,54 +11,106 @@ import { Telemetry, ClientInfo } from "./telemetry.js";
 
 function defineQSharpMode() {
     console.log("Loading IQ# kernel-specific extension...");
+
+    let rules = [
+        {
+            token: "comment",
+            regex: /(\/\/).*/,
+            beginWord: false,
+        },
+        {
+            token: "string",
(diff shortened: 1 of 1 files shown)


tcNickolas:
Could you please attach a screenshot showing the fixed behavior, similar to the one in #77?

rmshaffer:
> Could you please attach a screenshot showing the fixed behavior, similar to the one in #77?

Yes, good idea! Here it is:

image

PR #152, opened 2020-06-01
Validate and load packages for Azure execution targets

rmshaffer:
This PR enables Azure Quantum job execution by validating and loading the required provider packages in response to an `%azure.target` call.

It also includes a few other minor improvements:
- Add basic deserialization of output histogram to display job results.
- Update Python interface to match new magic command names.
- Add syntax highlighting for %azure.* magic commands.
- Add some `channel.Stdout` calls to print helpful information during magic command execution.

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -277,6 +277,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
 
             // Set the active target and load the package.
             ActiveTarget = executionTarget;
+
+            channel.Stdout($"Loading package {ActiveTarget.PackageName} and dependencies...");
             await references.AddPackage(ActiveTarget.PackageName);
 
             return $"Active target is now {ActiveTarget.TargetName}".ToExecutionResult();
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -201,10 +201,9 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             {
                 channel.Stdout($"Submitting {operationName} to target {ActiveTarget.TargetName}...");
                 var job = await machine.SubmitAsync(entryPointInfo, entryPointInput);
+                channel.Stdout($"Job {job.Id} submitted successfully.");
+
                 MostRecentJobId = job.Id;
-                channel.Stdout("Job submission successful.");
-                channel.Stdout($"To check the status, run:\n    %azure.status {MostRecentJobId}");
-                channel.Stdout($"To see the results, run:\n    %azure.output {MostRecentJobId}");
 
                 // TODO: Add encoder for IQuantumMachineJob rather than calling ToJupyterTable() here.
                 return job.ToJupyterTable().ToExecutionResult();
@@ -314,9 +313,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
 
             if (!job.Succeeded || string.IsNullOrEmpty(job.Details.OutputDataUri))
             {
-                channel.Stderr($"Job ID {jobId} has not completed. Displaying the status instead.");
-                // TODO: Add encoder for CloudJob rather than calling ToJupyterTable() here directly.
-                return job.Details.ToJupyterTable().ToExecutionResult();
+                channel.Stderr($"Job ID {jobId} has not completed. To check the status, use:\n   %azure.status {jobId}");
+                return AzureClientError.JobNotCompleted.ToExecutionResult();
             }
 
             var stream = new MemoryStream();
@@ -363,8 +361,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 return AzureClientError.JobNotFound.ToExecutionResult();
             }
 
-            // TODO: Add encoder for CloudJob rather than calling ToJupyterTable() here directly.
-            return job.Details.ToJupyterTable().ToExecutionResult();
+            // TODO: Add encoder for CloudJob which calls ToJupyterTable() for display.
+            return job.Details.ToExecutionResult();
         }
 
         /// <inheritdoc/>
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -17,6 +17,8 @@ using Microsoft.Rest.Azure;
 using Microsoft.Azure.Quantum.Client.Models;
 using Microsoft.Azure.Quantum.Storage;
 using Microsoft.Azure.Quantum;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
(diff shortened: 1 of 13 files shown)


src/AzureClient/AzureClient.cs
```diff
@@ -262,13 +320,18 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
...
 
             var stream = new MemoryStream();
-            var protocol = await new JobStorageHelper(ConnectionString).DownloadJobOutputAsync(jobId, stream);
+            await new JobStorageHelper(ConnectionString).DownloadJobOutputAsync(jobId, stream);
             stream.Seek(0, SeekOrigin.Begin);
-            var outputJson = new StreamReader(stream).ReadToEnd();
+            var output = new StreamReader(stream).ReadToEnd();
+            var deserializedOutput = JsonConvert.DeserializeObject<Dictionary<string, JToken>>(output);
+            var histogram = new Dictionary<string, double>();
+            foreach (var entry in deserializedOutput["histogram"] as JObject)
```
cgranade, line 328: In other places, I've seen histograms serialized as lists of alternating keys and values where keys are lists of ints, since JSON doesn't support complex mapping keys, but this looks like it uses a string serialization of keys instead? Is that specific to some targets, or is there a conversion happening somewhere? Thanks for the help!

src/AzureClient/AzureClient.cs
```diff
@@ -210,19 +229,58 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
...
+            if (executionTarget == null)
+            {
+                channel.Stderr($"Target name {targetName} does not support executing Q# jobs.");
+                channel.Stdout($"Available execution targets: {ValidExecutionTargetsDisplayText}");
+                return AzureClientError.InvalidTarget.ToExecutionResult();
+            }
+
+            // Set the active target and load the package.
+            ActiveTarget = executionTarget;
+            await references.AddPackage(ActiveTarget.PackageName);
```
cgranade, line 281: Normally the `%package` magic prints out a lot of status information as it goes, since package loading can take a while on slow connections or when the NuGet cache needs to be populated for the first time. Would it be worth making transient displays here to help communicate to users that a package is being loaded?

src/AzureClient/AzureClient.cs
rmshaffer: From what I can tell, these histograms are always serialized in this way, where the keys are strings and the values are floats. @anpaz-msft, do you know if there is a better/more robust way to extract the histogram from the job output than what I am doing here? I don't see anything in the job definition that tells me how the output is serialized. There is the `OutputDataFormat` property, but that is just a string.

src/AzureClient/AzureClient.cs
rmshaffer: Yes, that's a good suggestion. Ideally I'd like to just call the `%package` magic directly here, but both it and the `IMagicSymbolResolver` are in the `Kernel` project and I can't access them directly.

I'll do the bare minimum for now and add a line of output indicating that the package and its dependencies are being loaded.

cgranade: Makes sense. Perhaps it would be good to put a task on the backlog to refactor and consolidate the two package load calls, then?

rmshaffer: Sounds good! I've added a task.

src/AzureClient/AzureExecutionTarget.cs
```diff
@@ -0,0 +1,35 @@
...
+        public string PackageName => $"Microsoft.Quantum.Providers.{GetProvider(TargetName)}";
+
+        public static bool IsValid(string targetName) => GetProvider(targetName) != null;
+
+        public static AzureExecutionTarget? Create(string targetName) =>
+            IsValid(targetName)
+            ? new AzureExecutionTarget() { TargetName = targetName }
+            : null;
+
+        private static AzureProvider? GetProvider(string targetName)
```
cesarzc, line 24: It would be good to add as a comment an example or a few examples of `targetName` strings.

src/AzureClient/Extensions.cs
```diff
@@ -90,7 +90,6 @@ public static string ToDescription(this AzureClientError azureClientError)
                     {
                         ("JobId", job => job.Id),
                         ("JobStatus", job => job.Status),
-                        ("JobUri", job => job.Uri.ToString()),
```
cesarzc, line 93: Why is the job's URI no longer included?

rmshaffer: The `Uri` property of `CloudJob` just throws a `NotImplementedException`: https://github.com/microsoft/qsharp-runtime/blob/4fba91a886ebededb75c17e6d6e05cf08d1e8c97/src/Azure/Azure.Quantum.Client/JobManagement/CloudJob.cs#L63

src/AzureClient/AzureExecutionTarget.cs
rmshaffer: Thanks, I've added documentation to this function for clarity!

src/AzureClient/Extensions.cs
cesarzc: I see, that is one of the tasks that we still have to do in qsharp-runtime. The issue tracking that work is the following: https://github.com/microsoft/qsharp-runtime/issues/236

Should an issue be created in this repo to update this extension once the issue in qsharp-runtime is resolved?

rmshaffer: Yes, I actually have a task in DevOps to update these extensions for better display. I will add this note to that task and link to the issue you referenced. Thanks!

cesarzc: Thanks!

src/AzureClient/AzureClient.cs
rmshaffer: @cesarzc, thanks for reviewing this PR! One last question for you: Do you have any opinion about the histogram deserialization discussed here?

cesarzc: This is expected to be consistent across all providers. In this case, the serialization is done the way @cgranade described it, as a list of alternating keys and values.

A job output is a serialized bond data structures. In this case the specific data structure would be `QSharpExecutionOutput` and it is part of the `Microsoft.Quantum.Providers.Shared.Datastructures.v1` namespace.

Using bond would be the preferred way to deserialize from the stream. The code to do it would look something like the following:
```csharp
var simpleJsonReader = new SimpleJsonReader(stream);
var deserializedOutput = Deserialize<QSharpExecutionOutput>.From(simpleJsonReader);
```

In case I missed something, you can find more details about Bond here: https://microsoft.github.io/bond/manual/bond_cs.html

rmshaffer: Thanks @cesarzc! But actually that's not the format I get here, since I'm getting the result from storage. `JobStorageHelper.DownloadJobOutputAsync` gives me a stream that looks something like:
```
{"duration":10758761,"histogram":{"0":0.2499999999999999,"1":0.2499999999999999,"2":0.2499999999999999,"3":0.2499999999999999}}
```
Is it expected that this format is different than the `QSharpExecutionOutput` that you describe? Is it safe to assume that this also would be consistent across all providers?

cesarzc: That's interesting, is this a QIO or Honeywell job?

I just checked an IonQ job, and its output was the following:
```
{"Histogram":["[0,0]",0.4999999999999999,"[1,1]",0.4999999999999999]}
```

rmshaffer: Mine is also an IonQ job. I think the difference is whether the job output comes directly from `IQuantumMachine.ExecuteAsync`, which is what you're showing, or whether the result is retrieved later from storage, which is what I'm showing.

rmshaffer: Followed up with @cesarzc offline. It turns out that there is a small bug in qsharp-runtime (at [this line](https://github.com/microsoft/qsharp-runtime/blob/c8f23017800974814f19d8ef005fb9786cd734bb/src/Azure/Azure.Quantum.Client/Storage/JobStorageHelper.cs#L122)) which is causing me to get the `"rawOutputData"` instead of the translated `"outputData"` for the job. I should indeed be getting the format @cgranade mentioned, and should be able to deserialize it as @cesarzc described.

I'll send a new PR for consuming that fix and updating the deserialization code here, once it's available.

rmshaffer: I've opened https://github.com/microsoft/qsharp-runtime/issues/239 to track this.

PR #154, opened 2020-06-03
Use SearchOption.TopDirectoryOnly for workspace cache

### LINKED ISSUE #149: IQ# kernel fails to start when "Application Data" folder is inaccessible
Originally opened by @n33r4j:

I recently updated my iqsharp(from 0.10 to 0.11) for jupyter notebooks (via Anaconda) and ever since, the Q# kernel stopped connecting when I open a new or existing Q# notebook(Python works fine). I have tried uninstalling and reinstalling jupyter notebook as well as iqsharp but nothing seems to have changed.
I think it may have something to do with "Access to the path 'C:\Users\neera\Application Data' is denied".

rmshaffer:
This PR should fix #149 by ensuring that we only look in the current folder for .qs files when checking the freshness of the workspace cache.

This is related to a similar change a while ago by PR #49 to address issue #20. That change updated the behavior to only actually load .qs files from the current folder into the workspace. But because the cache-freshness check was still enumerating all subfolders, any folders with permissions issues could still cause errors during kernel initialization.

I have not yet been able to reproduce the actual bug here, despite many attempts at various permissions-related things. If anyone can find concrete repro steps, that would help with the confidence in the fix.

### DIFF
--- a/src/Core/Workspace/Workspace.cs
+++ b/src/Core/Workspace/Workspace.cs
@@ -78,6 +78,11 @@ namespace Microsoft.Quantum.IQSharp
         public string Root { get; set; }
 
         /// <summary>
+        /// Gets the source files to be built for this Workspace.
+        /// </summary>
+        public IEnumerable<string> SourceFiles => Directory.EnumerateFiles(Root, "*.qs", SearchOption.TopDirectoryOnly);
+
+        /// <summary>
         /// Information about the assembly built from this Workspace.
         /// </summary>
         public AssemblyInfo AssemblyInfo { get; set; }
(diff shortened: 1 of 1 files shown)


src/Core/Workspace/Workspace.cs
```diff
@@ -77,6 +77,11 @@ public string CacheFolder
         /// </summary>
         public string Root { get; set; }
 
+        /// <summary>
+        /// Gets the source files to be built for this Workspace.
+        /// </summary>
+        public IEnumerable<string> SourceFiles => Directory.EnumerateFiles(Root, "*.qs", SearchOption.TopDirectoryOnly);
```
rmshaffer, line 83: Ordinarily I'd make this private, but I'm making it public here because I'll be using it in a subsequent PR.

PR #156, opened 2020-06-03
Auto-generate entry point for Azure job submission

rmshaffer:
This PR contains the code to auto-generate an `@EntryPoint()` operation for Azure Quantum job submission from IQ#. This operation then gets compiled into a separate executable assembly that references the Workspace and Snippets assemblies.

The key places to start reviewing would be:
- new function `BuildEntryPoint()` in `src/Core/Compiler/CompilerService.cs`
- new class `EntryPointGenerator` in `src/AzureClient/EntryPoint/EntryPointGenerator.cs`
- new class `EntryPoint` in `src/AzureClient/EntryPoint/EntryPoint.cs`
- consuming function `SubmitOrExecuteJobAsync()` in `src/AzureClient/AzureClient.cs`

Fixes [AB#13591](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/13591), fixes [AB#15418](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15418).


src/AzureClient/AzureClient.cs
```diff
@@ -176,57 +187,89 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
...
                 return AzureClientError.InvalidTarget.ToExecutionResult();
             }
 
-            var operationInfo = operationResolver.Resolve(operationName);
-            var entryPointInfo = new EntryPointInfo<QVoid, Result>(operationInfo.RoslynType);
-            var entryPointInput = QVoid.Instance;
+            channel.Stdout($"Submitting {operationName} to target {ActiveTarget.TargetName}...");
 
-            if (execute)
+            IEntryPoint? entryPoint = null;
```
ricardo-espinoza, line 200: Nice that we are embracing nullable reference types. :)

rmshaffer: Yes! @cgranade has been evangelizing this. We are trying to include `#nullable enable` at the top of all new C# files, and gradually add them to the old files.

src/AzureClient/EntryPoint/EntryPointOperationResolver.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal class EntryPointOperationResolver : IOperationResolver
+    {
+        private IEntryPointGenerator EntryPointGenerator { get; }
+
+        public EntryPointOperationResolver(IEntryPointGenerator entryPointGenerator) =>
+            EntryPointGenerator = entryPointGenerator;
+
+        public OperationInfo Resolve(string name) => OperationResolver.ResolveFromAssemblies(name, RelevantAssemblies());
+
+        private IEnumerable<AssemblyInfo> RelevantAssemblies()
```
ricardo-espinoza, line 21: I'm assuming that relevant assemblies does not need to include the info for the entrypoint assembly itself, right?
(EntryPointAssemblyInfo)

rmshaffer: Correct.  Any operation that we want to resolve here must live in one of the other assemblies.

src/AzureClient/AzureClient.cs
```diff
@@ -176,57 +187,89 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
...
-                return job.ToJupyterTable().ToExecutionResult();
+            if (!execute)
+            {
+                return await GetJobStatusAsync(channel, MostRecentJobId);
             }
+
+            var timeoutInSeconds = 30;
+            channel.Stdout($"Waiting up to {timeoutInSeconds} seconds for Azure Quantum job to complete...");
+
+            using (var cts = new System.Threading.CancellationTokenSource(TimeSpan.FromSeconds(30)))
```
cgranade, line 244: Should this timeout be configurable? That may be doable with the `%config` magic.

src/AzureClient/AzureClient.cs
```diff
@@ -176,57 +187,89 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
...
+            {
+                CloudJob? cloudJob = null;
+                do
+                {
+                    // TODO: Once jupyter-core supports interrupt requests (https://github.com/microsoft/jupyter-core/issues/55),
+                    //       handle Jupyter kernel interrupt here and break out of this loop 
+                    var pollingIntervalInSeconds = 5;
```
cgranade, line 251: Similarly here.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="outputType">Specifies the output parameter type for the
+        /// <see cref="Simulation.Core.EntryPointInfo"/> object provided as the <c>entryPointInfo</c> argument.</param>
+        /// <param name="operationInfo">Information about the Q# operation to be used as the entry point.</param>
+        public EntryPoint(object entryPointInfo, Type inputType, Type outputType, OperationInfo operationInfo)
+        {
+            EntryPointInfo = entryPointInfo;
+            InputType = inputType;
+            OutputType = outputType;
+            OperationInfo = operationInfo;
+        }
```
cgranade, line 39: Not at all about your code, and I know I keep saying it, but this is such a great argument for C# 9's init-only properties. Ah, well.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                parameterTypes.Add(parameter.ParameterType);
+                parameterValues.Add(parameterValue);
+            }
+
+            var entryPointInput =
+                parameterValues.Count == 0 ? QVoid.Instance :
+                parameterValues.Count == 1 ? parameterValues.Single() :
+                InputType.GetConstructor(parameterTypes.ToArray()).Invoke(parameterValues.ToArray());
```
cgranade, line 71:
```suggestion
            var entryPointInput = parameterValues.Count switch
            {
                0 => QVoid.Instance,
                1 => parameterValues.Single(),
                _ => InputType.GetConstructor(parameterTypes.ToArray()).Invoke(parameterValues.ToArray())
            };
```

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -0,0 +1,149 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation.
```

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                return SnippetsAssemblyInfo.Assembly;
+            }
+            else if (name.Name == Path.GetFileNameWithoutExtension(WorkspaceAssemblyInfo?.Location))
+            {
+                return WorkspaceAssemblyInfo.Assembly;
+            }
+
+            return null;
+        }
```
cgranade, line 69:
```suggestion
        public Assembly? Resolve(AssemblyLoadContext context, AssemblyName name) => name.Name switch
        {
            Path.GetFileNameWithoutExtension(EntryPointAssemblyInfo?.Location) => EntryPointAssemblyInfo.Assembly,
            Path.GetFileNameWithoutExtension(SnippetsAssemblyInfo?.Location) => SnippetsAssemblyInfo.Assembly,
            Path.GetFileNameWithoutExtension(WorkspaceAssemblyInfo?.Location) => WorkspaceAssemblyInfo.Assembly,
            _ => null
        };
```

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -0,0 +1,149 @@
...
+                Logger?.LogError($"Error compiling entry point for operation {operationName}.");
+                throw new CompilationErrorsException(logger.Errors.ToArray());
+            }
+
+            var entryPointOperationInfo = EntryPointAssemblyInfo.Operations.Single();
+
+            // Construct the EntryPointInfo<,> object
+            var parameterTypes = entryPointOperationInfo.RoslynParameters.Select(p => p.ParameterType).ToArray();
+            var typeCount = parameterTypes.Length;
+            Type entryPointInputType =
```
cgranade, line 136: This may also be clearer as a `switch` expression.

src/AzureClient/EntryPoint/IEntryPointGenerator.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Quantum.IQSharp.AzureClient
+{
+    /// <summary>
+    /// This service is capable of generating entry points for
+    /// job submission to Azure Quantum.
+    /// </summary>
+    public interface IEntryPointGenerator
```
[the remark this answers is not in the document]
cgranade: In general, IQ# has used dependency injection to make service classes like `EntryPointGenerator` available to their dependents. Best practice with the DI framework we've been using is generally to only depend on interfaces rather than classes, allowing for future extensibility without breaking existing dependencies.

src/Tests/SNIPPETS.cs
```diff
@@ -207,6 +207,15 @@ operation DependsOnChemistryWorkspace() : ((JordanWignerEncodingData, Int, Doubl
     }
 ";
 
+        public static string InvalidEntryPoint =
+@"
+    /// # Summary
+    ///     This script has an operation that is not valid to be marked as an entry point.
```
cgranade, line 213:
```suggestion
    /// This script has an operation that is not valid to be marked as an entry point.
```

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -0,0 +1,149 @@
...
+            if (name.Name == Path.GetFileNameWithoutExtension(SnippetsAssemblyInfo?.Location))
+            {
+                return SnippetsAssemblyInfo.Assembly;
+            }
+            else if (name.Name == Path.GetFileNameWithoutExtension(WorkspaceAssemblyInfo?.Location))
+            {
+                return WorkspaceAssemblyInfo.Assembly;
+            }
+
+            return null;
```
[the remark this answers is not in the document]
rmshaffer: Thanks, good catch! This was indeed causing a warning. Really need to go through the IQ# projects and cleanup/suppress warnings and then enable warnings as errors. Currently over 1000 warnings when building iqsharp.sln. ☹️

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            Logger?.LogDebug($"Generating entry point: operationName={operationName}, executionTarget={executionTarget}");
+
+            var logger = new QSharpLogger(Logger);
+            var compilerMetadata = References.CompilerMetadata;
+
+            // Clear references to previously-built assemblies
+            WorkspaceAssemblyInfo = null;
+            SnippetsAssemblyInfo = null;
+            EntryPointAssemblyInfo = null;
```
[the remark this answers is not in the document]
rmshaffer: Same here, I've made these nullable as needed to get rid of the warnings.

src/AzureClient/EntryPoint/EntryPointGenerator.cs
rmshaffer: Is this recommended? From what I can see, just about every source file in IQ# says "All rights reserved." Happy to go through and delete it everywhere if that's the best practice.

cgranade: This is the latest guidance, so far as I understand, and should be used for new code. That said, if I understand correctly, existing code does not need to be updated.

src/AzureClient/EntryPoint/IEntryPointGenerator.cs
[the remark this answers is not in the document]
rmshaffer: If I make these interfaces internal, I have to also make the `AzureClient` constructor internal which takes this as a parameter, and this causes the service instantiation in our dependency injection framework to fail (since that happens from a non-IQ# assembly and requires a public constructor). So, long story short, I think they have to be public. Good question though, I wasn't fully aware of how that worked until I just looked at it.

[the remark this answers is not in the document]
rmshaffer: Quoting @cgranade's comment, which explains it better than I could have and since it didn't appear as a response to this thread for some reason:

> In general, IQ# has used dependency injection to make service classes like `EntryPointGenerator` available to their dependents. Best practice with the DI framework we've been using is generally to only depend on interfaces rather than classes, allowing for future extensibility without breaking existing dependencies.

src/AzureClient/AzureClient.cs
rmshaffer: Yes! In fact I have made it configurable in https://github.com/microsoft/iqsharp/pull/161.

src/AzureClient/AzureClient.cs
rmshaffer: Same, this is configurable in https://github.com/microsoft/iqsharp/pull/161. Reading my mind...

src/AzureClient/EntryPoint/EntryPoint.cs
rmshaffer: Nice, somehow I was not aware of this syntax. C# has infinite shorthand for everything, it's quite amazing.

src/AzureClient/EntryPoint/EntryPointGenerator.cs
rmshaffer: Yes, makes sense! I think I need to add some `var s when s ==` in there to make it compile since the match expression is not a constant.

src/Tests/SNIPPETS.cs
rmshaffer: Leaving this as-is for readability, since it's consistent with the rest of the file.

src/AzureClient/Magic/SubmitMagic.cs
```diff
@@ -69,7 +58,14 @@ public override async Task<ExecutionResult> RunAsync(string input, IChannel chan
         {
             var inputParameters = ParseInputParameters(input, firstParameterInferredName: ParameterNameOperationName);
             var operationName = inputParameters.DecodeParameter<string>(ParameterNameOperationName);
-            return await AzureClient.SubmitJobAsync(channel, OperationResolver, operationName);
+
+            var decodedParameters = new Dictionary<string, string>();
+            foreach (var key in inputParameters.Keys)
+            {
+                decodedParameters[key] = inputParameters.DecodeParameter<string>(key);
+            }
```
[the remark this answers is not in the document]
rmshaffer: This is great, I always prefer LINQ, thanks for pointing this out! I'm going to apply this change in one of my other PRs since I've already started refactoring this a bit on another branch.

src/Core/Compiler/CompilerService.cs
```diff
@@ -67,33 +70,61 @@ private QsCompilation UpdateCompilation(ImmutableDictionary<Uri, string> sources
...
+            var entryPointUri = new Uri(Path.GetFullPath(Path.Combine("/", $"entrypoint.qs")));
+            var entryPointSnippet = @$"namespace ENTRYPOINT
+                {{
+                    open {operation.Header.QualifiedName.Namespace.Value};
+                    @{BuiltIn.EntryPoint.FullName}()
+                    operation {signature}
+                    {{
+                        return {operation.Header.QualifiedName}{argumentTuple};
+                    }}
+                }}";
```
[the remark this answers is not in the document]
rmshaffer: I had some discussions about this with @bettinaheim and @anpaz-msft. The reason I ended up doing it this way is that we want to ignore any `@EntryPoint()` attributes that happen to appear inside snippets (code from notebook cells), and we are doing this by compiling the snippets as a library and suppressing the resulting warning.

If we want to attach the `@EntryPoint()` attribute directly to the operation in the snippet code and compile the snippets as an executable, then we'd have to also go remove the attribute from any other snippet operations where it might exist. I think this is certainly possible by manipulating the syntax tree, as you suggest, and maybe we can change to go that route in the future if needed, but this seemed like the simplest solution at the moment.

One potential benefit of this approach is that in the future it may be possible to reference other Q# projects from IQ#, including Q# library projects. This approach would also give a way to submit those operations to Azure Quantum without having to manually create a wrapper operation in the notebook.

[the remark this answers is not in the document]
samarsha: Seems reasonable, I agree that this is probably simpler. Thanks for explaining!

> One potential benefit of this approach is that in the future it may be possible to reference other Q# projects from IQ#, including Q# library projects. This approach would also give a way to submit those operations to Azure Quantum without having to manually create a wrapper operation in the notebook.

I haven't tested it, but you could probably still attach the entry point attribute to callables in a reference. I think they all get added to the same syntax tree.

rmshaffer:
Thanks @ricardo-espinoza @samarsha @cgranade for the detailed review and feedback! I think I've responded to all of the comments, but please let me know if I've missed something. When you have a chance, please let me know whether you have any additional concerns/feedback, or whether you're comfortable approving these changes.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            // Find and invoke the method on IQuantumMachine that is declared as:
+            // Task<IQuantumMachineJob> SubmitAsync<TInput, TOutput>(EntryPointInfo<TInput, TOutput> info, TInput input)
+            var submitMethod = typeof(IQuantumMachine)
```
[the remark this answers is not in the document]
ricardo-espinoza: Closing the loop here. No objections with keeping it as it is, but it's something we could discuss in the future in terms of the design of the interface itself.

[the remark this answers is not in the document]
rmshaffer: Agreed, thanks @ricardo-espinoza! One minor note for additional context is that the `IQuantumMachine` interface is defined in the [qsharp-runtime repo](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Core/IQuantumMachine.cs) and is also used by the Q# standalone executable framework, which uses generated code to call these methods and so doesn't require reflection. The strong type safety is nice in that case.

PR #157, opened 2020-06-03
Return snippet operations in order of definition

### LINKED ISSUE #148: `qsharp.compile` not returning callables in the same order
**Describe the bug**
The `compile` function provided by the `qsharp` package seems to be alphabetizing the callables it returns, which doesn't always match the order in which they are provided in the string argument to `compile`.

**To Reproduce**
1. Open a Python Jupyter notebook in an environment/system that has the pre-requisites for the qsharp package installed. (included my environment.yml below)
2. Run the following code in the first cell:

```python
import qsharp

[zaz, alpha, foo] = qsharp.compile(
"""
operation Zaz() : Unit {
    Message("Zaz");
}
operation Alpha() : Unit {
    Message("Alpha");
}
operation Foo() : Unit {
    Message("Foo");
}
"""
)
```
3. Once that has run, check to see what python has stored for the output of the compile by running the following:

```python
[zaz, alpha, foo]
```
What you will see is that the compile operation changed the order of the callables that it returned such that the python variable names do not match the callables they represent.

**Expected behavior**
I expect the callables to be returned from the compile step in the same order they were written in the code snippet argument to `qsharp.compile`.

**Screenshots**
image

**System information**
 - OS: Windows 10
```
{'iqsharp': LooseVersion ('0.10.2002.2610'),
 'Jupyter Core': LooseVersion ('1.2.36563.0'),
 '.NET Runtime': LooseVersion ('.NETCoreApp,Version=v3.0'),
 'qsharp': LooseVersion ('0.11.2004.2825')}
```

**Additional context**
The conda environment below is where I was testing this:

```yaml
name: bug-report
channels:
- conda-forge
dependencies:
- notebook
- numpy
- scipy
- qutip
- matplotlib
- pip
- pip:
    - qsharp
```

rmshaffer:
Resolves #148 by always returning snippet operations in the order of definition within the snippet.

Note that magic commands `%who` and `%workspace` are unaffected by this and will continue to return operations in alphabetical order.

### DIFF
--- a/src/Tests/SNIPPETS.cs
+++ b/src/Tests/SNIPPETS.cs
@@ -80,7 +80,9 @@ namespace Tests.IQSharp
         public static string Op6b_Op6a =
 @"
     /// # Summary
-    ///     This to show that the order of snippet operations is preserved:
+    ///     This to show that the order of snippet operations returned
+    ///     from compilation is the same as the order in which they
+    ///     are defined in the code:
     operation Op6b() : Unit
     {
         HelloQ();
@@ -88,7 +90,7 @@ namespace Tests.IQSharp
 
     operation Op6a() : Unit
     {
-        Op6b();
+        HelloQ();
     }
 ";
 
--- a/src/Tests/SNIPPETS.cs
+++ b/src/Tests/SNIPPETS.cs
@@ -77,6 +77,21 @@ namespace Tests.IQSharp
     }
 ";
 
+        public static string Op6b_Op6a =
+@"
+    /// # Summary
+    ///     This to show that the order of snippet operations is preserved:
+    operation Op6b() : Unit
+    {
+        HelloQ();
+    }
+
+    operation Op6a() : Unit
+    {
+        Op6b();
+    }
+";
+
         public static string DependsOnWorkspace =
 @"
     /// # Summary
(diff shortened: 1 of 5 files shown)


src/Tests/SNIPPETS.cs
```diff
@@ -77,6 +77,21 @@ @ EntryPoint (  )
...
+    /// # Summary
+    ///     This to show that the order of snippet operations is preserved:
+    operation Op6b() : Unit
+    {
+        HelloQ();
+    }
+
+    operation Op6a() : Unit
+    {
+        Op6b();
```
ricardo-espinoza, line 91: NIT: Since Op6a() calls Op6b(), then the assert compile should see Op6b first and then Op6a? It might help here to add comments on what the intention of the test is.

src/Tests/SNIPPETS.cs
```diff
@@ -77,6 +77,21 @@ @ EntryPoint (  )
...
 ";
 
+        public static string Op6b_Op6a =
+@"
+    /// # Summary
+    ///     This to show that the order of snippet operations is preserved:
+    operation Op6b() : Unit
+    {
+        HelloQ();
+    }
```
cgranade, line 87: Very minor suggestion here, please feel free to disregard.
```suggestion
    /// # Summary
    /// This to show that the order of snippet operations is preserved:
    operation Op6b() : Unit {
        HelloQ();
    }
```

rmshaffer: I agree, but I'm going to leave it, just for consistency with the whole bunch of other snippets in this file that are using this format.

src/Tests/SNIPPETS.cs
rmshaffer: Thanks! I've clarified this a bit.

src/Tests/SNIPPETS.cs
cgranade: Sounds good!

PR #160, opened 2020-06-08
Add simple display encoders for CloudJob, TargetStatus, and Histogram

rmshaffer:
This PR adds display encoders for `CloudJob`, `TargetStatus`, and `Histogram` objects, which ensures both that they are displayed properly in Jupyter clients and that a reasonable object is returned when invoking from Python.

Fixes [AB#15606](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15606), fixes [AB#15312](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15312), fixes [AB#15310](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15310).


src/AzureClient/AzureClient.cs
```diff
@@ -56,6 +57,17 @@ private string ValidExecutionTargetsDisplayText
             EntryPointGenerator = entryPointGenerator;
             Logger = logger;
             eventService?.TriggerServiceInitialized<IAzureClient>(this);
+
+            if (engine is BaseEngine baseEngine)
+            {
+                baseEngine.RegisterDisplayEncoder(new CloudJobToHtmlEncoder());
+                baseEngine.RegisterDisplayEncoder(new CloudJobToTextEncoder());
+                baseEngine.RegisterDisplayEncoder(new CloudJobToJsonEncoder());
```
cgranade, line 65: Early comment, but for JSON, you can also [directly register the JsonConverter class](https://github.com/microsoft/jupyter-core/blob/master/examples/moon-kernel/MoonEngine.cs#L31).

rmshaffer: Thanks! That's much cleaner, just made that change.

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        public string MimeType => MimeTypes.Html;
+
+        public EncodedData? Encode(object displayable)
+        {
+            if (displayable is CloudJob job) displayable = new List<CloudJob>() { job };
+
+            return displayable is IEnumerable<CloudJob> jobs
+                ? tableEncoder.Encode(jobs.ToJupyterTable())
+                : null;
```
theRoughCode, line 58: Is the reason that we check if `displayable` is of type `IEnumerable<CloudJob>` here because the `displayable` being passed in to `Encode` might be of type `IEnumerable<CloudJob>` instead of `CloudJob`?

rmshaffer: Yes, the `displayable` can be any type. I want this encoder to work for both `CloudJob` and `IEnumerable<CloudJob>`, and to return `null` for any other type so that other encoders get a chance to try.

rmshaffer: In particular, if you look at [this part of the diff](https://github.com/microsoft/iqsharp/pull/160/files#diff-d9fa7e3ad17ec89dbf70f6a428fa661aR406), you'll see that in some cases the code is returning `job`, which is a single `CloudJob`, and in some cases `jobs`, which is an enumeration of `CloudJob` objects.

src/AzureClient/Visualization/TargetStatusEncoders.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public class TargetStatusToHtmlEncoder : IResultEncoder
+    {
+        private static readonly IResultEncoder tableEncoder = new TableToHtmlDisplayEncoder();
+
+        public string MimeType => MimeTypes.Html;
+
+        public EncodedData? Encode(object displayable)
+        {
+            if (displayable is TargetStatus target) displayable = new List<TargetStatus>() { target };
```
cgranade, line 50: Very minor, but given that you've used this pattern a few times so far, would it be worth a new extension method that ensures something is an `IEnumerable`?

```
internal static IEnumerable<T>? AsEnumerableOf<T>(this object? source) =>
    source match
    {
        T singleton => new List<T> { singleton },
        IEnumerable<T> collection => collection.
        _ => null
    };
```
My thought with the naming here matches was to match the `as` operator by returning `null` if the type comparison fails.

src/AzureClient/AzureClient.cs
```diff
@@ -357,19 +364,14 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel channel, string jo
...
-            var histogram = new Dictionary<string, double>();
-            foreach (var entry in deserializedOutput["histogram"] as JObject)
-            {
-                histogram[entry.Key] = entry.Value.ToObject<double>();
-            }
+            using var stream = new MemoryStream();
+
+            // TODO: Use JobStorageHelper.DownloadJobOutputAsync() after https://github.com/microsoft/qsharp-runtime/issues/239 is fixed.
+            // await new JobStorageHelper(ConnectionString).DownloadJobOutputAsync(jobId, stream);
+            await new JobStorageHelper(ConnectionString).StorageHelper.DownloadBlobAsync(
```
anpaz-msft, line 371: The URL with the job output should should be read from the outputData property of CloudJob.

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -0,0 +1,95 @@
...
+            new Table<CloudJob>
+            {
+                Columns = new List<(string, Func<CloudJob, string>)>
+                {
+                    // TODO: add cloudJob.Uri after https://github.com/microsoft/qsharp-runtime/issues/236 is fixed.
+                    ("Job ID", cloudJob => cloudJob.Id),
+                    ("Job Name", cloudJob => cloudJob.Details.Name),
+                    ("Job Status", cloudJob => cloudJob.Status),
+                    ("Provider", cloudJob => cloudJob.Details.ProviderId),
+                    ("Target", cloudJob => cloudJob.Details.Target),
```
anpaz-msft, line 40: Should we include submission/completion times?

src/AzureClient/Visualization/TargetStatusEncoders.cs
```diff
@@ -0,0 +1,91 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Quantum.IQSharp.AzureClient
+{
+    internal static class TargetStatusExtensions
+    {
+        internal static Dictionary<string, object> ToDictionary(this TargetStatus target) =>
+            new Dictionary<string, object>()
+            {
+                { "targetName", target.Id },
```
anpaz-msft, line 22: I think this should be 'targetId'.
We are also considering adding a "Friendly name" to the target metadata that can be used as name later.

src/AzureClient/Visualization/TargetStatusEncoders.cs
```diff
@@ -0,0 +1,91 @@
...
+                { "currentAvailability", target.CurrentAvailability },
+                { "averageQueueTime", target.AverageQueueTime },
+                { "statusPage", target.StatusPage },
```
anpaz-msft, line 25: Not sure we should expose status page here. I think that is more of an internal azure quantum field.

src/AzureClient/AzureClient.cs
rmshaffer: Thanks, this is much more straightforward.

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -0,0 +1,95 @@
...
+        internal static Dictionary<string, object> ToDictionary(this CloudJob cloudJob) =>
+            new Dictionary<string, object>()
+            {
+                // TODO: add cloudJob.Uri after https://github.com/microsoft/qsharp-runtime/issues/236 is fixed.
+                { "id", cloudJob.Id },
+                { "name", cloudJob.Details.Name },
+                { "status", cloudJob.Status },
+                { "provider", cloudJob.Details.ProviderId },
+                { "target", cloudJob.Details.Target },
+            };
```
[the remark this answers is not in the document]
rmshaffer: Thanks! I had no idea this syntax existed.

src/AzureClient/Visualization/TargetStatusEncoders.cs
rmshaffer: Thanks, that's great to know. I'll make a pass through and consistently call this "id" everywhere.

src/AzureClient/Visualization/TargetStatusEncoders.cs
rmshaffer: Sure, I'll remove it. We can always add it back later if needed.

src/AzureClient/Visualization/CloudJobEncoders.cs
rmshaffer: Yes, I think that's a great idea.

src/AzureClient/Visualization/TargetStatusEncoders.cs
rmshaffer: Yes, that's a good suggestion, I'll do this.

PR #161, opened 2020-06-09
Support several parameters for IQ# Azure Quantum job execution

rmshaffer:
This PR adds support for `jobName`, `shots`, `timeout`, and `poll` parameters when submitting Azure Quantum jobs from IQ#.

It also unifies the submission and execution code paths, which had some duplication previously and otherwise would have required more duplication to add these parameters.

To complete before merging:
- [x] Retarget this PR to `feature/azure-client` after https://github.com/microsoft/iqsharp/pull/156 and https://github.com/microsoft/iqsharp/pull/160 are merged.


src/AzureClient/AzureSubmissionContext.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        ///     The execution timeout for the job, expressed in seconds.
+        /// </summary>
+        /// <remarks>
+        ///     This setting only applies to %azure.execute. It is ignored for %azure.submit.
+        ///     The timeout determines how long the IQ# kernel will wait for the job to complete;
+        ///     the Azure Quantum job itself will continue to execute until it is completed.
+        /// </remarks>
+        public int ExecutionTimeout { get; set; } = DefaultExecutionTimeoutInSeconds;
```
cgranade, line 53: Is the plan for these two properties to connect up to `%config`, or will there be another way to set them?

rmshaffer: With these changes, the two properties can be set via `timeout` and `poll` parameters to the `%azure.execute` magic command. It’s worth discussing whether they should be on `%config` instead, but since they only apply to `%azure.execute` it seemed to make sense to put them here for now.

PR #164, opened 2020-06-10
Refactor Azure authentication and service calls

rmshaffer:
This PR contains no functional changes. It is purely refactoring of authentication and service calls out of the `AzureClient` class into the existing `AzureEnvironment` class and a new `AzureWorkspace` class that implements a new `IAzureWorkspace` interface.

In a subsequent PR, I plan to add `AzureEnvironmentType.Mock`, which will use a mock implementation of this new `IAzureWorkspace` interface. This should enable end-to-end integration testing of most of the functionality without any Azure dependency.

Fixes [AB#15335](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15335), fixes [AB#15316](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15316).

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -59,8 +59,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         }
 
         /// <inheritdoc/>
-        public async Task<ExecutionResult> ConnectAsync(
-            IChannel channel,
+        public async Task<ExecutionResult> ConnectAsync(IChannel channel,
             string subscriptionId,
             string resourceGroupName,
             string workspaceName,
--- a/src/AzureClient/AzureSubmissionContext.cs
+++ b/src/AzureClient/AzureSubmissionContext.cs
@@ -19,6 +19,12 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         private static readonly int DefaultShots = 500;
         private static readonly int DefaultExecutionTimeoutInSeconds = 30;
         private static readonly int DefaultExecutionPollingIntervalInSeconds = 5;
+        
+        internal static readonly string ParameterNameOperationName = "__operationName__";
+        internal static readonly string ParameterNameJobName = "jobName";
+        internal static readonly string ParameterNameShots = "shots";
+        internal static readonly string ParameterNameTimeout = "timeout";
+        internal static readonly string ParameterNamePollingInterval = "poll";
 
         /// <inheritdoc/>
         public string FriendlyName { get; set; } = string.Empty;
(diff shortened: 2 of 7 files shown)


src/AzureClient/AzureClient.cs
```diff
@@ -77,98 +69,33 @@ private string ValidExecutionTargetsDisplayText
...
-            {
-                AvailableProviders = await QuantumClient.Providers.GetStatusAsync();
-            }
-            catch (Exception e)
+            AvailableProviders = await ActiveWorkspace.GetProvidersAsync();
+            if (AvailableProviders == null)
             {
-                Logger?.LogError(e, $"Failed to download providers list from Azure Quantum workspace: {e.Message}");
                 return AzureClientError.WorkspaceNotFound.ToExecutionResult();
             }
```
anpaz-msft, line 83: This should be a different error, as there is a workspace...

src/AzureClient/AzureClient.cs
```diff
@@ -77,98 +69,33 @@ private string ValidExecutionTargetsDisplayText
...
 
             return ValidExecutionTargets.ToExecutionResult();
         }
 
         /// <inheritdoc/>
         public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel channel)
         {
-            if (QuantumClient == null || AvailableProviders == null)
+            if (ActiveWorkspace == null || AvailableProviders == null)
             {
```
anpaz-msft, line 94: I wonder if it's the right message that you can't connect to a workspace unless it has Q# providers.

src/AzureClient/AzureEnvironment.cs
```diff
@@ -5,66 +5,151 @@
...
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
-using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.Quantum.Client;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensions.Msal;
+using Microsoft.Jupyter.Core;
```
anpaz-msft, line 15: This code looks great.
Wondering if we should move it to be part of the Azure.Quantum.Client package itself. Pretty much the same should be done wherever we try to call Azure Quantum.

src/AzureClient/AzureClient.cs
rmshaffer: The behavior in this case is that the connection to the workspace will succeed, but there will simply be no valid targets to choose from. But it's a good point, and I'll add an output message to make this clear. It would still be possible, for example, to query the jobs list in this case.

src/AzureClient/AzureClient.cs
rmshaffer: Yes, we have created the "workspace" object on the client, but downloading the providers list is the first time we actually attempt to communicate with Azure Quantum. For example, if the user has mistyped the workspace name, this is the error they will see. The friendly string displayed is: "No Azure Quantum workspace was found that matches the specified criteria."

src/AzureClient/AzureEnvironment.cs
rmshaffer: Agreed, I had the same thought while I was writing it. There are a few Jupyter dependencies sprinkled in, and I want some of these abstractions in IQ# anyway in order to easily mock them, but a lot of this can and probably should be moved into Azure.Quantum.Client. I'll create a task for this.

PR #166, opened 2020-06-11
Can display phase as number, arrow, or both in dumpMachine

Reemeela2121:
%config now has an option to change the display of Phase in dumpMachine as either
1-Arrow
2-Number
3-Arrow and Number
or None, in this case no phase representation is shown when dumpMachine is printed to console.

### DIFF
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -62,6 +62,11 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         public double TruncationThreshold =>
             GetOptionOrDefault("dump.truncationThreshold", 1e-10);
 
+        /// <summary>
+        ///     Allows for options to view phase as arrows, or in radians
+        ///     or both in arrow format and radians. This also allows the
+        ///     option to show None. 
+        /// </summary>
         public PhaseDisplayStyle PhaseDisplayStyle =>
             GetOptionOrDefault("dump.phaseDisplayStyle", PhaseDisplayStyle.ArrowOnly);
     }
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -61,6 +61,9 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// </summary>
         public double TruncationThreshold =>
             GetOptionOrDefault("dump.truncationThreshold", 1e-10);
+
+        public PhaseDisplayStyle PhaseDisplayStyle =>
+            GetOptionOrDefault("dump.phaseDisplayStyle", PhaseDisplayStyle.ArrowOnly);
     }
 
     /// <summary>
--- a/src/Jupyter/Visualization/StateDisplayEncoders.cs
+++ b/src/Jupyter/Visualization/StateDisplayEncoders.cs
@@ -227,11 +227,11 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
                         var (amplitude, basisLabel) = item;
                         var displayPhaseAsArrows = ConfigurationSource.PhaseDisplayStyle;
 
+                        //different options for displaying phase style
                         var phaseCell = ConfigurationSource.PhaseDisplayStyle switch
                         {
                             PhaseDisplayStyle.None => FormattableString.Invariant($@"
                                 <td> 
-                                
                                 </td>
                             "),
                             PhaseDisplayStyle.ArrowOnly => FormattableString.Invariant($@"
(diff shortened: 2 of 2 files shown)


src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -216,6 +225,32 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
                         var (amplitude, basisLabel) = item;
+                        var displayPhaseAsArrows = ConfigurationSource.PhaseDisplayStyle;
+
+                        //different options for displaying phase style
+                        var phaseCell = ConfigurationSource.PhaseDisplayStyle switch
+                        {
+                            PhaseDisplayStyle.None => FormattableString.Invariant($@"
+                                <td> 
+                                </td>
+                            "),
```
cgranade, line 236: Looking at the screenshots you showed me, it may be helpful to not add a cell in this case, rather than an empty cell. That would also take modifying the value of `outputTable` below to not include a `<th>` if `PhaseDisplayStyle` is `None`.
```suggestion
                            PhaseDisplayStyle.None => "",
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -15,6 +15,15 @@
 
 namespace Microsoft.Quantum.IQSharp.Jupyter
 {
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum PhaseDisplayStyle
+    {
+        None,
```
rmshaffer, line 21: Is having a `None` option something that is useful here? I think it may just leave empty cells in the table. Is there a reason someone might prefer that?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -15,6 +15,15 @@
 
 namespace Microsoft.Quantum.IQSharp.Jupyter
 {
+    [JsonConverter(typeof(StringEnumConverter))]
```
cgranade, line 18: It may be helpful to provide an API documentation comment (that is, a comment starting with `///` instead of just `//`) to this new enum explaining what it represents.
```suggestion
    /// <summary>
    ///     Represents different styles for displaying the phases of complex
    ///     amplitudes when displaying state vectors as HTML.
    /// </summary>
    [JsonConverter(typeof(StringEnumConverter))]
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -216,6 +225,32 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                     vector.SignificantAmplitudes(ConfigurationSource).Select(item =>
                     {
                         var (amplitude, basisLabel) = item;
+                        var displayPhaseAsArrows = ConfigurationSource.PhaseDisplayStyle;
```
rmshaffer, line 228: Small suggestion - it looks like this `displayPhaseAsArrows` variable is unused; you could delete this line.

src/Jupyter/Visualization/StateDisplayEncoders.cs
cgranade, line 21: Similarly, you may want to provide API documentation comments for each enum option.
```suggestion
        /// <summary>
        ///     Suppress phase information.
        /// </summary>
        None,
        /// <summary>
        ///     Display phase information as an arrow (<c>↑</c>) rotated by an angle
        ///     dependent on the phase.
        /// </summary>
```

src/Jupyter/ConfigurationSource.cs
```diff
@@ -61,6 +61,9 @@ public interface IConfigurationSource
         /// </summary>
         public double TruncationThreshold =>
             GetOptionOrDefault("dump.truncationThreshold", 1e-10);
+
+        public PhaseDisplayStyle PhaseDisplayStyle =>
```
cgranade, line 65: It may be helpful to have a comment similar to the one above on `TruncateSmallAmplitudes` here as well.

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: I think giving people an option to not show phase at all might be helpful to keep. Also, in case I want to add more features to show phase, it might be helpful to have an option to show nothing/

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Gotcha! Nice catch!

src/Jupyter/ConfigurationSource.cs
cgranade: Looks good, thank you!

src/Jupyter/Visualization/StateDisplayEncoders.cs
rmshaffer: That makes sense!

PR #167, opened 2020-06-12
Add mock Azure classes, improve qsharp.azure Python API, implement end-to-end C# and Python tests

rmshaffer:
This PR contains end-to-end integration tests in C# and Python for IQ# Azure functionality. To avoid a dependency on the Azure service, mock classes `MockAzureWorkspace`, `MockCloudJob`, and `MockQuantumMachine` are added, which provide a fake implementation of all of the Azure-related functionality. When the environment variable `AZURE_QUANTUM_ENV` is set to `mock`, IQ# will use these mock classes rather than the real implementations.

The C# and Python tests set this environment variable. The C# tests validate the end-to-end functionality by making calls directly to the `AzureClient` object. The Python tests are even more comprehensive because they also include the magic command parsing code, which then in turn calls the `AzureClient` object.

Also as part of these changes, the Python API for Azure functionality is defined more explicitly, including specific types returned for each `qsharp.azure` method, and the `AzureClient` implementation is updated to always return the expected type.

Fixes [AB#13627](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/13627), fixes [AB#15165](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15165).


src/Tests/AzureClientTests.cs
```diff
@@ -55,21 +92,177 @@ public void TestAzureExecutionTarget()
 
             targetId = "ionq.targetId";
             executionTarget = AzureExecutionTarget.Create(targetId);
-            Assert.IsNotNull(executionTarget);
```
ricardo-espinoza, line 58: Is this assert not required to validate if Create() returned a valid object?

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -47,11 +47,11 @@ internal static class CloudJobExtensions
...
                     ("Target", cloudJob => cloudJob.Details.Target),
-                    ("Creation Time", cloudJob => cloudJob.Details.CreationTime.ToDateTime()?.ToString()),
-                    ("Begin Execution Time", cloudJob => cloudJob.Details.BeginExecutionTime.ToDateTime()?.ToString()),
-                    ("End Execution Time", cloudJob => cloudJob.Details.EndExecutionTime.ToDateTime()?.ToString()),
+                    ("Creation Time", cloudJob => cloudJob.Details.CreationTime.ToDateTime()?.ToString() ?? string.Empty),
+                    ("Begin Execution Time", cloudJob => cloudJob.Details.BeginExecutionTime.ToDateTime()?.ToString() ?? string.Empty),
+                    ("End Execution Time", cloudJob => cloudJob.Details.EndExecutionTime.ToDateTime()?.ToString() ?? string.Empty),
                 },
-                Rows = jobsList.OrderByDescending(job => job.Details.CreationTime).ToList()
+                Rows = jobsList.OrderByDescending(job => job.Details.CreationTime).ToList(),
```
ricardo-espinoza, line 54: NIT: Was that intentional?

rmshaffer: Yes, there was a compiler warning that all properties in an object initializer should end with a trailing comma.

src/Tests/AzureClientTests.cs
rmshaffer: Correct, it's redundant here, because the following asserts will fail if `executionTarget` is null.

rmshaffer:
Thanks for the review, @ricardo-espinoza!

PR #170, opened 2020-06-15
Improvements to magic command argument parsing

rmshaffer:
This PR improves the magic command argument parsing code in the `AbstractMagic` base class by using regular expression matching rather than manual string parsing.

Previously, arguments were supported only as `key=value`. This had two problems:
1. It was inconsistent with `%config` magic, which supports `key = value` (with whitespace around the equals sign).
2. It did not allow a way to specify `value` as a string containing whitespace. This is desirable, for example, when specifying a job name for an Azure Quantum submission.

With these changes, arguments such as `key = value` and `key = "value with whitespace"` are supported, where whitespace is optionally allowed on both sides of the equals sign. We also support `key` and `"value only"` arguments to magic commands that expect a single-token argument.

Parsing behavior for arguments provided as JSON (i.e., when called from Python) is unaffected by these changes, and there are a number of existing tests which validate that code path.

### DIFF
--- a/src/Jupyter/Magic/AbstractMagic.cs
+++ b/src/Jupyter/Magic/AbstractMagic.cs
@@ -67,24 +67,6 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         ///     Parses the input to a magic command, interpreting the input as
         ///     a name followed by a JSON-serialized dictionary.
         /// </summary>
-        public static (string, Dictionary<string, string>) ParseInput(string input)
-        {
-            if (input == null) return (string.Empty, new Dictionary<string, string> { });
-            var BLANK_SPACE = new char[1] { ' ' };
-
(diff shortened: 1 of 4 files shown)


src/Jupyter/Magic/AbstractMagic.cs
```diff
@@ -97,46 +80,63 @@ public static (string, Dictionary<string, string>) ParseInput(string input)
         {
             Dictionary<string, string> inputParameters = new Dictionary<string, string>();
 
-            var args = input.Split(null as char[], StringSplitOptions.RemoveEmptyEntries);
+            // This regex looks for four types of matches:
+            // 1. (\{.*\})
```
ricardo-espinoza, line 84: Thanks for the detailed explanation of the RE in the comments.

rmshaffer: That was as much for myself as for anyone else! Otherwise there's no way that I would be able to reverse-engineer this regex a month from now. 😄

PR #173, opened 2020-06-19
Support connection to Azure Quantum workspace via resource ID 

rmshaffer:
This PR adds support for single Azure Quantum resource ID to connect to an Azure Quantum workspace, rather than requiring four separate parameters. The storage account connection string also becomes optional.

For example, a connection command may now look like:
```
%azure.connect resourceId=/subscriptions/f846b2bd-d0e2-4a1d-8141-4c6944a9d387/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Quantum/Workspaces/WORKSPACE_NAME
```
where the argument is a resource ID string that can be copied and pasted directly from the Azure portal.

The separate parameters are still supported, and have been renamed to `subscription`, `resourceGroup`, `workspace`, and `storage` for better consistency with the standalone executable command-line argument names.

Resolves [AB#15814](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15814).


src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -93,16 +103,41 @@ public ConnectMagic(IAzureClient azureClient)
...
+            }
+            else
+            {
+                // look for each of the parameters individually
+                subscriptionId = inputParameters.DecodeParameter<string>(ParameterNameSubscriptionId, defaultValue: string.Empty);
+                resourceGroupName = inputParameters.DecodeParameter<string>(ParameterNameResourceGroupName, defaultValue: string.Empty);
+                workspaceName = inputParameters.DecodeParameter<string>(ParameterNameWorkspaceName, defaultValue: string.Empty);
+            }
+
+            var storageAccountConnectionString = inputParameters.DecodeParameter<string>(ParameterNameStorageAccountConnectionString, defaultValue: string.Empty);
```
samarsha, line 140: So the resource ID doesn't include the storage string, but if there is no linked storage, it can be specified like `%azure.connect resourceId=... storage=...`? It looks like that particular combination isn't included in the documentation or tests, should it be added?

src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -93,16 +103,41 @@ public ConnectMagic(IAzureClient azureClient)
...
+            var resourceId = inputParameters.DecodeParameter<string>(ParameterNameResourceId, defaultValue: string.Empty);
+            var subscriptionId = string.Empty;
+            var resourceGroupName = string.Empty;
+            var workspaceName = string.Empty;
+
+            // A valid resource ID looks like:
+            // /subscriptions/f846b2bd-d0e2-4a1d-8141-4c6944a9d387/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Quantum/Workspaces/WORKSPACE_NAME
+            var resourceIdRegex = new Regex(
+                @"^\/subscriptions\/([a-zA-Z0-9\-]*)\/resourceGroups\/([^\s\/]*)\/providers\/Microsoft\.Quantum\/Workspaces\/([^\s\/]*)$");
+            if (!string.IsNullOrEmpty(resourceId) && resourceIdRegex.IsMatch(resourceId))
```
samarsha, line 120: If `resourceId` is never null (which it looks like might be true, since `DecodeParameter` has `defaultValue: string.Empty`), you could do this to avoid matching the regex twice:

```suggestion
            var match = resourceIdRegex.Match(resourceId);
            if (match.Success)
```

You could also do `Match(resourceId ?? string.Empty)` if `resourceId` can be null.

src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -93,16 +103,41 @@ public ConnectMagic(IAzureClient azureClient)
...
-            var workspaceName = inputParameters.DecodeParameter<string>(ParameterNameWorkspaceName);
+            var resourceId = inputParameters.DecodeParameter<string>(ParameterNameResourceId, defaultValue: string.Empty);
+            var subscriptionId = string.Empty;
+            var resourceGroupName = string.Empty;
+            var workspaceName = string.Empty;
+
+            // A valid resource ID looks like:
+            // /subscriptions/f846b2bd-d0e2-4a1d-8141-4c6944a9d387/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Quantum/Workspaces/WORKSPACE_NAME
+            var resourceIdRegex = new Regex(
+                @"^\/subscriptions\/([a-zA-Z0-9\-]*)\/resourceGroups\/([^\s\/]*)\/providers\/Microsoft\.Quantum\/Workspaces\/([^\s\/]*)$");
```
samarsha, line 119: Could you use `/` instead of `\/`? I don't think `/` needs to be escaped.

src/Python/qsharp/tests/test_azure.py
```diff
@@ -51,19 +51,31 @@ def test_empty_workspace():
     jobs = qsharp.azure.jobs()
     assert jobs == []
 
-def test_workspace_with_providers():
+def test_create_workspace_with_parameters():
     """
-    Tests behavior of a mock workspace with mock providers.
+    Tests creation of a mock workspace with mock providers using parameters.
```
samarsha, line 56: You could also consider factoring out the part of `test_workspace_with_providers` that tests job submission, so that you could have a full test for both a workspace created with separate parameters and one created with a resource ID.

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: You're right, `resourceId` cannot be null here. Applying this suggestion, thanks!

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: Good catch, yes, that would improve the readability here.

src/AzureClient/Magic/ConnectMagic.cs
rmshaffer: Yes, good point. From what I understand, the common case should have linked storage, but it would still be good to document the case where it doesn't, and even more important to test it.

src/Python/qsharp/tests/test_azure.py
rmshaffer: Good idea. And actually by doing that I just found a bug related to connecting to multiple workspaces in one session. 😊

PR #175, opened 2020-06-19
Use CompilationUnitManager.GetFileId instead of Uri.AbsolutePath for file IDs

samarsha:
In microsoft/qsharp-compiler#468, I changed the compiler's representation of file IDs from `AbsolutePath` to `LocalPath` to fix a bug. This PR is needed to update a few places in IQ# that assumed that the file IDs were `AbsolutePath`s.

Although, actually, instead of using `LocalPath` explicitly (and needing to update it again if the file ID representation changes in the future), I would rather use `CompilationUnitManager.TryGetFileId` in all cases. The `Try`-ness of `TryGetFileId` makes it awkward to use in places where I believe it's guaranteed to succeed, though.

Since IQ# is accessing `AbsolutePath` without checks, I think that means the URIs in IQ# are guaranteed to be absolute. Are they also guaranteed to be file URIs? (`IsAbsolute` and `IsFile` being true are the only two things that `TryGetFileId` needs.)

@bettinaheim Would you be OK with a non-`Try` version of `TryGetFileId` being added to the compiler?


cgranade:
It looks fine to me, but that's the part of the iqsharp codebase that @anpaz-msft would know better than I would. The only edge case I could imagine is if someone has a file named `__snipets__.qs` as part of a Q# that they then include from a package reference, such that library functions and operations may then have the same local path as the files used to compile snippets.

samarsha:
> The only edge case I could imagine is if someone has a file named `__snipets__.qs` as part of a Q# that they then include from a package reference, such that library functions and operations may then have the same local path as the files used to compile snippets.

I think that this edge case would exist independent of this PR - `AbsolutePath` vs `LocalPath` would be the difference between `C:/foo%20bar/__snippets__.qs` and `C:\foo bar\__snippets__.qs`, so they are both absolute local paths in that sense. The formatting is different but since all the URIs should be changed to `LocalPath` consistently, the potential for conflicts should be the same, I think.

rmshaffer:
@samarsha just FYI, the CI failure you're hitting is because the conda tests that I recently activated in the iqsharp CI aren't working correctly with -beta package versions. There is a workaround for this included in #177, which should be merged to master hopefully within a day. If you want to unblock your CI build here in the meantime, you could just apply this snippet directly to your branch:

https://github.com/microsoft/iqsharp/blob/afaf38f34e0787020209cd554fd4dc40cbc4c054/conda-recipes/iqsharp/test.ps1#L8-L18

src/Core/Core.csproj
```diff
@@ -34,9 +34,9 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="3.0.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.12.20062505-beta" />
-    <PackageReference Include="Microsoft.Quantum.CsharpGeneration" Version="0.12.20062505-beta" />
-    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.12.20062505-beta" />
+    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.2203-beta" />
+    <PackageReference Include="Microsoft.Quantum.CsharpGeneration" Version="0.11.2006.2203-beta" />
+    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.11.2006.2203-beta" />
```
rmshaffer, line 39: I'd recommend reverting this change before merging so that we keep the newer version, since I think @cgranade's recent commit depends on having a new-enough build here.

src/Tool/appsettings.json
```diff
@@ -6,26 +6,25 @@
...
-    "Microsoft.Quantum.Research::0.12.20062505-beta",
-    
+    "Microsoft.Quantum.Research::0.11.2006.2203-beta",
 
-    "Microsoft.Quantum.Providers.IonQ::0.12.20062505-beta",
-    "Microsoft.Quantum.Providers.Honeywell::0.12.20062505-beta",
-    "Microsoft.Quantum.Providers.QCI::0.12.20062505-beta"
+    "Microsoft.Quantum.Providers.IonQ::0.11.2006.2203-beta",
+    "Microsoft.Quantum.Providers.Honeywell::0.11.2006.2203-beta",
+    "Microsoft.Quantum.Providers.QCI::0.11.2006.2203-beta"
```
rmshaffer, line 28: Same comment as in Core.csproj, probably better to keep the newer versions here. (The CI build may fail anyway due to this.)

samarsha: Yes, sorry, I was sort of expecting this to fail but wanted to try it anyway. I'll rebuild the packages.

samarsha: Surprisingly, it didn't fail.

rmshaffer: Interesting, I guess this one was new enough then.

PR #176, opened 2020-06-22
Support kernel interrupt for %azure.execute

rmshaffer:
This PR consumes the latest changes from Microsoft.Jupyter.Core, specifically https://github.com/microsoft/jupyter-core/pull/59 and https://github.com/microsoft/jupyter-core/pull/61, in order to support kernel interrupt for the `%azure.execute` command. For example, this means that if a user clicks the "Interrupt" button in the Jupyter Notebook interface while waiting for a job to finish executing, the cell will immediately return with a message saying that the job has not yet completed. Addresses [AB#15594](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15594).

As part of this, I'm also adding the `CancellationToken` parameter to all of the `AzureClient` magic commands, since ideally in the future all of the Azure service calls will respond to kernel interrupt. And I've updated the `AbstractMagic` base class to derive from the new `CancellableMagicSymbol`, such that any of the built-in IQ# magics will be able to easily support cancellation if they choose. (But no changes to the built-in magics are included in this PR.)

To do before merging:

- [x] Update jupyter.csproj to consume the public Microsoft.Jupyter.Core package version when released.


src/AzureClient/AzureClient.cs
```diff
@@ -178,38 +183,45 @@ private async Task<ExecutionResult> SubmitOrExecuteJobAsync(IChannel channel, Az
...
             {
                 return await GetJobStatusAsync(channel, MostRecentJobId);
             }
 
+            // If the command was %azure.execute, wait for the job to complete and return the job output.
             channel.Stdout($"Waiting up to {submissionContext.ExecutionTimeout} seconds for Azure Quantum job to complete...");
 
-            using (var cts = new System.Threading.CancellationTokenSource(TimeSpan.FromSeconds(submissionContext.ExecutionTimeout)))
+            using (var executionTimeoutTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(submissionContext.ExecutionTimeout)))
+            using (var executionCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(executionTimeoutTokenSource.Token, cancellationToken))
```
cgranade, line 196: Since these using statements both end at the end of scope, it may be cleaner to use the new `using var` syntax to remove a level of indentation.
```suggestion
            using var executionTimeoutTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(submissionContext.ExecutionTimeout));
            using var executionCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(executionTimeoutTokenSource.Token, cancellationToken);
```
Following that, can remove the `{` and L197 and the matching `}` at L213.

src/AzureClient/AzureClient.cs
```diff
@@ -178,38 +183,45 @@ private async Task<ExecutionResult> SubmitOrExecuteJobAsync(IChannel channel, Az
...
-                while (cloudJob == null || cloudJob.InProgress);
             }
 
             return await GetJobResultAsync(channel, MostRecentJobId);
         }
 
         /// <inheritdoc/>
-        public async Task<ExecutionResult> SubmitJobAsync(IChannel channel, AzureSubmissionContext submissionContext) =>
-            await SubmitOrExecuteJobAsync(channel, submissionContext, execute: false);
+        public async Task<ExecutionResult> SubmitJobAsync(IChannel channel, AzureSubmissionContext submissionContext, CancellationToken cancellationToken) =>
```
cgranade, line 219: Since this is public, it may be a nice convienence to have `cancellationToken` default to `CancellationToken.None`.

rmshaffer: Yes, that's a good idea. It looks like I can't use `CancellationToken.None` as a default value because it's not a compile-time constant, but I'll make this take a `CancellationToken?`, default to `null`, and have the code use `cancellationToken ?? CancellationToken.None`.

src/AzureClient/AzureClient.cs
rmshaffer: Makes sense!

PR #179, opened 2020-06-24
Register new display encoders when loading packages.

cgranade:
This PR adds a new event handler that loads display encoders from new packages as they are loaded, allowing packages to extend the display encoder mechanism to handle types not known at IQ# compile time.


src/Core/Core.csproj
```diff
@@ -34,9 +34,9 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.403" />
-    <PackageReference Include="Microsoft.Quantum.CsharpGeneration" Version="0.11.2006.403" />
-    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.11.2006.403" />
+    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.12.20062320-beta" />
```
rmshaffer, line 37: Is that a valid build? Looks like the build failed and some providers packages aren't available. Maybe 0.12.20062314-beta instead?

Also, since you're doing this (thank you!), would it make sense to also update the public CI build to minor version 12?\
https://github.com/microsoft/iqsharp/blob/ccadb8ff3b8998bec65e1d60d79eefe078218900/build/ci.yml#L7

src/Kernel/IQSharpEngine.cs
```diff
@@ -75,9 +79,71 @@ IMagicSymbolResolver magicSymbolResolver
...
+        ///     for extensions to this engine (in particular, for result encoders).
+        /// </summary>
+        private void RegisterPackageLoadedEvent(IServiceProvider services, ILogger logger, IReferences references)
+        {
+            var knownAssemblies = references
+                .Assemblies
+                .Select(asm => asm.Assembly.GetName())
+                .ToImmutableHashSet()
+                // Except assemblies known at compile-time as well.
+                .Add(typeof(StateVectorToHtmlResultEncoder).Assembly.GetName());
```
rmshaffer, line 97: When #177 is merged, the AzureClient assembly will also have some display encoders in it. Just FYI to whichever one of us merges second.

src/Kernel/IQSharpEngine.cs
```diff
@@ -75,9 +79,71 @@ IMagicSymbolResolver magicSymbolResolver
...
+            foreach (var knownAssembly in knownAssemblies) System.Console.WriteLine($"{knownAssembly.FullName}");
+
+            // Register new display encoders when packages load.
+            references.PackageLoaded += (sender, args) =>
+            {
+                logger.LogDebug("Scanning for display encoders after loading {Package}.", args.PackageId);
+                foreach (var assembly in references.Assemblies.Select(asm => asm.Assembly))
+                {
+                    if (!knownAssemblies.Contains(assembly.GetName()))
```
rmshaffer, line 106: [nit] It looks like you could include this if condition as a `.Where()` clause in the `foreach` statement above.

src/Core/Core.csproj
cgranade: Ah, normally broken builds don't seem to make it to the alpha feed... thanks for the catch! As for whether to change `appsettings.json`, my only thought was to update everywhere that version number appeared in a uniform fashion, but I'm happy to revert the changes to appsettings.

src/Kernel/IQSharpEngine.cs
cgranade: Good to know, thanks! Given what you said about conda breaking until #177, it probably makes sense for that one to go first and for me to catch up?

src/Kernel/IQSharpEngine.cs
cgranade: Good point, agreed.

src/Core/Core.csproj
rmshaffer: No, sorry, I meant you *should* update appsettings, and I completely missed the fact that you already did. Too late at night to be reading code. :-)

src/Kernel/IQSharpEngine.cs
rmshaffer: Probably yes, that makes sense. Hoping to be able to merge that first thing tomorrow morning.

PR #180, opened 2020-06-24
T relara/numerical measurement probability

Reemeela2121:
Add config options for measurement probability in DumpMachine.

- [x] add % to numerical representation of measurement probability
- [x] when None is selected, remove column and column header

### DIFF
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -69,7 +69,16 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// </summary>
         public PhaseDisplayStyle PhaseDisplayStyle =>
             GetOptionOrDefault("dump.phaseDisplayStyle", PhaseDisplayStyle.ArrowOnly);
+
+        /// <summary>
+        ///
+        /// </summary>
+
+        public MeasurementDisplayStyle MeasurementDisplayStyle =>
+            GetOptionOrDefault("dump.measurementDisplayStyle", MeasurementDisplayStyle.BarOnly);
     }
+
+
 
     /// <summary>
     ///     An implementation of the
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -73,7 +73,6 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// <summary>
         ///
         /// </summary>
-
         public MeasurementDisplayStyle MeasurementDisplayStyle =>
             GetOptionOrDefault("dump.measurementDisplayStyle", MeasurementDisplayStyle.BarOnly);
     }
--- a/src/Jupyter/Visualization/StateDisplayEncoders.cs
+++ b/src/Jupyter/Visualization/StateDisplayEncoders.cs
@@ -230,7 +230,7 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
     public class StateVectorToHtmlResultEncoder : IResultEncoder
     {
         private const double TWO_PI = 2.0 * System.Math.PI;
-
+        private double count = 0.0;
         /// <inheritdoc />
         public string MimeType => MimeTypes.Html;
         private IConfigurationSource ConfigurationSource;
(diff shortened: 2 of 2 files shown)


src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +295,62 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                                    <progress
+                                        max=""100""
+                                        value=""{System.Math.Pow(amplitude.Magnitude, 2.0) * 100}""
+                                        style=""width: 70%;""
+                                    > 
+                                    <td>
+                                    <p id=""round${count}""> 
+                                    <script>
+                                    var num = {System.Math.Pow(amplitude.Magnitude, 2.0) * 100};
+                                    num = num.toFixed(2);
```
rmshaffer, line 329: This will be really nice! Just a suggestion: for the numerical display, perhaps we shouldn't multiply by 100, so that the users just see the probabilities in decimal form? And if we do that, I suspect it will be useful to show more than 2 digits after the decimal point. I'm not sure what is a good number, maybe 4 or 6?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -203,7 +230,7 @@ IConfigurationSource configurationSource
     public class StateVectorToHtmlResultEncoder : IResultEncoder
     {
         private const double TWO_PI = 2.0 * System.Math.PI;
-
+        private double count = 0.0;
```
rmshaffer, line 233: A tip here: instead of adding a new `count` variable, you could edit this line below:
https://github.com/microsoft/iqsharp/blob/8589ea284a6ff48da5cf1f7438a8c424c85b18da/src/Jupyter/Visualization/StateDisplayEncoders.cs#L272
to become:
```
vector.SignificantAmplitudes(ConfigurationSource).Select((item, index) => 
```
and then `index` will be an integer that is automatically incremented for each iteration. No need to keep track of it yourself.

src/Jupyter/Visualization/StateDisplayEncoders.cs
cgranade: That could even be configurable, actually, with something like 4 as a default.

src/Jupyter/Visualization/StateDisplayEncoders.cs
theRoughCode: I would be careful of duplicate divs with the same `id` if you go with this approach (i.e. multiple cells => `getElementById` might return the wrong one).

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Created a new config option to set the measurement probability precision, default is 4. Also got rid of multiplication by 100 when viewing the measurement probability as a decimal.

src/Jupyter/ConfigurationSource.cs
```diff
@@ -69,8 +69,17 @@ public interface IConfigurationSource
         /// </summary>
         public PhaseDisplayStyle PhaseDisplayStyle =>
             GetOptionOrDefault("dump.phaseDisplayStyle", PhaseDisplayStyle.ArrowOnly);
+
+        /// <summary>
+        ///
```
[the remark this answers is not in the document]
Reemeela2121: Added more commenting :)

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +295,62 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+
+                        //different options for displaying measurement style
+    
+                        var measurementCell = ConfigurationSource.MeasurementDisplayStyle switch
+                        {
+                            MeasurementDisplayStyle.None =>FormattableString.Invariant($@"
+                                <td> 
+                                    
+                                </td>
+                            "),
```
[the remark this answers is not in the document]
Reemeela2121: Fixed!

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: I'm worried making this change will break how I've printed out all the percentages for numerical measurement probability display

rmshaffer: Makes sense! I didn't consider how this variable maintains its state across many calls to the encoder. I guess there may be other clever ways to do this, but this is fine as an initial implementation! (One more small suggestion: `count` could probably be an `int` rather than a `double`.)

src/Jupyter/Visualization/StateDisplayEncoders.cs
rmshaffer: @Reemeela2121 it looks like maybe the change you're referring to didn't get applied properly? I don't see it here.

PR #183, opened 2020-06-25
Azure Quantum integration fixes

### LINKED ISSUE #178: Some Q# operations can't be found from Python/IQ# when an @EntryPoint exists in the folder
**Describe the bug**
If the folder contains a Q# operation with an `@EntryPoint()` attribute, then some subset of the Q# operations in the folder can't be found by Python or IQ#.

It appears that only Q# operations which are directly or indirectly invoked by the `@EntryPoint()` attribute are available for use from Python or IQ#.

Opening this issue based on the discussion in https://github.com/MicrosoftDocs/quantum-docs-pr/issues/925.

**To Reproduce**
1. Create a Q# file with a few operations in it, one of which has an `@EntryPoint()` attribute, and one of which is not called at all when invoking the `@EntryPoint()`. Something like this:

    ```
    namespace Test {
        open Microsoft.Quantum.Intrinsic;
        open Microsoft.Quantum.Canon;
    
        operation Test1() : Unit {
            Message("Hello from Test1!");
        }
    
        operation Test2() : Unit {
            Message("Hello from Test2!");
        }
    
        @EntryPoint()
        operation Test3() : Unit {
            Message("Hello from Test3!");
            Test2();
        }
    }
    ```

2. Run the following command to see the list of Q# operations available in Python/IQ#:

    ```
    python -c "import qsharp; print(qsharp.get_available_operations())"
    ```

3. The output shows that only a subset of the Q# operations are listed:

    ```
    ['Test.Test2', 'Test.Test3']
    ```

4. Note that if you remove the `@EntryPoint()` operation and run the above command again, then all of the operations become available.

**Expected behavior**
This is confusing. I would have expected either:
1. to *only* be able to use the Q# operation with the `@EntryPoint()` attribute
2. to be able to use any Q# operation defined in the folder

**System information**
 - OS: Windows
 - Version: 0.11.2006.403

**Additional context**
This is related to the fact that IQ# compiles .qs files into an executable, as per the `compileAsExecutable: true` flag here:
https://github.com/microsoft/iqsharp/blob/2f21e62f3989cb943417f257c88a9e64c3788cc2/src/Core/Compiler/CompilerService.cs#L84-L91

rmshaffer:
Fixing a few minor bugs found in Azure Quantum integration testing:
- Token cache was using the same filename as the Python QIO SDK, and the files are saved in different formats. Using a different filename to avoid this.
- Regex parsing of the resource ID string is made case-insensitive. Turns out Azure is not consistent about the casing of the various parts of this string.
- Better error message when workspace connection parameters are missing or invalid.
- Compiling the workspace Q# files as a library, in addition to compiling the Q# snippets as a library. This fixes #178 and also avoids some compilation errors in IQ# when compiling some otherwise valid Q# programs for execution on Azure Quantum.

### DIFF
--- a/src/Python/qsharp/tests/test_azure.py
+++ b/src/Python/qsharp/tests/test_azure.py
@@ -74,7 +74,7 @@ def test_workspace_create_with_resource_id():
     resourceGroupName = "test"
     workspaceName = "WorkspaceNameWithMockProviders"
     targets = qsharp.azure.connect(
-        resourceId=f"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Quantum/Workspaces/{workspaceName}")
+        resourceId=f"/subscriptions/{subscriptionId}/RESOurceGroups/{resourceGroupName}/providers/Microsoft.Quantum/Workspaces/{workspaceName}")
     assert isinstance(targets, list)
     assert len(targets) > 0
 
--- a/src/Python/qsharp/tests/test_azure.py
+++ b/src/Python/qsharp/tests/test_azure.py
@@ -69,6 +69,7 @@ def test_workspace_create_with_parameters():
 def test_workspace_create_with_resource_id():
     """
     Tests behavior of a mock workspace with providers, using resource ID to connect.
+    Also verifies case-insensitivity of resource ID parsing.
     """
     subscriptionId = "f846b2bd-d0e2-4a1d-8141-4c6944a9d387"
     resourceGroupName = "test"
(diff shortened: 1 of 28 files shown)


src/Python/qsharp/tests/test_azure.py
```diff
@@ -74,7 +74,7 @@ def test_workspace_create_with_resource_id():
     resourceGroupName = "test"
     workspaceName = "WorkspaceNameWithMockProviders"
     targets = qsharp.azure.connect(
-        resourceId=f"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Quantum/Workspaces/{workspaceName}")
+        resourceId=f"/subscriptions/{subscriptionId}/RESOurceGroups/{resourceGroupName}/providers/Microsoft.Quantum/Workspaces/{workspaceName}")
```
ricardo-espinoza, line 77: Was this an accidental change?

anpaz-msft: no, it's a test to make sure it's not case sensitive.

rmshaffer: No, it's to test the case-insensitivity. 😊 I'll add a comment so that's clear, thanks for questioning.

PR #186, opened 2020-06-26
Adding MSAL token cache parameters for Mac and Linux

### LINKED ISSUE #184: MSAL authentication doesn't work on MacOS
**Describe the bug**
On MacOS, running `%azure.connect` gives an error message like: `keyChainServiceName is null`

Probably a similar error happens on Linux.

**Expected behavior**
The fix should involve adding calls to `WithMacKeyChain()` and `WithLinuxKeyring()` to the `StorageCreationPropertiesBuilder` here:
https://github.com/microsoft/iqsharp/blob/6d7b8be40a701d8cdb42672c32a2a986ec6b1fe2/src/AzureClient/AzureEnvironment.cs#L78

Thanks to @anpaz-msft for reporting this!

rmshaffer:
The `%azure.connect` command currently works only on Windows because the usage of `StorageCreationPropertiesBuilder` is incomplete. See #184 for details.

Manual verification and Azure Quantum job submission performed:
- [x] on Windows
- [x] on Mac OS X (used 10.15.5)
- [x] on Linux

Fixes #184.

### DIFF
--- a/src/AzureClient/AzureEnvironment.cs
+++ b/src/AzureClient/AzureEnvironment.cs
@@ -75,7 +75,17 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
 
             // Register the token cache for serialization
             var cacheFileName = "iqsharp.bin";
-            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, ClientId).Build();
+            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, ClientId)
+                .WithMacKeyChain(
+                    serviceName: "Microsoft.Quantum.IQSharp",
+                    accountName: "MSALCache")
+                .WithLinuxKeyring(
+                    schemaName: "com.microsoft.quantum.iqsharp",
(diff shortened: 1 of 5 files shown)


src/AzureClient/AzureEnvironment.cs
```diff
@@ -75,7 +75,17 @@ public static AzureEnvironment Create(string subscriptionId)
 
             // Register the token cache for serialization
             var cacheFileName = "iqsharp.bin";
-            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, ClientId).Build();
+            var storageCreationProperties = new StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, ClientId)
+                .WithMacKeyChain(
+                    serviceName: "Microsoft.Quantum.IQSharp",
+                    accountName: "MSALCache")
```
efratshabtai, line 81: Please verify on Mac and Linux before completing.

rmshaffer: Yes, already done, thanks for confirming.

PR #187, opened 2020-06-26
Automatically retrieve embedded resources defined in Kernel.csproj

theRoughCode:
Currently, each new `EmbeddedResource` we add in `Kernel.csproj` has to be manually added in as a KV-pair in the dictionary provided to `.WithKernelSpecResources` in `Program.cs`.

This PR automates this process by finding all `EmbeddedResources` within `Kernel.csproj` and creates a dictionary of all resources.

Note: `WithKernelSpecResources` cannot handle nested directory file names (possibly because it cannot write to a non-existing directory [[link](https://github.com/microsoft/jupyter-core/blob/e144ee6bfa88b17a8be2b368626511701a56bfe6/src/KernelApplication.cs#L467)]). So, we add the embedded resource with its period-separated path (i.e. `folder/file.js` -> `folder.file.js`) as a key to the dictionary.

TypeScript files are also compiled into one out file (`res/compiled.js`) for improved browser caching and avoiding the above pitfall with nested directories.

### DIFF
--- a/src/Tool/Program.cs
+++ b/src/Tool/Program.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Quantum.IQSharp
 
         /// <summary>
         /// Creates dictionary of kernelspec file names to the embedded resource path
-        /// of all embedded resources found in `Kernel.csproj`.
+        /// of all embedded resources found in <c>Kernel.csproj</c>.
         /// Note: `WithKernelSpecResources` cannot handle keys that include directories
         /// so we treat all path names as period-separated file names.
         /// </summary>
--- a/src/Tool/Program.cs
+++ b/src/Tool/Program.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Quantum.IQSharp
         /// <summary>
         /// Creates dictionary of kernelspec file names to the embedded resource path
         /// of all embedded resources found in <c>Kernel.csproj</c>.
-        /// Note: `WithKernelSpecResources` cannot handle keys that include directories
+        /// Note: <c>WithKernelSpecResources</c> cannot handle keys that include directories
         /// so we treat all path names as period-separated file names.
         /// </summary>
         private static Dictionary<string, string> GetEmbeddedKernelResources() =>
--- a/src/Tool/Program.cs
+++ b/src/Tool/Program.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Jupyter.Core;
 using Microsoft.Quantum.IQSharp.Kernel;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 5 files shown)


src/Tool/Program.cs
```diff
@@ -35,6 +36,28 @@ public static bool TelemetryOptOut
...
+                        // Compute "filename.extension"
+                        // Note: WithKernelSpecResources cannot handle keys that include directories
+                        // so we only use the filename without directory path.
+                        var fileName = String.Join(".", resName.Split(".").TakeLast(2));
+                        res.Add(fileName, resName);
+                    }
+                }
+            }
+            return res;
+        }
```
rmshaffer, line 59: If you want to be clever with Linq, I think you could do this all in one line. Something like:
```
        public static Dictionary<string, string> GetEmbeddedKernelResources() =>
            AppDomain.CurrentDomain.GetAssemblies()
                .Single(asm => asm.GetName().Name == "Microsoft.Quantum.IQSharp.Kernel")
                .GetManifestResourceNames()
                .ToDictionary(resName => string.Join(".", resName.Split(".").TakeLast(2)));
```

src/Tool/Program.cs
```diff
@@ -35,6 +36,28 @@ public static bool TelemetryOptOut
             => true;
 #endif
 
+        // Creates dictionary of kernelspec file names to the embedded resource path
+        // of all embedded resources found in `Kernel.csproj`.
```
rmshaffer, line 40: It's best for documentation like this to use triple-slash notation so that it gets picked up by IntelliSense, doc generation, etc. Like:
```
/// <summary>
/// Text here.
/// </summary>

File: src/Tool/Program.cs [hunk 3]
```diff
@@ -35,6 +36,28 @@ public static bool TelemetryOptOut
...
+        public static Dictionary<string, string> GetEmbeddedKernelResources()
```
rmshaffer, line 41: This should probably be `private`, since it's just for use inside this class.

src/Tool/Program.cs
```diff
@@ -35,6 +36,28 @@ public static bool TelemetryOptOut
...
+        {
+            var res = new Dictionary<string, string>();
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                if (String.Equals(assembly.GetName().Name, "Microsoft.Quantum.IQSharp.Kernel"))
+                {
+                    foreach (var resName in assembly.GetManifestResourceNames())
+                    {
+                        // Compute "filename.extension"
```
rmshaffer, line 50: Maybe also worth noting (in the documentation for the function?) that this won't work for EmbeddedResources that have multiple `.`'s in the filename.

src/Tool/Program.cs
theRoughCode: @rmshaffer wow, that's is really nice haha! Thanks for the suggestion!

rmshaffer: Linq allows you to do ridiculously complicated things with very compact syntax. Highly recommend. 👍

src/Tool/Program.cs
theRoughCode: Would it make more sense then to not only take the last 2 components of the path, but use everything after `Microsoft.Quantum.Kernel.res`? For example, `Microsoft.Quantum.Kernel.res.class.subclass.js` => `class.subclass.js`? Any file that has directories will also be treated in this fashion (which would break imports if any file is trying to reference it).

rmshaffer: Yes, I think that would also make sense.

src/Tool/Program.cs
```diff
@@ -35,6 +36,19 @@ public static bool TelemetryOptOut
             => true;
 #endif
 
+        /// <summary>
+        /// Creates dictionary of kernelspec file names to the embedded resource path
+        /// of all embedded resources found in `Kernel.csproj`.
```
cgranade, line 41: C# doesn't use Markdown for `///`, but uses XML instead. Thus, you'll likely want to modify:
```suggestion
        /// of all embedded resources found in <c>Kernel.csproj</c>.
```

src/Tool/Program.cs
```diff
@@ -35,6 +36,19 @@ public static bool TelemetryOptOut
             => true;
 #endif
 
+        /// <summary>
+        /// Creates dictionary of kernelspec file names to the embedded resource path
+        /// of all embedded resources found in <c>Kernel.csproj</c>.
+        /// Note: `WithKernelSpecResources` cannot handle keys that include directories
```
rmshaffer, line 42:
```suggestion
        /// Note: <c>WithKernelSpecResources</c> cannot handle keys that include directories
```

PR #188, opened 2020-06-26
Implied resourceId parameter for %azure.connect

rmshaffer:
This PR allows `%azure.connect` to recognize a `resourceId` parameter without requiring the key name `resourceId` to be explicitly specified.

For example, a command like this:
```
%azure.connect resourceId=/subscriptions/.../Microsoft.Quantum/Workspaces/WORKSPACE_NAME
```
can now be expressed as just:
```
%azure.connect /subscriptions/.../Microsoft.Quantum/Workspaces/WORKSPACE_NAME
```

### DIFF
--- a/src/AzureClient/Magic/ConnectMagic.cs
+++ b/src/AzureClient/Magic/ConnectMagic.cs
@@ -52,7 +52,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                         The Azure Quantum workspace can be identified by resource ID:
 
                         - `{ParameterNameResourceId}=<string>`: The resource ID of the Azure Quantum workspace.
-                        This can be obtained from the workspace page in the Azure portal.
+                        This can be obtained from the workspace page in the Azure portal. The `{ParameterNameResourceId}=` prefix
+                        is optional for this parameter, as long as the resource ID is valid.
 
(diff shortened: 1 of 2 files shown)


src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -147,9 +148,15 @@ public override async Task<ExecutionResult> RunAsync(string input, IChannel chan
 
             // A valid resource ID looks like:
             // /subscriptions/f846b2bd-d0e2-4a1d-8141-4c6944a9d387/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Quantum/Workspaces/WORKSPACE_NAME
-            var match = Regex.Match(resourceId,
+            var resourceIdRegex = new Regex(
```
anpaz-msft, line 151: should this be a constant or a static variable?

efratshabtai:
Is this consistent with other %azure.x commands, meaning, are the other commands require a 'parameter_name = ' or is this the only azure.x command that requires it?
Same question for the CLI.

rmshaffer:
> Is this consistent with other %azure.x commands, meaning, are the other commands require a 'parameter_name = ' or is this the only azure.x command that requires it?
> 
> Same question for the CLI.

Yes, every other Azure magic command accepts input without requiring a parameter name.

PR #193, opened 2020-07-02
T relara/measurement probability histogram

Reemeela2121:
Set up end-to-end communication between kernel and browser. This is still a work in progress, need to: 

- [x] rewrite JSONProperty tags so that displayable state is listed, which will modularize code better and improve readability
- [x] make the div an interactive element for proof that the browser can have interactive elements


src/Jupyter/ConfigurationSource.cs
```diff
@@ -69,8 +69,32 @@ public interface IConfigurationSource
...
+        ///     Allows for options to view measurement as horizontal histograms for each
+        ///     basis state, as well as the option to toggle measurement probability as
+        ///     a decimal, progress bar, both, or None. 
+        /// </summary>
+        public MeasurementDisplayStyle MeasurementDisplayStyle =>
+            GetOptionOrDefault("dump.measurementDisplayStyle", MeasurementDisplayStyle.BarOnly);
+
+        /// <summary>
+        ///     Allows for options to change measurement probability precision when viewing
+        ///     measurement in a decimal format.
```
cgranade, line 83:
```suggestion
        ///     measurement probabilities in a decimal format. 
```

src/Jupyter/Visualization/MeasurementHistogramContent.cs
```diff
@@ -0,0 +1,28 @@
+#nullable enable
```
cgranade, line 1: We'll need to make sure that this file has the same copyright header as other C# files in the repository.

src/Jupyter/Visualization/MeasurementHistogramContent.cs
```diff
@@ -0,0 +1,28 @@
...
+        [JsonProperty("n_qubits")]
+        public int NQubits { get; set; }
+
+        [JsonProperty("amplitudes")]
+        public IEnumerable<(Complex, string)> Amplitudes { get; set; }
+
+    }
+}
+
+//TODO: Why is are they highlighted in red?
```
cgranade, line 28: It looks like this comment is outdated?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -92,6 +119,11 @@ public class DisplayableState
                 )
             );
 
+        /// <summary>
+        ///     Where the vertical measurement probability histogram will be.
```
cgranade, line 123:
```suggestion
        ///     ID for an HTML element where the vertical measurement probability histogram
        ///     will be displayed.
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -203,6 +235,9 @@ IConfigurationSource configurationSource
     public class StateVectorToHtmlResultEncoder : IResultEncoder
     {
         private const double TWO_PI = 2.0 * System.Math.PI;
+        private double count = 0.0;
```
cgranade, line 238: I'd strongly suggest making this an int rather than a double, since it's only being used to hold integer values.
```suggestion
        private int count = 0;
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -203,6 +235,9 @@ IConfigurationSource configurationSource
...
+
+        private int measurement_off = 0; //when equal to 1 then if statement works
```
cgranade, line 240: For consistency with other C# variable names, I'd suggest `camelCase` here:
```suggestion
        private int measurementOffset = 0; //when equal to 1 then if statement works
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                             _ => throw new ArgumentException($"Unsupported style {ConfigurationSource.PhaseDisplayStyle}")
                         };
 
-                        return FormattableString.Invariant($@"
-                            <tr>
-                                <td>$\left|{basisLabel}\right\rangle$</td>
-                                <td>${amplitude.Real:F4} {(amplitude.Imaginary >= 0 ? "+" : "")} {amplitude.Imaginary:F4} i$</td>
+                        count = count + 1;
```
cgranade, line 306:
```suggestion
                        count += 1;
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+
+                        //different options for displaying measurement style
```
cgranade, line 308:
```suggestion
                        // Different options for displaying measurement style.
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                        var measurementHistogram = ConfigurationSource.MeasurementDisplayHistogram;
+
+                        var measurementPrecision = ConfigurationSource.MeasurementDisplayPrecision;
+    
+                        var measurementCell = ConfigurationSource.MeasurementDisplayStyle switch
+                        {
+                            MeasurementDisplayStyle.None => String.Empty,
+                            //measurement_off => 1;
```
cgranade, line 316: What does this line indicate?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                                        style=""width: 100%;""
+                                    > 
+                                    <td>
+                                    <p id=""round${count}""> 
+                                    <script>
+                                    var num = {System.Math.Pow(amplitude.Magnitude, 2.0) * 100};
+                                    num = num.toFixed({measurementPrecision});
+                                    var num_string = num + ""%"";
+                                     document.getElementById(""round${count}"").innerHTML = num_string;
+                                    </script> </p>
```
cgranade, line 340: I'd suggest that it would be simpler here to generate the measurement percentages in C# rather than injecting JavaScript here, since you don't have any hook for those measurement probabilities to change on the client side.
That would also allow you to remove the dependency on `count` entirely, actually. 
```suggestion
                                    <p id=""round${count}"">
                                        {(
                                            System.Math.Pow(amplitude.Magnitude, 2.0) * 100
                                        ).ToString($"F{measurementPrecision}")}
                                    </p>
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                                    </td>
+                                </td>
+                            "), // not sure what style width does
```
cgranade, line 343: The width attribute forces the progress bar to take up all available space.

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -268,17 +303,65 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                            //TODO: fix viewing % (might have accomplished this)
```
cgranade, line 344: Looks like there's a remaining TODO here?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -294,6 +377,7 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                     BasisStateLabelingConvention.BigEndian => " (big endian)",
                     _ => ""
                 };
+                //TODO: rewrite all this code here for less code written
```
cgranade, line 380: Looks like a remaining TODO here as well?

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -306,11 +390,51 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                             </tr>
                         </thead>
 
+                        <tbody>
+                            {formattedData}
+                        </tbody>
+                    </table>
+                "; 
+                if (ConfigurationSource.PhaseDisplayStyle == PhaseDisplayStyle.None) {
```
cgranade, line 398: You should be able to consolidate this with the code at line 381 a bit:

```C#
var outputTable = $@"
                    <table style=""table-layout: fixed; width: 100%"">
                        <thead>
                            {qubitIdsRow}
                            <tr>
                                <th style=""width: {basisWidth}ch)"">Basis state{basisStateMnemonic}</th>
                                <th style=""width: 20ch"">Amplitude</th>
                                {
                                   ConfigurationSource.MeasurementDisplayStyle == MeasurementDisplayStyle.None
                                    ? String.Empty
                                    : $"<th style=""width: calc(100% - 26ch - {basisWidth}ch)"">Meas. Pr.</th>"
                                }
                                {
                                    ConfigurationSource.PhaseDisplayStyle == PhaseDisplayStyle.None
                                    ? String.Empty
                                    : "<th style=\"width: 6ch\">Phase</th>"
                                }
                            </tr>
                        </thead>
                        <tbody>
                            {formattedData}
                        </tbody>
                    </table>
                "; 

```

src/Jupyter/Visualization/StateDisplayOperations.cs
```diff
@@ -97,13 +98,35 @@ public override bool Dump(IQArray<Qubit>? qubits = null)
...
+            };
+            Channel.Display(state);
+
+            Channel.SendIoPubMessage(new Message
+            {
+                Header = new MessageHeader()
+                {
+                    MessageType = "iqsharp_state_dump"
+                },
+                Content = new MeasurementHistogramContent() //TODO: throwing error
```
cgranade, line 120: What error is being thrown here?

src/Jupyter/Visualization/StateDisplayOperations.cs
```diff
@@ -97,13 +98,35 @@ public override bool Dump(IQArray<Qubit>? qubits = null)
             // At this point, _data should be filled with the full state
             // vector, so let's display it, counting on the right display
             // encoder to be there to pack it into a table.
-            Channel.Display(new DisplayableState
+
+            var count = 0; //TODO: figure out how to increment
```
cgranade, line 102: Probably easier to just make a new GUID here.
```suggestion
            var id = System.Guid.NewGuid(); //TODO: figure out how to increment
```

src/Jupyter/Visualization/StateDisplayOperations.cs
```diff
@@ -97,13 +98,35 @@ public override bool Dump(IQArray<Qubit>? qubits = null)
...
+            var state = new DisplayableState
             {
                 // We cast here as we don't support a large enough number
                 // of qubits to saturate an int.
                 QubitIds = qubits?.Select(q => q.Id) ?? Simulator.QubitIds.Select(q => (int)q),
                 NQubits = (int)_count,
-                Amplitudes = _data
+                Amplitudes = _data,
+                DivId = $"dump-machine-div-{count}"
```
cgranade, line 110:
```suggestion
                DivId = $"dump-machine-div-{id}" 
```

src/Jupyter/Visualization/StateDisplayOperations.cs
```diff
@@ -97,13 +98,35 @@ public override bool Dump(IQArray<Qubit>? qubits = null)
...
+                Header = new MessageHeader()
+                {
+                    MessageType = "iqsharp_state_dump"
+                },
+                Content = new MeasurementHistogramContent() //TODO: throwing error
+                {
+                    DivId = state.DivId,
+                    QubitIds = state.QubitIds,
+                    NQubits = state.NQubits,
+                    Amplitudes = state.SignificantAmplitudes(
```
cgranade, line 125: As discussed elsewhere, it would be good to make sure that _all_ amplitudes are passed to the client here.

src/Kernel/client/kernel.ts
```diff
@@ -124,9 +124,23 @@ class Kernel {
         IPython.notebook.kernel.events.on("kernel_ready.Kernel", args => {
             this.requestEcho();
             this.requestClientInfo();
+            this.setupMyMessageListener(); //TODO: call this something else
```
cgranade, line 127: Looks like a remaining TODO here as well.

src/Jupyter/Visualization/StateDisplayEncoders.cs
rmshaffer: In fact, as @cgranade also commented elsewhere, it may be better to just use `System.Guid.NewGuid()` instead of keeping track of a `count` variable. There is a slight danger in just using an integer like `count` to uniquely identify your HTML elements, because you don't control the entire HTML tree, and it's possible that some code somewhere else would create an element with the same ID. Using `NewGuid()` gives you a randomly-generated ID that is guaranteed to be unique, e.g., `0233e4ed-60b5-497e-bef0-ab1b0d203510`. Yes, it's long and ugly, but the JavaScript doesn't care... 😊

src/Kernel/client/kernel.ts
```diff
@@ -124,9 +124,23 @@ class Kernel {
...
     }
 
+    setupMyMessageListener() {
+        IPython.notebook.kernel.register_iopub_handler(
+            "iqsharp_state_dump",
+            message => {
+                console.log("my message received", message);
+                message.content.amplitudes;
+                var htmlGraph = "hi :) ";
+                document.getElementById(message.content.div_id).innerHTML = htmlGraph;
```
rmshaffer, line 138: You may want to safely handle the case where the `<div>` you're looking for doesn't exist, for example, when `ConfigurationSource.MeasurementDisplayHistogram` is false and you don't create that `<div>` in the first place.

src/Jupyter/Visualization/StateDisplayOperations.cs
rmshaffer: That makes sense. Sorry, I think I initially made the suggestion to use `SignificantAmplitudes` here, but I agree it would be nice for the client to get everything and be able to decide which amplitudes to display.

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -320,25 +328,28 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                                     document.getElementById(""round${count}"").innerHTML = num_string;
                                     </script> </p>
                                     </td>
                                 </td>
                             "), // not sure what style width does
+                            //TODO: fix viewing % (might have accomplished this)
                             MeasurementDisplayStyle.NumberOnly => FormattableString.Invariant($@"
                                 <td> 
-                                    <p id=""round${count}""> 
+                                    <p id=""round${count}""; style=""text-align: right"">
```
rmshaffer, line 347: Small suggestion, since semicolons aren't needed between attributes:
```suggestion
                                    <p id=""round${count}"" style=""text-align: right""> 
```

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -378,11 +390,51 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
...
+                        </thead>
+
                         <tbody>
                             {formattedData}
                         </tbody>
                     </table>
                 ";
+                };
+                
+                if (ConfigurationSource.MeasurementDisplayHistogram == true) {
```
rmshaffer, line 435: Minor suggestion, but `== true` isn't needed when testing a `bool` value:
```suggestion
                if (ConfigurationSource.MeasurementDisplayHistogram) {
```

src/Jupyter/Visualization/MeasurementHistogramContent.cs
Reemeela2121: Deleted the comment

src/Jupyter/Visualization/StateDisplayOperations.cs
Reemeela2121: The comment is outdated, sorry!

src/Kernel/client/kernel.ts
Reemeela2121: Renamed it!

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Deleted the variable because I didn't use it anywhere else

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Noted! Made the correction

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Deleted the outdated comment

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Removed outdated comment!

src/Jupyter/Visualization/StateDisplayEncoders.cs
Reemeela2121: Fixed this! thanks!

src/Kernel/client/kernel.ts
Reemeela2121: Got it, thanks!

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -295,17 +305,16 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                             _ => throw new ArgumentException($"Unsupported style {ConfigurationSource.PhaseDisplayStyle}")
                         };
 
-                        count = count + 1;
+                        var count = System.Guid.NewGuid(); //randomly generate ID for each <p>
```
rmshaffer, line 308: Very minor, but consider calling this variable something like `elementId`, since it's not a count anymore. Also, it may be convenient to include the full element ID in this variable, instead of duplicating the prefix a few times as is currently done below. For example, this line could become: 
```
var elementId = $"round{System.Guid.NewGuid()}"; //randomly generate ID for each <p>
```
And then below, you can just use `{elementId}` in each place, instead of `round${count}`.

src/Kernel/client/kernel.ts
```diff
@@ -133,10 +133,16 @@ class Kernel {
             "iqsharp_state_dump",
             message => {
                 console.log("my message received", message);
-                //message.content.amplitudes;
-                var htmlGraph = "hi :)";
+                //message.content.state.amplitudes;
+                function printNQubits() {
+                    document.getElementById("ex").innerHTML = message.content.n_qubits;
+                }
+                var htmlGraph = `<button onclick="printNQubits()"> NQubits </button> <p id="ex"> </p>`;
```
cgranade, line 140: Using the `onclick` attribute to reference the named function `printNQubits` could cause issues, since it's the same function name for every div that gets printed. My suggestion would be to create buttons as DOM objects so that you can attach unique event handlers to each.

For example:
```JavaScript
let div = document.getElementById(message.content.state.div_id);
if (div !== null) {
    let button = document.createElement("button");
    let p = document.createElement("p");
    button.appendChild(document.createTextNode("NQubits"));
    button.addEventListener("click", event => {
         p.innerHtml = message.content.n_qubits;
    });
    div.appendChild(button);
    div.appendChild(p);
}
```

Reemeela2121: can I still reference the created attributes in strings I set as inner html?

cgranade: Creating elements as DOM objects lets you get to the attributes of elements using methods instead of string processing (https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute), which can be helpful to prevent problems caused by improperly escaping strings (escaping _right_ is really hard).

Issue #158, opened 2020-06-04
[Feature Request] Circuit Visualization


theRoughCode:
**Is your feature request related to a problem? Please describe.**
As a developer, I might want to visualize the circuit associated with the execution of a Q# operation at any point in time. This would make debugging easier and even allow tutorials/notebooks to be more visual.

**Describe the solution you'd like**
The circuit should be displayed (as an SVG/HTML/image) based on a new magic command (i.e. `%print operation_name`). This `%print` command will follow similar semantics for arguments as `%simulate` with the result instead being the printing/visualization of the circuit.
The circuit displayed should represent the operations invoked starting directly from the `EntryPoint` up to `n` levels deep (where `n` is a configurable attribute via `%config`).
For certain well-known operations, we will show their custom display, such as:

- _SWAP_
- _M_
- _Measure_
- Operations under the `Microsoft.Quantum.Array` namespace
- _CNOT_
- Controlled operations (can be controlled on classical registers)
- Adjoint operations

Other operations will just show their name as a label (with non-qubit arguments) inside a box. The boxes should increase in width to fit the label.

**Additional context**
Other examples of quantum circuit visualizers:

- [Quantum Inspire](https://www.quantum-inspire.com/projects/new/)
- [quantum-circuit](https://quantum-circuit.com/example.html)
- [Bono](https://github.com/microsoft/bono)
- [Circuitizer](https://github.com/vadym-kl/qsharp-runtime/tree/circuitizer/src/Simulation/Circuitizer)

PR #195, opened 2020-07-02
Add `ExecutionPathTracer` to repo.

theRoughCode:
This PR addresses the first out of 3 components (Simulator, JavaScript renderer, magic command) required to implement #158 .

This PR adds the `ExecutionPathTracer` files and the extension method to attach to any simulator and trace out the operations performed in a given execution path.

This will be merged in to the feature branch (`feature/path_visualizer`).

Note: This should be merged in after updating the `qsharp-runtime` version to include [this PR](https://github.com/microsoft/qsharp-runtime/pull/301).

### DIFF
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -39,6 +39,7 @@
     <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.12.20071622-beta" />
     <PackageReference Include="NuGet.Resolver" Version="5.1.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
   </ItemGroup>
 
 </Project>
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -39,7 +39,6 @@
     <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.12.20071622-beta" />
     <PackageReference Include="NuGet.Resolver" Version="5.1.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
-    <PackageReference Include="System.Text.Json" Version="4.7.2" />
   </ItemGroup>
 
 </Project>
--- a/src/Core/ExecutionPathTracer/ExecutionPath.cs
+++ b/src/Core/ExecutionPathTracer/ExecutionPath.cs
@@ -4,7 +4,8 @@
 #nullable enable
 
 using System.Collections.Generic;
-using System.Text.Json;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
 {
@@ -31,11 +32,13 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <summary>
         /// A list of <see cref="QubitDeclaration"/> that represents the declared qubits used in the execution path.
         /// </summary>
+        [JsonProperty("qubits")]
         public IEnumerable<QubitDeclaration> Qubits { get; private set; }
 
         /// <summary>
         /// A list of <see cref="Operation"/> that represents the operations used in the execution path.
         /// </summary>
+        [JsonProperty("operations")]
         public IEnumerable<Operation> Operations { get; private set; }
 
         /// <summary>
@@ -45,12 +48,11 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// Pretty prints the JSON (i.e. with white space and indents) if <c>true</c>.
         /// </param>
         public string ToJson(bool prettyPrint = false) =>
-            JsonSerializer.Serialize(this,
-                new JsonSerializerOptions
+            JsonConvert.SerializeObject(this,
+                (prettyPrint) ? Formatting.Indented : Formatting.None,
+                new JsonSerializerSettings
                 {
-                    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                    IgnoreNullValues = true,
-                    WriteIndented = prettyPrint,
+                    NullValueHandling = NullValueHandling.Ignore,
                 }
             );
     }
@@ -69,7 +71,7 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <param name="numChildren">
         /// Number of associated classical registers.
         /// </param>
-        public QubitDeclaration(int id, int? numChildren = null)
+        public QubitDeclaration(int id, int numChildren = 0)
         {
             this.Id = id;
             this.NumChildren = numChildren;
@@ -78,12 +80,20 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <summary>
         /// Id of qubit.
         /// </summary>
+        [JsonProperty("id")]
         public int Id { get; private set; }
 
         /// <summary>
         /// Number of associated classical registers.
         /// </summary>
-        public int? NumChildren { get; private set; }
+        [JsonProperty("numChildren")]
+        public int NumChildren { get; private set; }
+
+        /// <summary>
+        /// Used by <see cref="Newtonsoft" /> to determine if <see cref="NumChildren" />
+        /// should be included in the JSON serialization.
+        /// </summary>
+        public bool ShouldSerializeNumChildren() => NumChildren > 0;
     }
 
     /// <summary>
@@ -94,6 +104,7 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <summary>
         /// Label of gate.
         /// </summary>
+        [JsonProperty("gate")]
         public string Gate { get; set; } = "";
 
         /// <summary>
@@ -107,26 +118,31 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <remarks>
         /// Currently not used as this is intended for classically-controlled operations.
         /// </remarks>
+        [JsonProperty("children")]
         public IEnumerable<IEnumerable<Operation>>? Children { get; set; }
 
         /// <summary>
         /// True if operation is a controlled operations.
         /// </summary>
+        [JsonProperty("controlled")]
         public bool Controlled { get; set; }
 
         /// <summary>
         /// True if operation is an adjoint operations.
         /// </summary>
+        [JsonProperty("adjoint")]
         public bool Adjoint { get; set; }
 
         /// <summary>
         /// List of control registers.
         /// </summary>
+        [JsonProperty("controls")]
         public IEnumerable<Register> Controls { get; set; } = new List<Register>();
 
         /// <summary>
         /// List of target registers.
         /// </summary>
+        [JsonProperty("targets")]
         public IEnumerable<Register> Targets { get; set; } = new List<Register>();
     }
 }
--- a/src/Core/ExecutionPathTracer/ExecutionPath.cs
+++ b/src/Core/ExecutionPathTracer/ExecutionPath.cs
@@ -1,10 +1,12 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+
+#nullable enable
 
 using System.Collections.Generic;
 using System.Text.Json;
 
-namespace Microsoft.Quantum.IQSharp
+namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
 {
     /// <summary>
     /// Represents the qubit resources and operations traced out in an execution path of a Q# operation.
@@ -15,43 +17,42 @@ namespace Microsoft.Quantum.IQSharp
         /// Initializes a new instance of the <see cref="ExecutionPathTracer"/> class.
         /// </summary>
         /// <param name="qubits">
-        /// An array of <c>QubitDeclaration</c>s that represents the declared qubits used in the execution path.
+        /// A list of <see cref="QubitDeclaration"/> that represents the declared qubits used in the execution path.
         /// </param>
         /// <param name="operations">
-        /// An array of <c>Operation</c>s that represents the operations used in the execution path.
+        /// A list of <see cref="Operation"/> that represents the operations used in the execution path.
         /// </param>
-        public ExecutionPath(QubitDeclaration[] qubits, Operation[] operations)
+        public ExecutionPath(IEnumerable<QubitDeclaration> qubits, IEnumerable<Operation> operations)
         {
             this.Qubits = qubits;
             this.Operations = operations;
         }
 
         /// <summary>
-        /// An array of <c>QubitDeclaration</c>s that represents the declared qubits used in the execution path.
+        /// A list of <see cref="QubitDeclaration"/> that represents the declared qubits used in the execution path.
         /// </summary>
-        public QubitDeclaration[] Qubits { get; set; }
+        public IEnumerable<QubitDeclaration> Qubits { get; private set; }
 
         /// <summary>
-        /// An array of <c>Operation</c>s that represents the operations used in the execution path.
+        /// A list of <see cref="Operation"/> that represents the operations used in the execution path.
         /// </summary>
-        public Operation[] Operations { get; set; }
+        public IEnumerable<Operation> Operations { get; private set; }
 
         /// <summary>
-        /// Serializes <c>ExecutionPath</c> into its JSON representation.
+        /// Serializes <see cref="ExecutionPath"/> into its JSON representation.
         /// </summary>
         /// <param name="prettyPrint">
-        /// Pretty prints the JSON (i.e. with white space and indents) if true (default = false).
+        /// Pretty prints the JSON (i.e. with white space and indents) if <c>true</c>.
         /// </param>
-        public string ToJson(bool prettyPrint = false)
-        {
-            var options = new JsonSerializerOptions
-            {
-                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                IgnoreNullValues = true,
-                WriteIndented = prettyPrint,
-            };
-            return JsonSerializer.Serialize(this, options);
-        }
+        public string ToJson(bool prettyPrint = false) =>
+            JsonSerializer.Serialize(this,
+                new JsonSerializerOptions
+                {
+                    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                    IgnoreNullValues = true,
+                    WriteIndented = prettyPrint,
+                }
+            );
     }
 
     /// <summary>
@@ -77,12 +78,12 @@ namespace Microsoft.Quantum.IQSharp
         /// <summary>
         /// Id of qubit.
         /// </summary>
-        public int Id { get; set; }
+        public int Id { get; private set; }
 
         /// <summary>
         /// Number of associated classical registers.
         /// </summary>
-        public int? NumChildren { get; set; }
+        public int? NumChildren { get; private set; }
     }
 
     /// <summary>
@@ -93,12 +94,12 @@ namespace Microsoft.Quantum.IQSharp
         /// <summary>
         /// Label of gate.
         /// </summary>
-        public string Gate { get; set; }
+        public string Gate { get; set; } = "";
 
         /// <summary>
         /// Non-qubit arguments provided to gate.
         /// </summary>
-        public string ArgStr { get; set; }
+        public string? ArgStr { get; set; }
 
         /// <summary>
         /// Group of operations for each classical branch.
@@ -106,7 +107,7 @@ namespace Microsoft.Quantum.IQSharp
         /// <remarks>
         /// Currently not used as this is intended for classically-controlled operations.
         /// </remarks>
-        public List<List<Operation>> Children { get; set; }
+        public IEnumerable<IEnumerable<Operation>> Children { get; set; } = new List<List<Operation>>();
 
         /// <summary>
         /// True if operation is a controlled operations.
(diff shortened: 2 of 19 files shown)


src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+                .OrderBy(k => k)
+                .Select(k =>
+                {
+                    var qubitDecl = new QubitDeclaration(k);
+                    if (this.classicalRegisters.ContainsKey(k))
+                    {
+                        qubitDecl.NumChildren = this.classicalRegisters[k].Count;
+                    }
+
+                    return qubitDecl;
```
cgranade, line 56: Won't help directly in this case, but check out [C# 9 with-expressions](https://devblogs.microsoft.com/dotnet/welcome-to-c-9-0/#with-expressions).

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            var t = operation.GetType();
+            if (t == null) return false;
+
+            return t.IsGenericType && t.GetGenericTypeDefinition() == typeof(OperationPartial<,,>);
+        }
+
+        /// <summary>
+        /// Retrieves the <c>QubitRegister</c> associated with the given <c>Qubit</c> or create a new
+        /// one if it doesn't exist.
```
cgranade, line 104: It's a bit confusing for a "Get★" method to have a side-effect of creating something...

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+                this.qubitRegisters[qubit.Id] = new QubitRegister(qubit.Id);
+            }
+
+            return this.qubitRegisters[qubit.Id];
+        }
+
+        private List<QubitRegister> GetQubitRegisters(IEnumerable<Qubit> qubits)
+        {
+            return qubits.Select(this.GetQubitRegister).ToList();
+        }
```
cgranade, line 119:
```suggestion
        private List<QubitRegister> GetQubitRegisters(IEnumerable<Qubit> qubits) =>
            qubits.Select(this.GetQubitRegister).ToList();
```

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+
+        /// <summary>
+        /// Creates a new <c>ClassicalRegister</c> and associate it with the given <c>Qubit</c>.
+        /// </summary>
+        private ClassicalRegister CreateClassicalRegister(Qubit measureQubit)
```
cgranade, line 124: This feels similar enough to the previous that you really want an extension method on `IDictionary<TKey, TValue>` that gets or creates a key.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            var qId = measureQubit.Id;
+
+            if (!this.classicalRegisters.ContainsKey(qId))
+            {
+                this.classicalRegisters[qId] = new List<ClassicalRegister>();
+            }
+
+            var cId = this.classicalRegisters[qId].Count;
+            ClassicalRegister register = new ClassicalRegister(qId, cId);
```
cgranade, line 134:
```suggestion
            var register = new ClassicalRegister(qId, cId);
```

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+            return t.IsGenericType && t.GetGenericTypeDefinition() == typeof(OperationPartial<,,>);
+        }
+
+        /// <summary>
+        /// Retrieves the <c>QubitRegister</c> associated with the given <c>Qubit</c> or create a new
+        /// one if it doesn't exist.
+        /// </summary>
+        private QubitRegister GetQubitRegister(Qubit qubit)
+        {
+            if (!this.qubitRegisters.ContainsKey(qubit.Id))
```
cgranade, line 108: A lot of this could be consolidated by using the `TryGetValue` method with an `out var` to capture the value if it exists.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+
+            // Get most recent measurement on given control qubit
+            var cId = this.classicalRegisters[qId].Count - 1;
+            return this.classicalRegisters[qId][cId];
+        }
+
+        /// <summary>
+        /// Given a <c>Type</c> and its value, extract its fields and format it as a string.
+        /// </summary>
+        private string? ExtractArgs(Type t, object value)
```
cgranade, line 164: Does this need to be an instance method, or can it be a static extension method?

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            List<string?> fields = new List<string?>();
```
cgranade, line 166:
```suggestion
            var fields = new List<string?>();
```

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,307 @@
...
+            foreach (var f in t.GetFields())
+            {
+                // If field is a tuple, recursively extract its inner arguments and format as a tuple string.
+                if (f.FieldType.IsTuple())
+                {
+                    var nestedArgs = f.GetValue(value);
+                    if (nestedArgs != null)
+                    {
+                        var nestedFields = this.ExtractArgs(f.FieldType, nestedArgs);
+                        fields.Add(nestedFields);
```
cgranade, line 177: Having a `.Add` method in a foreach loop is a great code smell for an opportunity to use `Select` instead. The trick here will be that you may want SelectMany to flatten out tuples...

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
cgranade, line 164: The name `Extract` doesn't communicate that this works to format as a string as well.

cgranade, line 164: When should this method return null?

src/Core/ExecutionPathTracer/Register.cs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/Core/ExecutionPathTracer/Register.cs
```diff
@@ -0,0 +1,77 @@
...
+// Licensed under the MIT License.
+
```
cgranade, line 3:
```suggestion
#nullable enable

```

src/Core/ExecutionPathTracer/Register.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Microsoft.Quantum.IQSharp
+{
+    /// <summary>
+    /// Enum for the 2 types of registers: Qubit and Classical.
+    /// </summary>
+    public enum RegisterType
+    {
+        Qubit,
+        Classical,
```
cgranade, line 12: Since this is a public enum, some API documentation comments would be really helpful here.

src/Core/Core.csproj
```diff
@@ -34,11 +34,13 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="3.0.0" />
+    <PackageReference Include="Microsoft.Quantum.Canon" Version="0.5.1904.1302" />
```
rmshaffer, line 37: Just curious, what are you using from this package? I don't think `Microsoft.Quantum.Canon` is what you want, as that one has been deprecated for quite some time.

cgranade: If you're looking for `typeof(Microsoft.Quantum.Canon.ApplyToEach)`, you'll want Microsoft.Quantum.Standard here instead. That said, the libraries repo depends on IQ# such that that then creates a circular dependency between repositories.

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+        public ExecutionPath(QubitDeclaration[] qubits, Operation[] operations)
+        {
+            this.Qubits = qubits;
+            this.Operations = operations;
+        }
+
+        /// <summary>
+        /// An array of <c>QubitDeclaration</c>s that represents the declared qubits used in the execution path.
+        /// </summary>
+        public QubitDeclaration[] Qubits { get; set; }
```
rmshaffer, line 32: Can these properties use `private set;` to prevent consumers from changing them after object creation?

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+            this.NumChildren = numChildren;
+        }
+
+        /// <summary>
+        /// Id of qubit.
+        /// </summary>
+        public int Id { get; set; }
+
+        /// <summary>
+        /// Number of associated classical registers.
```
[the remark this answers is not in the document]
rmshaffer: Same question... if you can just use `0` as a default instead of `null`, it may be simpler, unless `0` and `null` have different meanings here.

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </summary>
+        public string ArgStr { get; set; }
+
+        /// <summary>
+        /// Group of operations for each classical branch.
+        /// </summary>
+        /// <remarks>
+        /// Currently not used as this is intended for classically-controlled operations.
+        /// </remarks>
+        public List<List<Operation>> Children { get; set; }
```
rmshaffer, line 109: Can this use `IEnumerable` rather than `List`? Also, similar comment as elsewhere about using `private set` wherever possible.

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// List of control registers.
+        /// </summary>
+        public List<Register> Controls { get; set; } = new List<Register>();
+
+        /// <summary>
+        /// List of target registers.
+        /// </summary>
+        public List<Register> Targets { get; set; } = new List<Register>();
```
rmshaffer, line 129: Same here, consider using `IEnumerable` instead of `List` for the property type. Of course, using `new List` is still totally fine. 😊

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExecutionPathTracer"/> class.
+        /// </summary>
+        /// <param name="qubits">
+        /// An array of <c>QubitDeclaration</c>s that represents the declared qubits used in the execution path.
+        /// </param>
+        /// <param name="operations">
+        /// An array of <c>Operation</c>s that represents the operations used in the execution path.
+        /// </param>
+        public ExecutionPath(QubitDeclaration[] qubits, Operation[] operations)
```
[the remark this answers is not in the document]
theRoughCode: Yea, they can be `IEnumerable`s as long as the JSON serializer converts it into an array (which should be the case).

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <summary>
+    /// Represents the qubit resources and operations traced out in an execution path of a Q# operation.
+    /// </summary>
+    public class ExecutionPath
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExecutionPathTracer"/> class.
+        /// </summary>
+        /// <param name="qubits">
+        /// An array of <c>QubitDeclaration</c>s that represents the declared qubits used in the execution path.
```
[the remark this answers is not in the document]
theRoughCode: Will do!

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Ah, I guess that could've been an artifact of what it was doing initially (extracting it to a list of string). I guess `ArgsToString` would be a better name.

theRoughCode: It returns null if there aren't any arguments found. This was so that the `argStr` value in the resulting `Operation` object would be null and wouldn't be rendered in the final JSON string.

src/Core/Core.csproj
theRoughCode: Yea, @cgranade is right, it was to resolve `Microsoft.Quantum.Canon.ApplyToEach`. I did import `Microsoft.Quantum.Standard` in the first commit, but that broke some tests. Any ideas on why that broke it vs using `Microsoft.Quantum.Canon` would be appreciated!

I guess moving it to `qsharp-runtime` would help solve this. cc-ing @anpaz-msft regarding thoughts on this.

src/Core/ExecutionPathTracer/ExecutionPath.cs
theRoughCode: Yea, these objects aren't modified after creation, so we can definitely do that.

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
theRoughCode: The `null` value here is so that this key doesn't get rendered in the JSON string. What are your thoughts on making it `null` so that it doesn't get rendered vs making it `0` for simplicity?

src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        /// <summary>
+        /// Label of gate.
+        /// </summary>
+        public string Gate { get; set; }
+
+        /// <summary>
+        /// Non-qubit arguments provided to gate.
+        /// </summary>
+        public string ArgStr { get; set; }
```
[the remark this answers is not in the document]
theRoughCode: We only want to render the non-qubit arguments in the argument string when we visualize them as the qubits they act on are already implied by the visualization. Would `Arguments` be a better property name?

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: That's cool! I guess we could use it with something like this:
```C#
qubitDecl = qubitDecl with {
  NumChildren = (this.classicalRegisters.ContainsKey(k))
    ? this.classicalRegisters[k].Count
    : null
};
```

cgranade: Yep! Doesn't work for us yet since this project is still on C# 8, but it's exciting to keep an eye on where new features can help in the future.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Would `GetOrCreate*` be better?

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Oh, interesting, that would make sense. Thanks!

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Do you mean something similar to [this](https://stackoverflow.com/questions/16192906/net-dictionary-get-or-create-new)? Not sure what you mean by using an `out var` to capture the value.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: This can definitely be a static method. Thanks!

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Ah, thanks for the tip! Still getting used to the tricks of C# :)

src/Core/ExecutionPathTracer/ExecutionPath.cs
theRoughCode: Thanks for the tip! Really appreciate it :D

src/Core/ExecutionPathTracer/ExecutionPath.cs
theRoughCode: I think in this case, since we don't have a constructor, we probably can't use `private set` here? Not sure what the convention is, but there would be a lot of arguments to pass into the construction for the `Operation` class, so I figured manually setting it would be the best way.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: I created an extension method like [this](https://stackoverflow.com/a/16193323) for reusability.

src/Core/Core.csproj
cgranade: @theRoughCode: The Q# libraries are my area, so I'm happy to help with and to discuss. The short version is that we have sometimes moved operations from the Q# standard libraries into the Q# core library to make them available to the simulation runtime and to IQ# (e.g.: `MResetZ`). This doesn't have any user-facing effect, as both the core and standard libraries are made available to Q# applications, but can help with dependency ordering. From that perspective, moving ApplyToEach* to the core so that they can be used here is a completely reasonable suggestion.

The flip side is that it's always a bit of an awkward cross-repo dance to move something from standard to core, so I'd want to be quite sure that's the best solution before proceeding.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
cgranade: That makes sense; I'd suggest including some version of that discussion in the documentation comment, perhaps as a `<returns>` element.

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
cgranade: I'm a bit mixed here, as there's some potential for confusion with Q# syntax. That is, qubits are just ordinary values of type `Qubit`, such that the input tuple would look confusing if you just pulled out the qubits. For example, how would this property represent an input of type [`FixedPoint`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arithmetic.fixedpoint), since that UDT is declared as having `(Int, Qubit[])` as its base type?

src/Core/Core.csproj
theRoughCode: @cgranade Yea, that makes sense. I was talking to Andres on a sync earlier and he suggested I restructure the core logic surrounding that parsing of each operation into the metadata `Operation` (i.e. `ParseOperation` function) into qsharp-runtime. I'll be looping you in as I make progress on that front. This PR might be reworked a bit till then, so I'll make this a draft PR for now to signify that.

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
theRoughCode: That would most likely (if represented as a tuple type) would return `(Int)`. So, if we have `Foo(FixedPoint)` the visualization would create a gate acting on `Qubit[]` wires and the label `Foo(Int)`.

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
rmshaffer: Is there any case where `0` would mean something different than `null`? My preference would be to just use `0` and make the property non-nullable, as `NumChildren == null` doesn't seem very intuitive. You can just ignore this if it's `0` when parsing the JSON, correct?

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
rmshaffer: In general I agree with @cgranade's comment, but in this particular case I like `Get*`, since it's basically just allowing a lazy-initialization of the `qubitRegisters` dictionary. Up to you, though, no big deal either way.

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
theRoughCode: Yea, I agree. I think I was looking into this for `System.Text.Json` which I'm using, but there isn't a nice way _yet_ to ignore a property based on a value. I _think_ this is possible with `Newtonsoft`?

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Cool, thanks for the input! I'll stick to `Get*` unless there are strong objections against this. :)

src/Core/ExecutionPathTracer/ExecutionPath.cs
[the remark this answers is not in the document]
theRoughCode: Changed the name to `DisplayArgs` to denote that these might not contain all arguments, only the arguments we want to display.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -0,0 +1,150 @@
...
+                Controls = this.GetQubitRegisters(metadata.Controls),
+                Targets = this.GetQubitRegisters(metadata.Targets),
+            };
+
+            // Create classical registers for measurement operations
+            if (metadata.IsMeasurement)
+            {
+                var measureQubit = metadata.Targets.ElementAt(0);
+                var clsReg = this.CreateClassicalRegister(measureQubit);
+                // TODO: Change this to using IsMeasurement
```
theRoughCode, line 141: Note: this will be addressed in a separate PR so as not to clutter this one.

PR #198, opened 2020-07-08
Inject chart.js and separate kernel.ts into submodules.

cgranade:


cgranade:
> This is nice! One thought, would it be reasonable to merge the `kernel.ts` refactoring directly to `master`, and then add `plotting.ts` to the feature branch?

That could make sense, yeah. My thought was that that refactoring is pretty coupled to separating out plotting.ts, but you're right that it could be decoupled that way.

rmshaffer:
> > This is nice! One thought, would it be reasonable to merge the `kernel.ts` refactoring directly to `master`, and then add `plotting.ts` to the feature branch?
> 
> That could make sense, yeah. My thought was that that refactoring is pretty coupled to separating out plotting.ts, but you're right that it could be decoupled that way.

@cgranade either way is fine with me, whatever you prefer. Thanks!

cgranade:
Sounds good, @rmshaffer. If it's alright, then, I'll go on and merge this in as-is since it's already written this way. Thanks!

PR #201, opened 2020-07-10
Add path visualizer JS files

theRoughCode:
This PR addresses the second out of 3 components (Simulator, JavaScript renderer, magic command) required to implement #158 .
This PR moves the files for the JS execution path visualizer from [ADO](https://dev.azure.com/t-rkoh/CircuitViz/_git/CircuitViz/pullrequest/8) into iqsharp (feature branch `feature/path_visualizer`).

Jest test files will be added as a separate PR.


src/Kernel/client/ExecutionPathVisualizer/formatters/gateFormatter.ts
```diff
@@ -0,0 +1,308 @@
...
+    if (y.length === 0) return "";
+
+    // Sort y in ascending order
+    y.sort((y1, y2) => y1 - y2);
+
+    // Group adjacent registers
+    let prevY: number = y[0];
+    const regGroups: number[][] = y.reduce((acc: number[][], currY: number) => {
+        // Registers are defined to be adjacent if they differ by registerHeight in their y coord
+        // Raphael: Is there a better way of doing this? (i.e. split into y and heights in processOperations)
```
rmshaffer, line 107: Just a reminder that this note is here, in case you wanted to do something about it.

src/Kernel/client/ExecutionPathVisualizer/formatters/gateFormatter.ts
```diff
@@ -0,0 +1,308 @@
...
+ */
+const _unitary = (label: string, x: number, y: number[], width: number, argStr?: string, renderDashedLine: boolean = true): string => {
+    if (y.length === 0) return "";
+
+    // Sort y in ascending order
+    y.sort((y1, y2) => y1 - y2);
+
+    // Group adjacent registers
+    let prevY: number = y[0];
+    const regGroups: number[][] = y.reduce((acc: number[][], currY: number) => {
```
rmshaffer, line 105: What does `acc` mean here? Maybe a longer name could be helpful for ignorant readers like myself 😊

src/Kernel/client/ExecutionPathVisualizer/pathVisualizer.ts
```diff
@@ -0,0 +1,121 @@
...
+    <!-- Control button text -->
+    .cls-control-unknown .cls-control-text {
+        fill: black;
+        stroke: none;
+    }
+    .cls-control-one .cls-control-text,
+    .cls-control-zero .cls-control-text {
+        fill: white;
+        stroke: none;
+    }
```
rmshaffer, line 92: Do you know how these colors behave under different system settings, e.g. high contrast mode?

src/Kernel/client/ExecutionPathVisualizer/formatters/gateFormatter.ts
theRoughCode: Thanks for the heads-up! I think that this method works for now, but it's there because it might break in the future if we want to add variable heights for each register. I can edit this comment to be more of a note rather than a "note-to-self".

src/Kernel/client/ExecutionPathVisualizer/formatters/gateFormatter.ts
theRoughCode: Oh oops, it's an accumulator 😅

src/Kernel/client/ExecutionPathVisualizer/pathVisualizer.ts
theRoughCode: Not really, what are the normal procedures for adding colours in?

rmshaffer: I'm not aware of any place in IQ# where we are specifying colors in this way; you may be breaking new ground. 😊 You could try it out quickly by just turning on High Contrast mode (in Windows 10, if you click Start and type "High Contrast" it should take you there). Basically we just want to make sure that accessibility settings like High Contrast are respected by our features. I am guessing that the browser will do approximately the right thing here, but it's worth testing.

cgranade: I'd also suggest using Edge / Chrome extensions to simulate common forms of colorblindness can help with accessibility as well.

src/Kernel/client/ExecutionPathVisualizer/fontSizes.ts
```diff
@@ -0,0 +1,587 @@
+// Retrieved from https://github.com/adambisek/string-pixel-width/blob/master/src/widthsMap.js
```
rmshaffer, line 1: Would it be possible to use the [string-pixel-width](https://www.npmjs.com/package/string-pixel-width) npm package instead of copying this code here?

cgranade: It turns out to be a little squirrly to import npm packages into the Jupyter Notebook client. I came up with one approach at https://github.com/microsoft/iqsharp/commit/510979a1196a63c6773a3a893a062b77909e81bd, but it's not great that it relies on deprecated TypeScript features...

theRoughCode: Good catch. This was mainly used for determining text width, which was needed for my prototype project in Node,js. But now that it's purely in-browser, we might be able to just use hidden divs as calculate the width from there.

theRoughCode: I made a new commit where I use in-browser DOM manipulation instead. This would require an additional dev package when we bring my jest tests into this repo.

src/Kernel/client/ExecutionPathVisualizer/pathVisualizer.ts
theRoughCode: Thanks for the feedback! I installed a chrome extension called "Colorblinding" (couldn't find an edge one) and the only case where the colours between "true" and "false" were not differentiable was with the "monochromatic" setting.

PR #202, opened 2020-07-10
Break test dependencies on downstream packages

anpaz-msft:
Currently, some of our tests depend on QDK packages that are built on other repositories that when built end-to-end are built afterwards. For example, the tests depends on loading at runtime the Microsoft.Quantum.Chemistry package which lives in QuantumLibraries, but this repo depends on IQ# for some of the magic commands, so we end up with a circular dependency.
To break this, I'm now mocking the NugetPackages class, so during the tests calling Add(package) is mostly a no-op. Also changed the tests accordingly.
Loading packages is still tested, it just that it depends on specific package versions.


src/Tests/NugetPackagesTests.cs
```diff
@@ -189,23 +194,37 @@ public async Task AddPackage()
...
 
                 // Make sure we're case insensitive.
-                await mgr.Add("microsoft.quantum.chemistry::0.11.2003.3107  ");
+                await mgr.Add($"microsoft.quantum.chemistry::{QDK_LIBRARIES_VERSION}  ");
                 Assert.AreEqual(start + 2, mgr.Items.Count());
             }
         }
+
+        [TestMethod]
+        public async Task AddInvaliPackage()
```
rmshaffer, line 217:
```suggestion
        public async Task AddInvalidPackage()
```

src/Core/Compiler/CompilerService.cs
```diff
@@ -98,8 +104,10 @@ public AssemblyInfo BuildEntryPoint(OperationInfo operation, CompilerMetadata me
         /// </summary>
         public AssemblyInfo BuildSnippets(Snippet[] snippets, CompilerMetadata metadatas, QSharpLogger logger, string dllName, string executionTarget = null)
         {
+            string open_stmtms = string.Join("", AutoOpenNamespaces.Select(ns => $"open {ns};"));
```
rmshaffer, line 107:
```suggestion
            string openStatements = string.Join("", AutoOpenNamespaces.Select(ns => $"open {ns};"));
```

src/Core/Compiler/CompilerService.cs
```diff
@@ -98,8 +104,10 @@ public AssemblyInfo BuildEntryPoint(OperationInfo operation, CompilerMetadata me
...
+                
             string WrapInNamespace(Snippet s) =>
-                $"namespace {Snippets.SNIPPETS_NAMESPACE} {{ open Microsoft.Quantum.Intrinsic; open Microsoft.Quantum.Canon; {s.code} }}";
+                $"namespace {Snippets.SNIPPETS_NAMESPACE} {{ {open_stmtms} {s.code} }}";
```
rmshaffer, line 110:
```suggestion
                $"namespace {Snippets.SNIPPETS_NAMESPACE} {{ {openStatements} {s.code} }}";
```

src/Core/References/NugetPackages.cs
```diff
@@ -108,7 +108,10 @@ public IEnumerable<SourceRepository> Repositories
         // packages to use, since all of them need to be in-sync.
         public IReadOnlyDictionary<string, NuGetVersion> DefaultVersions { get; }
 
-        public NugetPackages(IOptions<Settings>? config, Microsoft.Extensions.Logging.ILogger logger)
+        public NugetPackages(
+            IOptions<Settings> config,
+            ILogger<References> logger
```
rmshaffer, line 113:
```suggestion
            ILogger<NugetPackages> logger
```

src/Jupyter/Jupyter.csproj
```diff
@@ -6,6 +6,7 @@
     <RootNamespace>Microsoft.Quantum.IQSharp.Jupyter</RootNamespace>
     <AssemblyName>Microsoft.Quantum.IQSharp.Jupyter</AssemblyName>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn>
```
rmshaffer, line 9: Should we open an issue to add all missing documentation for publicly-accessible elements and re-enable this warning?

src/Mock.Chemistry/Mock.Chemistry.csproj
```diff
@@ -0,0 +1,11 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
+    <IncludeQsharpCorePackages>false</IncludeQsharpCorePackages>    <!-- otherwise the standard library is included by the Sdk -->
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.12.20063023-beta" />
+  </ItemGroup>
+</Project>
```
rmshaffer, line 11: Any reason we don't want to choose an official release version here, e.g. `0.12.20070124`? (for both the `Sdk` and the `PackageReference`)

src/Tests/Mocks.cs
```diff
@@ -9,6 +9,12 @@
 using Microsoft.Jupyter.Core.Protocol;
 using Microsoft.Quantum.IQSharp;
 using Microsoft.Extensions.Logging;
+using NuGet.Packaging.Core;
+using NuGet.Versioning;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Linq;
+
+#nullable enable
```
rmshaffer, line 17: [nit] move this directive to the top of the file?

src/Mock.Chemistry/Mock.Chemistry.csproj
anpaz-msft: It's just the version that iqsharp is currently using. We should upgrade the repo to the latest release if possible.

src/Mock.Chemistry/Library.qs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+///
```
cgranade, line 5: [nit] Since this is a `///` comment, it would be good for it to start with a `/// # Summary` line. Alternatively, could make them into ordinary comments instead.

src/AzureClient/AzureClient.cs
```diff
@@ -289,8 +289,11 @@ public async Task<ExecutionResult> SetActiveTargetAsync(IChannel channel, string
             // Set the active target and load the package.
             ActiveTarget = executionTarget;
 
-            channel.Stdout($"Loading package {ActiveTarget.PackageName} and dependencies...");
-            await References.AddPackage(ActiveTarget.PackageName);
+            if (!(ActiveWorkspace is MockAzureWorkspace))
+            {
+                channel.Stdout($"Loading package {ActiveTarget.PackageName} and dependencies...");
+                await References.AddPackage(ActiveTarget.PackageName);
+            }
```
rmshaffer, line 296: This makes sense for now. I've opened #223 to restore the test coverage for loading provider packages here.

src/Tests/SNIPPETS.cs
```diff
@@ -97,42 +97,41 @@ @ EntryPoint (  )
...
-    open Microsoft.Quantum.Canon;
-    open Microsoft.Quantum.Measurement;
+    open Tests.qss;
+    open Microsoft.Quantum.Diagnostics;
 ";
 
         public static string DependsOnMeasurementNamespace =
 @"
-    operation DependsOnMeasurementNamespace() : Result[]
+    operation DependsOnMeasurementNamespace() : Unit
```
rmshaffer, line 111: [nit] this and other references to `DependsOnMeasurementNamespace` can probably just become `DependsOnNamespace`, since it no longer actually depends on the `Measurement` namespace.

PR #204, opened 2020-07-11
Persist opened namespaces across snippets

### LINKED ISSUE #15: Error with standalone opens in Q# notebooks
I am super stoked about Q# notebooks, thanks for doing this! 😍 

I was just trying out some of the snippets from the [Intro To IQSharp](https://github.com/Microsoft/Quantum/blob/master/Samples/src/IntroToIQSharp/Notebook.ipynb) notebook and found something I think might be a bug. If you use any `open` statement in a cell by itself, it returns the following error message: `The given key was not present in the dictionary.`. It seems to work fine if other operations are in the cell with it, even if they do not use anything from the `open`.  I have included a screencap as well as code snippets and the %version from the notebook.
Thanks!
Sarah

Component | Version
-- | --
iqsharp | 0.5.1903.2902
Jupyter Core | 1.1.12077.0

```
open Microsoft.Quantum.Extensions.Math;
The given key was not present in the dictionary.
```
```
open Microsoft.Quantum.Extensions.Diagnostics;
The given key was not present in the dictionary.
```
```
open Microsoft.Quantum.Extensions.Math;
​
//operation HelloPi() : Unit {
//    let pi = Microsoft.Quantum.Extensions.Convert.ToStringD(PI());
//    HelloQ(pi);
//}
The given key was not present in the dictionary.
```
```
open Microsoft.Quantum.Extensions.Math;
​
operation HelloPi() : Unit {
    let pi = Microsoft.Quantum.Extensions.Convert.ToStringD(PI());
    HelloQ(pi);
}
HelloPi
```

```
open Microsoft.Quantum.Extensions.Math;
​
operation HelloPi() : Unit {
}
HelloPi
```
image

### LINKED ISSUE #121: Warnings for already open namespaces in cell
**Describe the bug**
When opening namespaces in Q# Jupyter notebooks, sometimes you get warnings that namespaces are already open. In particular this holds for the `Microsoft.Quantum.Canon` and the `Microsoft.Quantum.Intrinsic` namespaces which are probably already opened by some other process.

**To Reproduce**
1. Open a Q# Jupyter notebook.
2. Run the following code:
```
open Microsoft.Quantum.Canon;
open Microsoft.Quantum.Intrinsic;
open Microsoft.Quantum.Diagnostics;
```
The Diagnostics namespace is included to show you don't get a warning opening that namespace. You should get the following warning messages:
```
/snippet_.qs(1,90): warning QS6003: The namespace is already open.
/snippet_.qs(2,6): warning QS6003: The namespace is already open.
```

**Expected behavior**
I would expect this to either not have any message if they are already open, or to make you to open them and not have namespaces opened with out a user explicitly asking for them.

**Screenshots**
You can see the warnings that result here, and I included opening the diagnostic namespace to show no other warnings occur from opening that one.
image

**System information**
This can be reproduced in the current Q# kernel found at aka.ms/try-qsharp
image

rmshaffer:
This PR allows much more lenient handling of `open` directives in IQ# snippets:

1. Fixes #121 by disabling the compiler message `warning QS6003: The namespace is already open.` when opening a namespace that has already been opened. This would occur if a user tried to open one of the namespaces that is automatically opened for all snippets (`Microsoft.Quantum.Canon` or `Microsoft.Quantum.Intrinsic`).

1. Fixes #15 by tracking _all_ namespaces that have been opened in snippets in the current session and automatically opening these namespaces for all subsequent snippet compilations. This allows, for example, a user to write a cell at the top of the notebook with a bunch of `open` statements, and have those namespaces be automatically available for all future cell executions.

Note that some code here (specifically, the addition of the `AutoOpenNamespaces` property) is duplicated from #202. I will wait for that PR to be merged first.


src/Tests/SNIPPETS.cs
```diff
@@ -94,6 +94,31 @@ @ EntryPoint (  )
     }
 ";
 
+        public static string OpenNamespaces1 =
+@"
+    open Microsoft.Quantum.Intrinsic;
```
cgranade, line 99: Should there be a test to make sure that `open X and Y` is either supported, or fails with an actionable error message?

src/Core/Compiler/CompilerService.cs
```diff
@@ -98,11 +121,18 @@ public AssemblyInfo BuildEntryPoint(OperationInfo operation, CompilerMetadata me
         /// </summary>
         public AssemblyInfo BuildSnippets(Snippet[] snippets, CompilerMetadata metadatas, QSharpLogger logger, string dllName, string executionTarget = null)
         {
+            string openStatements = string.Join("", AutoOpenNamespaces.Select(ns => $"open {ns};"));
             string WrapInNamespace(Snippet s) =>
-                $"namespace {Snippets.SNIPPETS_NAMESPACE} {{ open Microsoft.Quantum.Intrinsic; open Microsoft.Quantum.Canon; {s.code} }}";
+                $"namespace {Snippets.SNIPPETS_NAMESPACE} {{ {openStatements} {s.code} }}";
```
cgranade, line 126: One issue we've run into before is that the line numbers and column offsets in error messages don't match up; as the wrapping code gets a bit longer, that may get worse. As a minor suggestion, would it be possible to set the offset when calling `BuildAssembly`?

src/Tests/SNIPPETS.cs
rmshaffer: Thanks for pointing this out! I hadn't considered the case of `open X as Y` directives. Yes, looks like I need to add support to avoid breaking that, ignore one more warning, and then I'll add tests for this case.

src/Core/Compiler/CompilerService.cs
rmshaffer: Yes, looks like this should be a simple enough change.

src/Core/Compiler/CompilerService.cs
```diff
@@ -34,30 +35,53 @@ namespace Microsoft.Quantum.IQSharp
...
+        {
+            var loader = CreateTemporaryLoader(source);
+            if (loader.VerifiedCompilation == null) { return ImmutableDictionary<string, string>.Empty; }
+            return loader.VerifiedCompilation.Tokenization.Values
+                .SelectMany(tokens => tokens.SelectMany(fragments => fragments))
+                .Where(fragment => fragment.Kind != null && fragment.Kind.IsOpenDirective)
+                .Select(fragment => ((QsFragmentKind.OpenDirective)fragment.Kind))
+                .ToDictionary(
+                    openDirective => openDirective.Item1.Symbol.AsDeclarationName(null),
+                    openDirective => openDirective.Item2.ValueOr(null)?.Symbol.AsDeclarationName(null));
```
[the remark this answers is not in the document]
rmshaffer: Certainly wouldn't hurt!

rmshaffer:
Thanks @cgranade for the review!

PR #205, opened 2020-07-13
Add JS tests for ExecutionPathVisualizer

theRoughCode:
This PR adds JS tests for the `ExecutionPathVisualizer` files using `Jest`. 
The tests are run using `npm test` within `src/Kernel`. This command is also added to `build/test.ps1` to run the tests during CI.


src/Kernel/package.json
```diff
@@ -1,6 +1,24 @@
...
+        "canvas": "^2.6.1",
+        "jest": "^26.1.0",
+        "ts-jest": "^26.1.2",
+        "typescript": "^3.9.6"
+    },
+    "jest": {
+        "transform": {
+            "^.+\\.tsx?$": "ts-jest"
+        },
+        "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.ts$",
```
rmshaffer, line 18: Should `__tests__` be `__test__` here, to match your folder structure? Also should `spec` be `snap`, or is that something different?

theRoughCode: Good catch! I guess it looks like the second half of the regex matches the incorrectly named folder. Fixing it!
`spec` is another naming convention for test files I believe. I'll remove it since it's not being used.

theRoughCode: actually, it seems like the [default case](https://jestjs.io/docs/en/configuration#testregex-string--arraystring) already matches my files, so I'll just remove this line

PR #206, opened 2020-07-13
Port chemistry Python library to QuantumLibraries

anpaz-msft:
In order to break inter-repo dependencies, we're migrating the chemistry portion of the Python `qsharp` package into the `QuantumLibraries` repo.
As part of this change, we're renaming the Python package created in this repo as `qsharp-core` and converting `qsharp` to be a meta-package that includes both `qsharp-core` and `qsharp-chemistry`. This metapackage will live on QuantumLibraries too.


conda-recipes/qsharp/build.ps1
```diff
@@ -1,10 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-# We need to set the PYTHON_VERSION environment variable
-# explicitly here, since conda-build doesn't by default pass
-# environment variables from the host environment.
-$Env:PYTHON_VERSION = $Env:PKG_VERSION
-Push-Location src/src/Python/
-    python setup.py install --prefix $Env:PREFIX
+Push-Location src/src/Python/dist
```
cgranade, line 4: If I understand right, this makes the `qsharp` conda package include all `qsharp-*` PyPI packages, rather than having a conda package for each PyPI package and then adding a `qsharp` conda metapackage?

build/pack-conda.ps1
```diff
@@ -66,7 +66,23 @@ function Pack-CondaRecipe() {
...
+function Copy-Wheels {
+    param(
+        [string] $Path
+    );
+
+    if ($Env:PYTHON_OUTDIR -ne "") {
+        If (-not (Test-Path -Path $Path)) { 
+            $_ = [IO.Directory]::CreateDirectory($Path) 
+        }    
+        Copy-Item $Env:PYTHON_OUTDIR\*.whl $Path
```
rmshaffer, line 78:
```suggestion
        Copy-Item $Env:PYTHON_OUTDIR/*.whl $Path
```

src/Python/qsharp-core/README.md
```diff
@@ -1,4 +1,4 @@
-# Python Interoperability for Q# #
+# Q# Interoperability for Python #
 
 The `qsharp` package for Python provides interoperability with the Quantum Development Kit and with the Q# language, making it easy to simulate Q# operations and functions from within Python.
```
rmshaffer, line 3:
```suggestion
The `qsharp-core` package for Python provides interoperability with the Quantum Development Kit and with the Q# language, making it easy to simulate Q# operations and functions from within Python.
```

src/Python/qsharp-core/README.md
```diff
@@ -15,7 +15,7 @@ conda install -c quantum-engineering qsharp
 ## Installing from Source ##
 
 If you'd like to contribute to or experiment with the Python interoperability feature, it may be useful to install from source rather than from the `qsharp` package on the Python Package Index (PyPI).
```
rmshaffer, line 17:
```suggestion
If you'd like to contribute to or experiment with the Python interoperability feature, it may be useful to install from source rather than from the `qsharp-core` package on the Python Package Index (PyPI).
```

src/Python/qsharp-core/README.md
```diff
@@ -33,4 +33,4 @@ python setup.py bdist_wheel
 ```
 
 By default, this will create a `qsharp` wheel in `dist/` with the version number set to 0.0.0.1.
```
rmshaffer, line 35:
```suggestion
By default, this will create a `qsharp-core` wheel in `dist/` with the version number set to 0.0.0.1.
```

conda-recipes/qsharp/build.ps1
anpaz-msft: Yes, that was my train of thought, have a kitchen-sink conda package with all that we have, not like we have that much :P. I don't see much value of having fine-grain packages but I could be easily convinced otherwise.

rmshaffer: I think this approach makes sense. If we find a need for finer-grained conda packages down the road, we could always add them. But given that the Python libraries are small at the moment, I'm not sure there's a huge benefit there. One downside I guess is that the conda packages built from the CI builds in iqsharp will no longer contain the chemistry library, but I don't think that's a big deal, as one could manually install the `qsharp-chemistry` package if needed.

cgranade: Agreed. My only point of concern is the asymmetry between the two, such that the PyPI `qsharp-chemistry` package has no direct `conda` analogue. I think that can be addressed with documentation and by primarily promoting the `qsharp` PyPI metapackage.

PR #207, opened 2020-07-13
T relara/measurement probability histogram graph

Reemeela2121:
This is still a work in progress, below is a list of things that need to be added/implemented: 

- [x] remove buttons (were originally added for proof of concept/debugging)
- [x] add X,Y axis labeling
- [x] Options to show {Amplitude, phase} and {Real, Imag}


rmshaffer:
Quick note: You'll want to target `feature/visualization-improvements` for merging this PR, rather than `master`, since you're building on other changes that are currently in `feature/visualization-improvements`.

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
                         });
                         div.appendChild(button);
                         div.appendChild(p);
+                        // todo: make a new div
```
cgranade, line 47: Has this been resolved?

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
...
+                        
+                        let button2 = document.createElement("button");
+                        let printthingy = document.createElement("p");
```
cgranade, line 50: Can we give these variables more descriptive names?

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
...
+
+                        let graph = document.createElement("canvas");
+                        button2.appendChild(document.createTextNode("Show Graph"));
+                        button.addEventListener("click", event => {
+                            var txt = "";
+                            var x;
+                            var amps = state.amplitudes;
+
+                            let newCount = amps.length;
+                            let nQubits = Math.log2(newCount) >>> 0;
```
cgranade, line 60: I think the DisplayableState interface has a number of qubits directly?

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
...
+                            let bit = [];
+                            bit = Array.from(Array(amps.length).keys()).map(idx => {
```
cgranade, line 62:
```suggestion
                            let bit = Array.from(Array(amps.length).keys()).map(idx => {
```

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
...
+                                var bitstring = (idx >>> 0).toString(2).padStart(nQubits, "0");
```
cgranade, line 63: What does right-shifting by zero bits do here?

src/Kernel/client/kernel.ts
```diff
@@ -43,8 +44,36 @@ class Kernel {
...
+                                return `|${bitstring}⟩`;
+                            });
+                            for (x in amps) {
```
cgranade, line 66: Similarly here, a more descriptive variable name than `x` would be helpful.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,79 @@
...
 
+export interface Complex {
+    Real: number;
+    Imag: number;
+    Magnitude: number;
+    Phase: number;
+};
+
+export interface DisplayableState {
+    n_qubits: number;
```
cgranade, line 25: Right, here's where the `n_qubits` property is defined; shouldn't need to take the log of the length of amplitudes.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,79 @@
...
+    div_id: string;
+    amplitudes: Complex[] | null;
+};
+
+
 export function initializePlotting() {
-    importScript("https://cdn.jsdelivr.net/npm/chart.js@2.8.0");
-}
+    // importScript("https://cdn.jsdelivr.net/npm/chart.js@2.8.0");
```
cgranade, line 32: This could probably be removed now.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,79 @@
...
+};
+
+
+
+export function createBarChart(element: HTMLCanvasElement, state: DisplayableState) {
+    var amps = state.amplitudes;
+
+    let newCount = amps.length;
+    let nQubits = Math.log2(newCount) >>> 0;
```
cgranade, line 41: Similarly here, this should be replaced by looking at `state.n_qubits`.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,79 @@
...
 export function initializePlotting() {
-    importScript("https://cdn.jsdelivr.net/npm/chart.js@2.8.0");
-}
+    // importScript("https://cdn.jsdelivr.net/npm/chart.js@2.8.0");
+};
+
+
+
+export function createBarChart(element: HTMLCanvasElement, state: DisplayableState) {
+    var amps = state.amplitudes;
```
cgranade, line 38: I'd suggest keeping to `let` over `var`; I recognize that's controversial, but it'd at least be good to not mix `let` and `var` together.

src/Kernel/client/kernel.ts
```diff
@@ -31,20 +31,43 @@ class Kernel {
                 console.log("my message received", message);
 
                 //create buttons as DOM objects in order to attach unique event handlers
-                let state_div = message.content.state.div_id;
+                let state: DisplayableState = message.content.state;
+                let state_div = state.div_id;
                 if (state_div != null) {
                     let div = document.getElementById(state_div);
                     if (div != null) {
+                        let button1 = document.createElement("button");
```
cgranade, line 39: I recognize this button is only used for a short period before getting added to `div`, but it may still be helpful to give it a more descriptive variable name.

src/Kernel/client/kernel.ts
```diff
@@ -31,20 +31,43 @@ class Kernel {
...
+                        let realImagGraph = document.createElement("canvas");
+                        button.appendChild(document.createTextNode("Show Basis States vs Real,Imag"));
+                        button.addEventListener("click", event => { 
+                            createBarChartRealImagOption(realImagGraph, state);
+                            div.appendChild(realImagGraph);
                         });
                         div.appendChild(button);
-                        div.appendChild(p);
+
+                        let button2 = document.createElement("button");
```
cgranade, line 58: Similarly here and at line 49, more descriptive names could help differentiate the three buttons.

src/Kernel/Kernel.csproj
```diff
@@ -29,10 +29,11 @@
   <ItemGroup>
     <EmbeddedResource Include="res\kernel.js" />
     <EmbeddedResource Include="res\bundle.js" />
+    <EmbeddedResource Include="node_modules\chart.js\dist\Chart.bundle.min.js" Link="res\chart.js" />
     <Content Include="tsconfig.json">
       <CopyToOutputDirectory>Never</CopyToOutputDirectory>
     </Content>
-    <PackageReference Include="Microsoft.TypeScript.MSBuild" Version="3.7.4">
+    <PackageReference Include="Microsoft.TypeScript.MSBuild" Version="3.9.6">
```
theRoughCode, line 36: What was this version change for?

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+export interface Complex {
+    Real: number;
+    Imag: number;
+    Magnitude: number;
+    Phase: number;
+};
+
+export interface DisplayableState {
+    n_qubits: number;
+    div_id: string;
```
theRoughCode, line 26: Not sure what the style guide is on this team, but in most cases, JavaScript uses CamelCase.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+
+
+
+export function createBarChart(element: HTMLCanvasElement, state: DisplayableState) {
+    var amps = state.amplitudes;
+
+    let newCount = amps.length;
+    let nQubits = Math.log2(newCount) >>> 0;
+
+    var measurementHistogram = new Chart(element, {
```
theRoughCode, line 43: I would use `const` here if we know that it won't be mutated and `let` otherwise.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+        type: 'bar',
+        data: {
+            labels: Array.from(Array(amps.length).keys()).map(idx => {
+                var bitstring = (idx >>> 0).toString(2).padStart(nQubits, "0");
```
theRoughCode, line 47: Same here, if it's mutable, I would use `let`. Otherwise, I would use `const`. In general, `const`/`let` is preferred over `var`.

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+                {
+                    data: Array.from(Array(amps.length).keys()).map(idx => {
+                        return (amps[idx].Real);
+                    }),
+                    backgroundColor: "#5390d9",
+                    borderColor: "#5390d9",
+                    label: "Real"
+                },
+                {
+                    data: Array.from(Array(amps.length).keys()).map(idx => {
```
theRoughCode, line 116:
```suggestion
data: amps.map(amp => amp.Imag),
```

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+    var measurementHistogram = new Chart(element, {
+        type: 'bar',
+        data: {
+            labels: Array.from(Array(amps.length).keys()).map(idx => {
+                var bitstring = (idx >>> 0).toString(2).padStart(nQubits, "0");
+                return `|${bitstring}⟩`;
+            }), //basis state labels
+            datasets: [
+                {
+                    data: Array.from(Array(amps.length).keys()).map(idx => {
```
theRoughCode, line 172:
```suggestion
amps.map(amp => amp.Magnitude),
```

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,219 @@
...
+                        return (amps[idx].Magnitude);
+                    }),
+                    backgroundColor: "#4c4cff",
+                    borderColor: "#4c4cff",
+                    label: "Amplitude"
+                },
+                {
+                    data: Array.from(Array(amps.length).keys()).map(idx => {
```
theRoughCode, line 180:
```suggestion
amps.map(amp => amp.Phase),
```

theRoughCode:
Looks good! Just left a few comments for you to go over :)

src/Kernel/client/kernel.ts
```diff
@@ -31,20 +31,43 @@ class Kernel {
                 console.log("my message received", message);
 
                 //create buttons as DOM objects in order to attach unique event handlers
-                let state_div = message.content.state.div_id;
+                let state: DisplayableState = message.content.state;
```
theRoughCode, line 34: If `state` is not going to be changed, I would use `const` instead. Same for the other variables. This helps other devs after you know the intent that the variable should not change!

src/Kernel/client/plotting.ts
theRoughCode: +1 to not using `var`. To go one step further, it looks like `amps` is not being mutated, so I would use `const`.

src/Kernel/client/plotting.ts
cgranade: This is an odd one. Our style guide is generally to use `camelCase`, as you note, but this interface constrains the type of a Jupyter message. Since Jupyter uses `snake_case` for its JSON serialization schemas, this interface uses `snake_case` to match as an intentional exception to the typical style guide.

theRoughCode: Ah gotcha, that makes sense. Thanks!

src/Kernel/Kernel.csproj
cgranade: This was needed to get the new `import type` statement.

src/Kernel/client/kernel.ts
Reemeela2121: changed the names of all the buttons to something more descriptive!

src/Kernel/client/plotting.ts
Reemeela2121: gotcha! changed all of them to const

src/Kernel/client/plotting.ts
Reemeela2121: changed all of them to let :) thanks!

src/Kernel/client/plotting.ts
Reemeela2121: changed var to let

src/Kernel/client/plotting.ts
Reemeela2121: got rid of the two variables and just used state.n_qubits

src/Kernel/client/plotting.ts
Reemeela2121: removed the entire function

cgranade:
Looks good to me! Thanks for all your work addressing comments and suggestions!

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,202 @@
...
+                    },
+                    ticks: {
+                        maxRotation: 0,
+                        minRotation: 0
+                    }
+                }],
+                yAxes: [{
+                    scaleLabel: {
+                        display: true,
+                        labelString: 'Meas. Probability'
```
rmshaffer, line 131: I think this label string should be `'Amplitude'`?

src/Kernel/client/plotting.ts
```diff
@@ -1,12 +1,202 @@
...
```
rmshaffer, line 192: Maybe this label string should be something more generic like `'Value'`, which would apply to both the amplitude and phase datasets?

PR #208, opened 2020-07-14
Replacing logo asset for Jupyter kernel.

ricardo-espinoza:
This change is replacing the Möbius icon for the new Q# asset.


src/Kernel/Kernel.csproj
```diff
@@ -10,15 +10,15 @@
 
   <ItemGroup>
     <None Remove="res\kernel.js" />
-    <None Remove="res\logo-64x64.png" />
```
rmshaffer, line 13: The Jupyter kernelspec property name for the logo is automatically derived from this filename, and Jupyter relies on this being named `logo-64x64.png` in order for it to show up as the logo. It may be easiest if you just add the new logo with the same filename.

ricardo-espinoza: Ah, thanks for the catch. Fixing. :)

Issue #25, opened 2019-07-13
Print code for operations in the current workspace


cgranade:
**Is your feature request related to a problem? Please describe.**
It would be helpful to have a `%printcode` magic command that prints and possibly syntax highlights the source code to a Q# operation in the current workspace.

**Additional context**
(h/t @vadym-kl for suggesting this feature!)

PR #209, opened 2020-07-14
Add trace magic command

theRoughCode:
This PR addresses the last out of 3 components (Simulator, JavaScript renderer, magic command) required to implement #158 .

This PR adds the `%trace` magic command required to render the execution path within IQ#.

This will be merged in to the feature branch (`feature/path_visualizer`).

Example screenshot of magic command:
image


src/Jupyter/Visualization/ExecutionPathEncoder.cs
```diff
@@ -0,0 +1,42 @@
+using Microsoft.Jupyter.Core;
+using Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer;
+
+namespace Microsoft.Quantum.IQSharp.Jupyter
+{
+    /// <summary>
+    ///     Contains the ID of the div used to populate the
+    ///     <see cref="ExecutionPath"/> visualization.
+    /// </summary>
+    public class ExecutionPathDisplayable
```
cgranade, line 10: Would it be worth generalizing this so that it can be used by other IQ# features that need to output a `<div>` with a known `id` and `class` attributes? Might be good to work with @Reemeela2121 to come up with a common solution.

src/Kernel/Magic/ViewMagic.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            this.Json = json;
+            this.Id = id;
+        }
+
+        /// <summary>
+        ///     JSON representation of the <see cref="ExecutionPath"/>.
+        /// </summary>
+        [JsonProperty("json")]
+        public string Json { get; }
```
cgranade, line 35: It looks like this will cause objects to get serialized to JSON twice, with the message content containing a JSON string whose value is a JSON object. It may be good to make this something like an `object` to rely on NewtonsoftJson correctly finding the right serializer, or something like `JToken` if you want to serialize it manually before setting it as the payload for this content.

src/Kernel/Magic/ViewMagic.cs
```diff
@@ -0,0 +1,130 @@
...
+                .WithStackTraceDisplay(channel)
+                .WithExecutionPathTracer(tracer);
+            var value = await symbol.Operation.RunAsync(qsim, inputParameters);
+
+            // Retrieve the `ExecutionPath` traced out by the `ExecutionPathTracer`
+            var executionPath = tracer.GetExecutionPath();
+
+            // Render empty div with unique ID as cell output
+            var divId = $"execution-path-container-{Guid.NewGuid().ToString()}";
+            var content = new ExecutionPathVisualizerContent(executionPath.ToJson(), divId);
```
cgranade, line 112: Following up on my comment above, the `ToJson()` call here is what causes the double-serialization. Converting to a JToken would allow composing the two JSON serialziations instead of nesting them.

src/Kernel/Magic/ViewMagic.cs
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
cgranade, line 3:
```suggestion
#nullable enable

```

src/Kernel/Magic/ViewMagic.cs
```diff
@@ -0,0 +1,130 @@
...
+    public class ViewMagic : AbstractMagic
+    {
+        private const string ParameterNameOperationName = "__operationName__";
+
+        /// <summary>
+        ///     Constructs a new magic command given a resolver used to find
+        ///     operations and functions, and a configuration source used to set
+        ///     configuration options.
+        /// </summary>
+        public ViewMagic(ISymbolResolver resolver, IConfigurationSource configurationSource) : base(
```
cgranade, line 57: The name `%view` is a bit generic; what is it that I'm viewing? It may be good to make it slightly more verbose, such as `%view-trace`?

src/Jupyter/Visualization/ExecutionPathEncoder.cs
theRoughCode: Yea, that would be a good idea. Maybe a `DisplayableDiv` class?

src/Kernel/Magic/ViewMagic.cs
theRoughCode: Would `%visualize` make more sense? It conveys more of the idea of visualizing a Q# operation whereas `trace` might not be as obvious to the end user.

src/Jupyter/Visualization/ExecutionPathEncoder.cs
cgranade: That could work, yes. My concern would be that the division between displayable objects and result encoders is intended to keep presentation logic out of displayable objects as much as possible, such that including "Div" in the name would break that separation. That said, this is a weird case in that there's no meaning whatsoever to "printing a `<div>` element" when considering any MIME type other than text/html...

src/Kernel/Magic/ViewMagic.cs
theRoughCode: We could also open this up on our internal channel to get feedback from others regarding the best name for the magic command?

src/Jupyter/Visualization/ExecutionPathEncoder.cs
theRoughCode: cc-ing @anpaz-msft @rmshaffer in case they have thoughts on the best way to accomplish rendering an HTML string

src/Kernel/Magic/ViewMagic.cs
theRoughCode: cc-ing @anpaz-msft and @rmshaffer for visibility

src/Jupyter/Visualization/ExecutionPathEncoder.cs
rmshaffer: One idea: rename this class to something like `DisplayableHtmlElement` and the encoder to `DisplayableHtmlElementEncoder`? Then it's clear that it only makes sense from HTML.

src/Kernel/Magic/ViewMagic.cs
rmshaffer: In my mind, `%trace` is more clearly associated with a program execution path than `%visualize`, which has a bit of the same problem as `%view` in being a bit generic. I agree, more brainstorming would be good here.

rmshaffer: For context, there's also issue https://github.com/microsoft/iqsharp/issues/25 which is suggesting a command called `%printcode`. Maybe we come up with some consistent naming scheme that could be used for both scenarios. e.g. `%viewtrace` and `%viewcode`, or perhaps `%printtrace` and `%printcode`.

src/Kernel/Magic/ViewMagic.cs
theRoughCode: I think this was meant for `src/Jupyter/Visualization/ExecutionPathEncoder.cs`.

src/Kernel/Magic/ViewMagic.cs
```diff
@@ -0,0 +1,135 @@
...
+            "view",
+            new Documentation
+            {
+                Summary = "Outputs an HTML-based visualization of the execution path of a given operation."
+            })
```
rmshaffer, line 63: Before merging this PR, it would be great to expand this documentation a little bit to include markdown for `Description` and `Examples`, similar to other magic commands. Probably you can look at the [`%simulate` documentation](https://github.com/microsoft/iqsharp/blob/337e9e717aa83c1cb9ad420ea4d1e1183776475f/src/Kernel/Magic/Simulate.cs#L28-L63) as a good starting point, since you have the same parameter requirements for this magic command.

This info gets automatically published to the public API docs thanks to some work @cgranade did recently, and it is also viewable inside Jupyter Notebook by appending a `?` to the magic command name, e.g. `%view?`.

src/Kernel/Magic/ViewMagic.cs
theRoughCode: `#viewtrace` makes more sense to me than `%printtrace` because `print` gives me the impression that it is text-based rather than an interactive HTML element. On the other hand, `%printcode` might make more sense since it's text-based. 
Then again, I'm not the best at naming, so I'll defer the decision to you guys haha!

rmshaffer: Agreed, I didn't like `%printtrace` from the moment I typed it. 😊 Personally I like the simplicity of just `%trace`, since the word `view` doesn't seem like it adds much value, but I'd be fine either way.

theRoughCode: Yea, I think `%trace` sounds like a good option! @anpaz-msft @cgranade: thoughts on `%trace` instead of `%view`?

cgranade: I like `%trace` in that it helps reinforce that the output isn't a visualization of the program, but of a particular path through that program. We don't in general expect tracing and debugging tools, for instance, to show every possible path through a program; similarly, we shouldn't expect to be able to reconstruct a Q# operation from a visualization of its trace.

theRoughCode:
Changed the base branch due to a GIthub bug when the base branch changed from `raphael/tracer` to `feature/path_visualizer`.

PR #210, opened 2020-07-14
Trying to fix #134

### LINKED ISSUE #134: Metadata in autogenerated docs cause warnings
**Describe the bug**
As noted in https://github.com/MicrosoftDocs/quantum-docs-pr/pull/805, the autogenerated IQ# docs cause some warnings about metadata. While these warnings don't adversely affect renderings, we should probably make sure to fix them if only to reduce noise in documentation builds. 

**To Reproduce**
1. Build documentation using the build_docs.py script in this repo.
2. Upload docs to MicrosoftDocs/quantum-docs-pr via a PR.

**Expected behavior**
Documentation PR completes without warnings.

**Screenshots**

image

geduardo:
Not sure if this will work

### DIFF
--- a/build/docs/build_docs.py
+++ b/build/docs/build_docs.py
@@ -120,8 +120,9 @@ def format_as_document(magic, uid_base : str) -> MagicReferenceDocument:
     uid = f"{uid_base}.{safe_name}"
     metadata = {
         'title': f"{magic_name} (magic command)",
+        'author': rmshaffer,
         'uid': uid,
-        'ms.date': datetime.date.today().isoformat(),
+        'ms.date': datetime.date.today().strftime("%m-%d-%Y"),
         'ms.topic': 'article'
     }
     header = f"# `{magic_name}`"
@@ -196,13 +197,14 @@ def format_index(all_magics : Dict[str, MagicReferenceDocument], uid_base : str)
     )
     metadata = {
         'title': "IQ# Magic Commands",
+        'author': rmshaffer,
         'uid': f"{uid_base}.index",
-        'ms.date': datetime.date.today().isoformat(),
+        'ms.date': datetime.date.today().strftime("%m-%d-%Y"),
         'ms.topic': 'article'
     }
     return cleanup_markdown(f"""
 # IQ# Magic Commands
-
+{metadata}
 | Magic Command | Summary |
 |---------------|---------|
 {index_content}
--- a/build/docs/build_docs.py
+++ b/build/docs/build_docs.py
@@ -197,7 +197,7 @@ def format_index(all_magics : Dict[str, MagicReferenceDocument], uid_base : str)
     )
     metadata = {
         'title': "IQ# Magic Commands",
-        'author': rmshaffer,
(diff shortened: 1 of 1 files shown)


build/docs/build_docs.py
```diff
@@ -120,8 +120,9 @@ def format_as_document(magic, uid_base : str) -> MagicReferenceDocument:
     uid = f"{uid_base}.{safe_name}"
     metadata = {
         'title': f"{magic_name} (magic command)",
+        'author': rmshaffer,
```
cgranade, line 123:
```suggestion
        'author': 'rmshaffer',
```
You'll likely also want an `ms.author` here.

build/docs/build_docs.py
```diff
@@ -196,13 +197,14 @@ def format_index(all_magics : Dict[str, MagicReferenceDocument], uid_base : str)
     )
     metadata = {
         'title': "IQ# Magic Commands",
+        'author': rmshaffer,
```
cgranade, line 200:
```suggestion
        'author': 'rmshaffer',
```

build/docs/build_docs.py
```diff
@@ -120,8 +120,10 @@ def format_as_document(magic, uid_base : str) -> MagicReferenceDocument:
     uid = f"{uid_base}.{safe_name}"
     metadata = {
         'title': f"{magic_name} (magic command)",
+        'author': 'rmshaffer',
```
rmshaffer, line 123: This is probably fine, but I wonder if it would be better to use some more general ID here. One option is to use `microsoftopensource`. I don't know if there is any service account that is more specific to the Quantum team. @cgranade, do you happen to be aware of anything?

geduardo:
@rmshaffer AFAIK the purpose of those GitHub IDs in the metadata is for when a user fills an issue through the docs platform to tag the GitHub ID automatically to warn the person that might know how to help. For pragmatism, we tend to put the people that wrote the article or that know better about the topic discussed so they can help with the issue. 

I don't know what is the standard for automated docs tho. I've been looking to the rest of the API docs and since the documentation is generated directly from the .yml documents it works differently (I think it tags the author that made the last commit that modified the file or something like that, but sometimes it doesn't tag anyone).

Anyways, I don't have any strong preference, we can leave it with `microsoftopensource`. It will lose the purpose of it since nobody will get the notification but it will avoid future annoyances to you in case more people use this script to create documents that you don't know about.

Maybe @cgranade knows what is the standard in these cases.

rmshaffer:
Got it, so there is a functional reason for it. In that case, sure, feel free to leave my ID there!

rmshaffer:
@geduardo to test this, I made a draft PR here containing the auto-generated docs with your fix:
https://github.com/MicrosoftDocs/quantum-docs-pr/pull/963

Looks the validation succeeded without warnings, and the generated .md looks correct, so I think (if you agree) this is good to go! Thanks again for doing this!

(Note: I won't merge that PR; instead, these docs will be updated with the next release of the API docs which should happen sometime around July 20.)

PR #213, opened 2020-07-15
Replacing deprecated IconUrl with Icon and updating the assets for NuGet packages.

ricardo-espinoza:


src/Tool/Tool.csproj.orig
```diff
@@ -16,7 +16,7 @@
     <PackageReleaseNotes>See: https://docs.microsoft.com/en-us/quantum/relnotes/</PackageReleaseNotes>
```
anpaz-msft, line 16: This file shouldn't even be in the repo anymore...
Not a blocker, though.

PR #219, opened 2020-07-17
Add composite operation handling

theRoughCode:
Add handling for composite operations like `ApplyToEach`.

Basically, when we see `ApplyToEach(H, qs)`, we render `H(q[0])...H(q[N])` individually instead. The way we do this is by setting the flag `IsComposite=true` for composite operations. In this PR, we check for that flag and recursively get the execution paths of each of its nested children operations.

This PR also adds a `Mock.Standard` library that includes mocked out versions of `ApplyToEach` and `ForEach` which are required in the new tests in this PR. Both `Mock.Standard` and `Mock.Chemistry` are moved under the folder `MockLibraries`.


src/Tests/ExecutionPathTracerTests.cs
```diff
@@ -623,5 +668,33 @@ public void MResetZTest()
...
+                    {
+                        Gate = "measure",
+                        Controls = new List<Register>() { new QubitRegister(0) },
+                        Targets = new List<Register>() { new ClassicalRegister(0, 0) },
+                    },
+                    new Operation()
+                    {
+                        Gate = "measure",
+                        Controls = new List<Register>() { new QubitRegister(1) },
+                        Targets = new List<Register>() { new ClassicalRegister(0, 0) },
```
anpaz-msft, line 693: Is `ClassicalRegister(0,0)` correct? Seems it is measuring to the same classical bit, although probably I just don't know exactly what this metadata represents.

src/Tests/ExecutionPathTracerTests.cs
```diff
@@ -558,6 +558,51 @@ public void BigTest()
...
+                    Targets = new List<Register>() { new QubitRegister(1) },
+                },
+                new Operation()
+                {
+                    Gate = "H",
+                    Targets = new List<Register>() { new QubitRegister(2) },
+                },
+                new Operation()
+                {
+                    Gate = "ResetAll",
```
anpaz-msft, line 593: I wonder if `ResetAll` should be marked as composite too?

src/Tests/ExecutionPathTracerTests.cs
theRoughCode: `ClassicalRegister` has the constructor `ClassicalRegister(int qId, int cId)`. In this case, this means that the classical register is associated with the qubit with ID `0` and its classical bit ID "under qubit 0" is `0`.

src/Tests/ExecutionPathTracerTests.cs
theRoughCode: We can do that too. I can create a PR in `qsharp-runtime` to address this and make the necessary changes in a separate PR.

theRoughCode: Made a PR for that: https://github.com/microsoft/qsharp-runtime/pull/318

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -55,15 +57,36 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
         {
             this.currentDepth++;
 
+            // If `compositeTracer` is initialized, pass operations into it instead
+            if (this.compositeTracer != null)
+            {
+                this.compositeTracer.OnOperationStartHandler(operation, arguments);
+            }
```
rmshaffer, line 64: This is purely a stylistic thing, but if I'm understanding correctly, inside this `if` block seems like a logical place for an early `return`, since the intent is to just delegate to `compositeTracer` rather than running any code below.

Then the `else` below also becomes unnecessary.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -55,15 +57,36 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
...
             // Parse operations at or above specified depth
-            if (this.currentDepth <= this.renderDepth)
+            else if (this.currentDepth <= this.renderDepth)
```
rmshaffer, line 66: Similar to the above comment, you could consider something like:
```
if (this.currentDepth > this.renderDepth) return;
```
such that all of the code below can be un-indented. Again, this is stylistic - for readability, I tend to prefer short-circuit return statements rather than long indented if blocks - feel free to ignore.

src/Core/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: That's a good catch! I'll do that. :)

PR #220, opened 2020-07-21
Remove unnecessary Console.WriteLine in RegisterPackageLoadedEvent

rmshaffer:
Removing a call to `Console.WriteLine` which is causing a list of loaded assemblies to be output during the `import qsharp` statement of every Python script. The `PackageLoaded` event itself still outputs relevant logging information.


src/Kernel/IQSharpEngine.cs
```diff
@@ -101,7 +101,6 @@ private void RegisterPackageLoadedEvent(IServiceProvider services, ILogger logge
                 // Except assemblies known at compile-time as well.
                 .Add(typeof(StateVectorToHtmlResultEncoder).Assembly.GetName())
                 .Add(typeof(AzureClientErrorToHtmlEncoder).Assembly.GetName());
-            foreach (var knownAssembly in knownAssemblies) System.Console.WriteLine($"{knownAssembly.FullName}");
```
anpaz-msft, line 104: Instead of removing, could we do a `logger.Info` or `logger.Debug`?

rmshaffer: Yes, I don't think this is very useful except for debugging, so using `logger.Debug` makes sense to me.

PR #222, opened 2020-07-21
Use multiple dotnet test runs

rmshaffer:
Running `dotnet test` on the full `iqsharp.sln` results in memory allocations accumulating to more than 6 GB, which can occasionally cause the CI build agents to run out of memory.

I haven't yet found an obvious way to release the memory being allocated by each test after it finishes. My suspicion is that each test is reloading all required assemblies (such as those from Microsoft.Quantum.Standard, etc.) and the memory from the previous load of many of those assemblies is not being freed.

To prevent build failures in the meantime, this change simply breaks the test run into several separate invocations of `dotnet test`. With this change, the maximum memory usage is only around 3 GB (which occurs during the `IQSharpEngine` tests).


build/test.ps1
```diff
@@ -52,7 +62,7 @@ function Test-Python {
     }
 }
 
-Test-One '../iqsharp.sln'
+Test-One '../iqsharp.sln' @("AzureClient", "IQSharpEngine", "Workspace")
```
anpaz-msft, line 65: This feels a bit fragile. I think we're going to forget to add the new classes to the script later.
Maybe a little more work, but what if we just break the tests into multiple projects?

rmshaffer: All tests are run regardless of the list of filters here. We run `dotnet test` with each of these filters, followed by `dotnet test` for tests that didn't match any of the filters:

https://github.com/microsoft/iqsharp/blob/42b29c2054e7a552acf238a5ced71235075b3add/build/test.ps1#L15-L18

I'll add comments to the code to make it more clear what is happening.

Yes, this still has the problem that if we add _enough_ new tests, we may run into the out-of-memory issue again unless we add more filters here. But we don't run the risk of missing tests.

I've opened #224 to track fixing the underlying issue here, so that we don't forget about it.

theRoughCode:
Not sure if this might work, but looking at your hypothesis of the cause, I was wondering if running a `dotnet build` and then passing `--no-build` to `dotnet test` might help? Or is the logger output being accumulated part of the reason (I doubt that, but throwing it out there)?

rmshaffer:
> Not sure if this might work, but looking at your hypothesis of the cause, I was wondering if running a `dotnet build` and then passing `--no-build` to `dotnet test` might help? Or is the logger output being accumulated part of the reason (I doubt that, but throwing it out there)?

The invocation of `dotnet test` is already using `--no-build` here. It really seems to be a problem with the code itself. Each test constructs a new copy of the various IQ# components and loads any required packages into memory. When the next test starts, much that memory apparently is not released. Each test causes the process to allocate more and more memory, and it's not until the `dotnet test` run is complete that the process is terminated and all of the memory is released.

theRoughCode:
> > Not sure if this might work, but looking at your hypothesis of the cause, I was wondering if running a `dotnet build` and then passing `--no-build` to `dotnet test` might help? Or is the logger output being accumulated part of the reason (I doubt that, but throwing it out there)?
> 
> The invocation of `dotnet test` is already using `--no-build` here. It really seems to be a problem with the code itself. Each test constructs a new copy of the various IQ# components and loads any required packages into memory. When the next test starts, much of that memory apparently is not released. Each test causes the process to allocate more and more memory, and it's not until the `dotnet test` run is complete that the process is terminated and all of the memory is released.

Yea, in that case, I'm not too sure if there's a flag or option to force memory to be released between tests. Your solution looks like the most straightforward way to tackle this problem. 👍  

Is there a way that you're choosing which classes to pass in as arguments vs those to be run together (at the end)? I'm guessing those 3 classes are the biggest ones? 

To @anpaz-msft's point, would it make sense to automate it a bit by breaking it up by namespaces instead of passing them in as arguments and automatically retrieve the list of test namespaces (i.e. `Tests.IQSharp.{...}`)?

rmshaffer:
> Is there a way that you're choosing which classes to pass in as arguments vs those to be run together (at the end)? I'm guessing those 3 classes are the biggest ones?

Yes, it's sort of arbitrary, I just chose substrings of class names that reasonably break up the tests into large chunks.

> would it make sense to automate it a bit by breaking it up by namespaces instead of passing them in as arguments and automatically retrieve the list of test namespaces (i.e. Tests.IQSharp.{...})?

That would be a more robust way of doing it, yes. But I mean, ideally we wouldn't do any of this at all... the tests should all be able to run sequentially in one process, and the fact that the memory doesn't get released properly is likely a real bug in IQ# that should be fixed. In that sense this is just a hacky workaround that should be removed when #224 is fixed, so I wouldn't want to invest too much time into this.

build/test.ps1
```diff
@@ -9,21 +9,48 @@ $all_ok = $True
...
-        /property:Version=$Env:ASSEMBLY_VERSION
+    if ($testClassFilters) {
+        $filterArgs = $testClassFilters | foreach { "ClassName~$_" }
+        $filterArgs += "(" + (($testClassFilters | foreach { "(ClassName!~$_)" }) -join " & ") + ")"
+    }
+    else {
+        $filterArgs = @("FullyQualifiedName!=_fake_")
+    }
+    $filterArgs | foreach {
+        dotnet test $project `
```
anpaz-msft, line 40: Can you add a message here saying what project you're testing and with what filter? It will help troubleshooting build failures.

rmshaffer: That message is already here, a few lines below: 
```
Write-Host "##vso[task.logissue type=error;]Failed to test $project with filter $_"
```

PR #225, opened 2020-07-22
Wrap DisplayArgs in parentheses

theRoughCode:
There's an unhandled case where the arguments passed in to an operation is not a tuple type and is also a non-qubit type (e.g. `Foo(2)`). In this case, `DisplayArgs` would just be `2` instead of `(2)`. We have 2 options:
1. Wrap all display arguments with parentheses. This would result in most operations looking like `((2, "bar"))` where there are double parentheses since most arguments are tuples.
2. Only wrap with parentheses if the incoming argument string is not already in parentheses. This would render tuple args with single parentheses, but handle the non-tuple case by adding parentheses.

I went with the second option because I thought it would be visually less cluttered if we avoid double parentheses around tuple args. Would be open to discussion around this decision though!

This PR also includes:
- a small change to the `%trace` description
- a rename of the `depth` flag to `--depth`
- restructuring of test code
- add missing measurement tests


src/Tests/ExecutionPathTracerTests.cs
```diff
@@ -533,4 +557,71 @@ public void BigTest()
             Assert.AreEqual(expected.ToJson(), path.ToJson());
         }
     }
+
+    [TestClass]
+    public class MeasurementTests : ExecutionPathTracerTests
```
theRoughCode, line 562: This is just a reordering of existing tests from lower down in this file (no new changes included here).

src/Tests/ExecutionPathTracerTests.cs
```diff
@@ -33,7 +32,11 @@ public ExecutionPath GetExecutionPath(string name, int depth = 1)
 
             return tracer.GetExecutionPath();
         }
+    }
 
+    [TestClass]
+    public class IntrisincTests : ExecutionPathTracerTests
```
rmshaffer, line 38:
```suggestion
    public class IntrinsicTests : ExecutionPathTracerTests
```

theRoughCode: I should write a script to catch all my spelling mistakes haha

theRoughCode: Actually: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

PR #228, opened 2020-07-23
set up communication from client to kernel

Reemeela2121:
Still a WIP, but wanted to keep track of TODOs for creating a quantum state debugger as well as minor fixes needed.

Minor fixes: 
- [x] 1 canvas , 3 graphs display is buggy

TODO: 
- [x] set up communication from client to kernel correctly
- [x] create displayable state object in DebugMagic.cs
- [x] make execution stop at each new event in order to collect trace history
- [x] change data to send dumpMachine output
- [x] display output in histogram format

### DIFF
--- a/src/Kernel/Magic/DebugMagic.cs
+++ b/src/Kernel/Magic/DebugMagic.cs
@@ -218,7 +218,6 @@ namespace Microsoft.Quantum.IQSharp.Kernel
 
             channel.Display(new RawHtmlPayload
             {
-                //needs to be changed?
                 Value = $@"
                     <div id=""{divId}""></div>
                 "
--- a/src/Kernel/Magic/DebugMagic.cs
+++ b/src/Kernel/Magic/DebugMagic.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Quantum.IQSharp.Kernel
                 return true;
             }
             private Complex[]? _data = null;
-            public Complex[] GetAmplitudes()
+            public Complex[] getAmplitudes()
             {
                 var count = this.Simulator.QubitManager.GetAllocatedQubitsCount();
                 _data = new Complex[1 << ((int)count)];
--- a/src/Kernel/Magic/DebugMagic.cs
+++ b/src/Kernel/Magic/DebugMagic.cs
@@ -1,0 +1,208 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Jupyter.Core;
+using Microsoft.Jupyter.Core.Protocol;
+using Microsoft.Quantum.IQSharp;
+using Microsoft.Quantum.IQSharp.Common;
+using Microsoft.Quantum.IQSharp.Jupyter;
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
+
+namespace Microsoft.Quantum.IQSharp.Jupyter
+{
+    public class RawHtmlPayload
+    {
+        public string Value { get; set; }
+    }
+
+    public class RawHtmlEncoder : IResultEncoder
+    {
+        public string MimeType => MimeTypes.Html;
+
+        public EncodedData? Encode(object displayable) =>
+            displayable is RawHtmlPayload payload
+            ? payload.Value.ToEncodedData()
+            : (Nullable<EncodedData>)null;
+    }
+
+    public class DebugMagic : AbstractMagic
+    {
+        public DebugMagic(
+                ISymbolResolver resolver, IConfigurationSource configurationSource, IShellRouter router, IShellServer shellServer,
+                ILogger<DebugMagic> logger
+        ) : base(
+            "debug",
+            new Documentation
+            {
+                Summary = "TODO"
+            })
+        {
+            this.SymbolResolver = resolver;
+            this.ConfigurationSource = configurationSource;
+            this.shellServer = shellServer;
+            this.logger = logger;
+            router.RegisterHandler("iqsharp_debug_request", this.HandleStartMessage);
+            router.RegisterHandler("iqsharp_debug_advance", this.HandleAdvanceMessage);
+        }
+
+        private ConcurrentDictionary<Guid, ManualResetEvent> sessionAdvanceEvents
+            = new ConcurrentDictionary<Guid, ManualResetEvent>();
+        private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+        private IShellServer shellServer;
+        private ILogger<DebugMagic> logger;
+
+        /// <summary>
+        ///      The symbol resolver used by this magic command to find
+        ///      operations or functions to be simulated.
+        /// </summary>
+        public ISymbolResolver SymbolResolver { get; }
+
+        /// <summary>
+        ///     The configuration source used by this magic command to control
+        ///     simulation options (e.g.: dump formatting options).
+        /// </summary>
+        public IConfigurationSource ConfigurationSource { get; }
+
+        /// <inheritdoc />
+        public override async Task<ExecutionResult> Run(string input, IChannel channel)
+        {
+            // Run the async task in a new thread so that we don't block the
+            // server thread.
+            return await Task.Run(() => RunAsync(input, channel));
+        }
+
+        private void HandleStartMessage(Message message)
+        {
+            var content = (message.Content as UnknownContent);
+            var session = content?.Data?["debug_session"];
+            if (session == null)
+            {
+                // TODO: Communicate error back here!
+            }
+            else
+            {
+                shellServer.SendShellMessage(
+                    new Message
+                    {
+                        Header = new MessageHeader
+                        {
+                            MessageType = "iqsharp_debug_reply"
+                        },
+                        Content = new UnknownContent
+                        {
+                            Data = new Dictionary<string, object>
+                            {
+                                ["debug_session"] = session
+                            }
+                        }
+                    }
+                    .AsReplyTo(message)
+                );
+            }
+        }
+
+        private void HandleAdvanceMessage(Message message)
+        {
+            logger.LogDebug("Got debug advance message:", message);
+            var content = (message.Content as UnknownContent);
+            var session = content?.Data?["debug_session"];
+            if (session == null)
+            {
+                logger.LogWarning("Got debug advance message, but debug_session was null.", message);
+            }
+            else
+            {
+                var sessionGuid = Guid.Parse(session.ToString());
+                ManualResetEvent? @event = null;
+                lock (sessionAdvanceEvents)
+                {
+                    @event = sessionAdvanceEvents[sessionGuid];
+                }
+                @event.Set();
+            }
+        }
+
+        private async Task WaitForAdvance(Guid session)
+        {
+            await Task.Run(() =>
+            {
+                ManualResetEvent? @event = null;
+                // Find the event we need to wait on.
+                lock (sessionAdvanceEvents)
+                {
+                    @event = sessionAdvanceEvents[session];
+                }
+                @event.Reset();
+                @event.WaitOne();
+            }, cancellationTokenSource.Token);
+        }
+
+        /// <summary>
+        ///     Simulates an operation given a string with its name and a JSON
+        ///     encoding of its arguments.
+        /// </summary>
+        public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
+        {
+            var (name, args) = ParseInput(input);
+
+            var symbol = SymbolResolver.Resolve(name) as IQSharpSymbol;
+            if (symbol == null) throw new InvalidOperationException($"Invalid operation name: {name}");
+
+            var session = Guid.NewGuid();
+            lock (sessionAdvanceEvents)
+            {
+                sessionAdvanceEvents[session] = new ManualResetEvent(true);
+            }
+            using var qsim = new QuantumSimulator();
+            qsim.OnOperationStart += (callable, args) =>
+            {
+                // Tell the IOPub channel that we're starting a new operation.
+                shellServer.SendIoPubMessage(
+                    new Message
+                    {
+                        Header = new MessageHeader
+                        {
+                            MessageType = "iqsharp_debug_opstart"
+                        },
+                        Content = new UnknownContent
+                        {
+                            //change this stuff here too?
+                            Data = new Dictionary<string, object>
+                            {
+                                ["debug_session"] = session,
+                                ["callable"] = callable.FullName,
+                                ["args"] = args
+                            }
+                        }
+                    }
+                );
+                WaitForAdvance(session).Wait();
+            };
+            channel.Display(new RawHtmlPayload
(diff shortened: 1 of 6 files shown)


src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    public class RawHtmlEncoder : IResultEncoder
+    {
+        public string MimeType => MimeTypes.Html;
+
+        public EncodedData? Encode(object displayable) =>
+            displayable is RawHtmlPayload payload
+            ? payload.Value.ToEncodedData()
+            : (Nullable<EncodedData>)null;
+    }
```
rmshaffer, line 36: Just a note here, @theRoughCode has classes in the `feature/path_visualizer` branch called `DisplayableHtmlElement` and `DisplayableHtmlElementEncoder` which do basically the same thing as this. I think you'll ultimately want to use those instead of adding these new classes here.

https://github.com/microsoft/iqsharp/blob/97443e6698563a7068b899a9bf1110e65eabc4bb/src/Jupyter/Visualization/DisplayableHtmlElementEncoder.cs#L9-L41

src/Kernel/client/kernel.ts
```diff
@@ -38,29 +85,41 @@ class Kernel {
                     if (div != null) {
                         let amplitudeSquaredButton = document.createElement("button");
                         let graph = document.createElement("canvas");
+                        graph.id = "first";
```
cgranade, line 88: When two `%debug` commands are called in the same notebook, this could wind up giving the same ID to multiple HTML elements, breaking an invariant used by the `document.getElementById` DOM method. In this case, you have a workaround in that you've already got `graph` as a local variable, such that you can use it directly at line 93 instead of looking the element up in the DOM again.

src/Kernel/client/kernel.ts
```diff
@@ -38,29 +85,41 @@ class Kernel {
...
+                            document.getElementById("first").style.display = "block";
+                            document.getElementById("second").style.display = "none";
+                            document.getElementById("third").style.display = "none";
                         });
                         div.appendChild(amplitudeSquaredButton);
                         
 
                         let realImagButton = document.createElement("button");
                         let realImagGraph = document.createElement("canvas");
+                        realImagGraph.id = "second";
```
cgranade, line 102: Similarly here, and at line 115.

src/Kernel/client/kernel.ts
```diff
@@ -21,9 +21,56 @@ class Kernel {
...
         });
     }
 
+    setupMeasurementHistogramDebugger() {
+        IPython.notebook.kernel.register_iopub_handler(
+            "iqsharp_debug_opstart",
+            message => {
+                console.log("iqsharp_debug_opstart message received", message);
+
+                //check to see if there's already a button
```
theRoughCode, line 34: Is this a TODO?

src/Kernel/client/kernel.ts
```diff
@@ -21,9 +21,56 @@ class Kernel {
...
+
+                let state: DisplayableState = message.content.state;
+                let state_div = state.div_id;
+                if (state_div != null) {
+                    let div = document.getElementById(state_div);
+                    if (div != null) {
+                        //handle displaying dumpMachine stuff here
+                        if (document.getElementById("ye") == null) {
```
theRoughCode, line 42: What does `ye` mean in this case? Or is it a placeholder?

src/Kernel/client/kernel.ts
```diff
@@ -21,9 +21,56 @@ class Kernel {
...
+                            //make button here and call debugger function in onclick message
+                            let debuggerFunctionButton = document.createElement("button");
```
theRoughCode, line 44: This can be a `const`, I believe.

src/Kernel/client/kernel.ts
```diff
@@ -38,29 +85,41 @@ class Kernel {
...
                             createBarChartRealImagOption(realImagGraph, state);
                             div.appendChild(realImagGraph);
+                            document.getElementById("second").style.display = "block";
+                            document.getElementById("first").style.display = "none";
+                            document.getElementById("third").style.display = "none";
                         });
                         div.appendChild(realImagButton);
 
                         let amplitudePhaseButton = document.createElement("button");
                         let amplitudePhaseGraph = document.createElement("canvas");
```
theRoughCode, line 114: These 2 can be `const` as well.

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,224 @@
...
+            }
+            else
+            {
+                var sessionGuid = Guid.Parse(session.ToString());
+                ManualResetEvent? @event = null;
+                lock (sessionAdvanceEvents)
+                {
+                    @event = sessionAdvanceEvents[sessionGuid];
+                }
+                @event.Set();
```
theRoughCode, line 151: Would this throw an error if `@event` is `null`?

theRoughCode: Same for L156/166.

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,224 @@
...
+            {
+                ManualResetEvent? @event = null;
+                // Find the event we need to wait on.
+                lock (sessionAdvanceEvents)
+                {
+                    @event = sessionAdvanceEvents[session];
+                }
+                @event.Reset();
+                @event.WaitOne();
+            }, cancellationTokenSource.Token);
```
theRoughCode, line 167: What's the purpose of this cancellation token? I don't think I see it being used as well to cancel this `Run` method.

src/Kernel/client/plotting.ts
```diff
@@ -27,6 +27,276 @@ export interface DisplayableState {
...
+    body.appendChild(button);
+    body.appendChild(button2);
+    body.appendChild(button3);
+
+    //create event listeners
+    button.addEventListener("click", function() {
+        //step one: remove each item currently in the dataset
+        let i = 0;
+        for (i = 0; i < newCount; i++){
+            removeData(measurementHistogram);
```
theRoughCode, line 140: It seems like we're always using `removeData` in a for loop. Should we consider changing `removeData` to accept `newCount` and perform the for loop in there?

theRoughCode: Maybe I'm mistaken what `newCount` is, but should `addData` be affecting this? i.e. when we perform `addData` on L156, should `newCount` be updated so that the next click event would remove those extra `newCount`s?

src/Kernel/client/plotting.ts
```diff
@@ -27,6 +27,276 @@ export interface DisplayableState {
...
+        let y = Array.from(Array(newCount).keys()).map(idx => {
+            let bitstring = (idx >>> 0).toString(2).padStart(newCount, "0");
+            return `|${bitstring}⟩`;
+        });
+        let x = Array.from(Array(newCount).keys()).map(idx => {
+            return (amps[idx].Magnitude ** 2);
+        });
+
+        for (i = 0; i < newCount; i++){
+            addData(measurementHistogram, y[i], x[i]);
```
theRoughCode, line 156: Maybe I'm mistaken what `newCount` is, but should `addData` be affecting it? i.e. when we perform `addData`, should `newCount` be updated so that the next click event would remove those extra `newCount`s in L140, for example?

src/Kernel/Magic/DebugMagic.cs
theRoughCode: My PR has been merged in to master. If you pull in the changes, you should be able to use it: https://github.com/microsoft/iqsharp/blob/master/src/Jupyter/Visualization/DisplayableHtmlElementEncoder.cs

src/Kernel/client/kernel.ts
Reemeela2121: It's a placeholder, sorry!

theRoughCode: No worries haha! Just wanted to check in case you forget to update it :)

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,246 @@
...
+        public override ExecutionResult Run(string input, IChannel channel) =>
+            RunAsync(input, channel).Result;
+
+        private async Task HandleStartMessage(Message message) => await Task.Run(() =>
+        {
+            var content = (message.Content as UnknownContent);
+            var session = content?.Data?["debug_session"];
+            if (session == null)
+            {
+                // TODO: Communicate error back here!
```
rmshaffer, line 111: It would be good to at least log an error here, for example:
```
this.logger.LogError("Missing debug_session field in iqsharp_debug_request message.");
```
Then you could probably remove the TODO. This is a pretty unexpected case, so I don't think any further error handling would be needed.

src/Kernel/Magic/DebugMagic.cs
rmshaffer: Actually this raises a good point. It would be awesome if the `%debug` magic could integrate with Jupyter kernel interrupt, so that the debug session is cancelled when the user chooses the "interrupt kernel" command in Jupyter.

To do this, you would simply rename the `Run` method in this class to `RunCancellable`, which also contains a `CancellationToken` parameter:
https://github.com/microsoft/iqsharp/blob/03db172008c671e193c93d45c0097ed2eccc34f2/src/Jupyter/Magic/AbstractMagic.cs#L153-L163
And then you'd pass that token into `RunAsync` and eventually through to here. You'd then also want to pass that token into the `Task.Run(() => symbol.Operation.RunAsync(qsim, inputParameters))` statement later on so that the whole operation is cancelled.

This is a separate change and can certainly be done in a later PR; just wanted to make you aware of this.

For now, as @theRoughCode points out, since this is unused you could probably just remove it.

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,246 @@
...
+            }, cancellationTokenSource.Token);
+        }
+
+        /// <summary>
+        ///     Simulates an operation given a string with its name and a JSON
+        ///     encoding of its arguments.
+        /// </summary>
+        public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
+
+        {
```
rmshaffer, line 177:
```suggestion
        public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
        {
```

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -0,0 +1,246 @@
...
+                            }
+                        }
+                    }
+                );
+                WaitForAdvance(session).Wait();
+            };
+
+            channel.Display(new RawHtmlPayload
+            {
+                //needs to be changed?
```
rmshaffer, line 221: This comment could probably be removed, unless there's something left to change here?

src/Kernel/client/kernel.ts
```diff
@@ -21,56 +21,130 @@ class Kernel {
...
+                let state_div = state.div_id;
+                if (state_div != null) {
+                    let div = document.getElementById(state_div);
+                    if (div != null) {
+                        let { canvas: graph, chart: chart} = createNewCanvas(div, state);
+                        attachDumpMachineToolbar(graph, chart, state);
+                    }
+                }
+                        //make buttons that show the 3 options
+                        //real + imag, amplitude + phase, original view
```
rmshaffer, line 146: I think these comments have been taken care of above?

src/Kernel/client/kernel.ts
```diff
@@ -21,56 +21,130 @@ class Kernel {
...
+
+    }
+
+    setupMeasurementHistogramDataListener() {
+        IPython.notebook.kernel.register_iopub_handler(
+            "iqsharp_state_dump",
+            message => {
+                console.log("my message received", message);
+                let state: DisplayableState = message.content.state;
+                //this.createHistogramFromStateData(state);
```
rmshaffer, line 136: This commented line can probably be removed?

src/Kernel/client/plotting.ts
```diff
@@ -27,6 +27,272 @@ export interface DisplayableState {
...
+            },
+            ticks: {
+                maxRotation: 0,
+                minRotation: 0
+            }
+        }],
+        yAxes: [{
+            scaleLabel: {
+                display: true,
+                labelString: 'Meas. Probability'
```
rmshaffer, line 110:
```suggestion
                labelString: 'Amplitude/Phase'
```

src/Kernel/client/plotting.ts
```diff
@@ -27,6 +27,272 @@ export interface DisplayableState {
...
```
rmshaffer, line 214:
```suggestion
                labelString: 'Real/Imag'
```

src/Kernel/client/plotting.ts
```diff
@@ -27,6 +27,272 @@ export interface DisplayableState {
...
+
+    //create buttons
+    let button = document.createElement("button");
+    button.innerHTML = "amplitude^2";
+
+    let button2 = document.createElement("button");
+    button2.innerHTML = "amplitude/phase";
+
+    var button3 = document.createElement("button");
+    button3.innerHTML = "real/imag";
```
rmshaffer, line 275: Minor comment, but perhaps some more descriptive names than `button`, `button2`, and `button3` would be useful. Perhaps `buttonAmplitudeSquared`, `buttonAmplitudePhase`, and `buttonRealImag`?

src/Kernel/client/kernel.ts
```diff
@@ -39,16 +39,19 @@ class Kernel {
                     let div = document.getElementById(state_div);
                     if (div != null) {
                         //handle displaying dumpMachine stuff here
-                        if (document.getElementById("ye") == null) {
+                        if (document.getElementById("ye") != null) {
```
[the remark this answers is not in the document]
Reemeela2121: It was a placeholder but I have now learned the dangers of manually setting ids of elements in the DOM...

src/Kernel/client/kernel.ts
Reemeela2121: I commented for notetaking reasons, but I think the comment can be removed!

src/Kernel/Magic/DebugMagic.cs
Reemeela2121: I'll try to use those classes when I go back into the main branch to fix future bugs that will arise from testing

src/Kernel/Magic/DebugMagic.cs
Reemeela2121: added that line to log the error

src/Kernel/Magic/DebugMagic.cs
Reemeela2121: I will add this to my list of issues that I need to address before we merge the final feature branch to master

src/Kernel/Magic/DebugMagic.cs
Reemeela2121: To be completely honest, I'm not sure...

src/Kernel/Magic/DebugMagic.cs
```diff
@@ -35,6 +37,27 @@ public class RawHtmlEncoder : IResultEncoder
...
+
+            }
+
+            public override bool Callback(uint idx, double real, double img)
+            {
+                if (_data == null) throw new Exception("Expected data buffer to be initialized before callback, but it was null.");
+                _data[idx] = new Complex(real, img);
+                return true;
+            }
+            private Complex[]? _data = null;
```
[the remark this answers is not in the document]
Reemeela2121: fixed!

PR #229, opened 2020-07-23
Allow default depth for %trace to be set in %config

theRoughCode:
As per @cgranade's suggestion during my midterm talk, this PR allows the user to set the default depth (`1` by default) using `%config trace.defaultDepth`.

### DIFF
--- a/src/Kernel/Magic/ConfigMagic.cs
+++ b/src/Kernel/Magic/ConfigMagic.cs
@@ -67,7 +67,7 @@ namespace Microsoft.Quantum.IQSharp.Kernel
 
                     **`trace.defaultDepth`**
 
-                    **Value:** integer (default `1`)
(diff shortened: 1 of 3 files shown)


src/Kernel/Magic/ConfigMagic.cs
```diff
@@ -66,6 +64,12 @@ public class ConfigMagic : AbstractMagic
 
                     Configures the phase visualization style in output from callables such as
                     `DumpMachine` or `DumpRegister`. Supports displaying phase as arrows, numbers (in radians), both, or neither.
+
+                    **`trace.defaultDepth`**
+
+                    **Value:** integer (default `1`)
```
rmshaffer, line 70:
```suggestion
                    **Value:** positive integer (default `1`)
```

PR #230, opened 2020-07-23
Implement `IsMeasurement` flag for execution path operations

theRoughCode:
Currently, we set all measurement operations (e.g. `M`, `MResetX`) to have a fixed `Operation.Gate` as `"measure"`. However, a better approach would be to use an `IsMeasurement` flag that will be set to `true` if the given `Operation` is a measurement-type operation. This would allow future extensibility to render the measurement gate depending on what the actual gate is (e.g. render `MResetX` as measurement gate with `X` in lower-right of box).

### DIFF
--- a/src/Core/ExecutionPathTracer/ExecutionPath.cs
+++ b/src/Core/ExecutionPathTracer/ExecutionPath.cs
@@ -132,14 +132,14 @@ namespace Microsoft.Quantum.IQSharp.Core.ExecutionPathTracer
         /// <summary>
         /// True if operation is a controlled operations.
         /// </summary>
-        [JsonProperty("controlled")]
-        public bool Controlled { get; set; }
+        [JsonProperty("isControlled")]
+        public bool IsControlled { get; set; }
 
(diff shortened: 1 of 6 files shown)


src/Core/ExecutionPathTracer/ExecutionPath.cs
```diff
@@ -123,6 +123,12 @@ public class Operation
         [JsonProperty("children")]
         public IEnumerable<IEnumerable<Operation>>? Children { get; set; }
 
+        /// <summary>
+        /// True if operation is a measurement operations.
+        /// </summary>
+        [JsonProperty("isMeasurement")]
+        public bool IsMeasurement { get; set; }
```
rmshaffer, line 130: Other `bool` properties in this class don't use the `Is` prefix. I'd suggest being consistent one way or the other. I personally am in favor of `Is`-like prefixes for `bool` variables, but I'll leave it up to you. (Similar comment for all the other places where these names are used.)

Issue #171, opened 2020-06-17
Errors emitted during Python initialization when Q# code depends on additional packages


rmshaffer:
**Describe the bug**
The first run of Python programs that import and use Q# packages via `qsharp.packages.add()` will emit a bunch of compilation errors during the `import qsharp` step, even if the program is correct and it compiles and runs successfully.

**To Reproduce**
1. Download one of the `MachineLearning' samples from the [Quantum](https://github.com/microsoft/Quantum) repo, e.g., the [wine sample](https://github.com/microsoft/Quantum/tree/master/samples/machine-learning/wine).
2. With the `qsharp` Python package already installed, run `python host.py` to execute the sample.
3. Notice the many compilation errors emitted due to missing namespaces, types, and identifiers before the program actually begins to execute successfully.
```
(qsharp) PS C:\qdk\quantum\samples\machine-learning\wine> python host.py
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6104: No namespace with the name "Microsoft.Quantum.MachineLearning" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6104: No namespace with the name "Microsoft.Quantum.MachineLearning.Datasets" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6005: No type with the name "ControlledRotation" exists in any of the open namespaces.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6005: No type with the name "LabeledSample" exists in any of the open namespaces.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6005: No type with the name "SamplingSchedule" exists in any of the open namespaces.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6005: No type with the name "ControlledRotation" exists in any of the open namespaces.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS6005: No type with the name "ControlledRotation" exists in any of the open namespaces.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "WineData" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "ValidateSequentialClassifier" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
      QS5022: No identifier with the name "SequentialModel" exists.
fail: Microsoft.Quantum.IQSharp.Workspace[0]
[...]

Issue #139, opened 2020-05-07
Compiling solutions with Q# unit test does not work


vivanwin:
**Describe the bug**
When using a solution with native Q# `@test()` the solution does not compile. It throws an error saying that `Microsoft.Quantum.Xunit` is not available, I have tried to add this in the `BUILT_IN_PACKAGES` inside of the references file, however more dependencies issues pop up.

**To Reproduce**
1. Create a Q# project which uses `@test()` statements, inside of the Jupyter notebook folder. I created a sample [here](https://github.com/vivanwin/QunitTest).
2. Load the notebook.

**Expected behavior**
The solution should compile without any issue, and the test should be able to be used within other projects like the [QuantumKatas](https://github.com/microsoft/QuantumKatas).

**Screenshot**
![image](https://user-images.githubusercontent.com/25651976/81278272-bca87d00-9055-11ea-9b4e-f29b41b5aa6f.png)

**System information**
 - OS: Windows
 - Browser: Edge (chromium)
 - Version: 0.11.2004.2825

rmshaffer:
One way to work around this is to run `%package Microsoft.Quantum.Xunit` (ignore the harmless error messages that are being fixed in #203) followed by `%workspace reload`. This should compile the code successfully and allow the Q# operations to be run from inside the Q# notebook. It does not fix the compilation errors that occur on initial load, however.

One way to fix this would be for IQ# to read the package references directly from the .csproj file and to load those packages prior to compiling the Q# code in the folder.

An alternative approach to fixing this would be to implement project references for IQ# (see #24), and then have the .ipynb notebook file in a different folder from the Q# project, such that the code will only be compiled once the project and its package references have been loaded.

rmshaffer:
This should now work when using the `<IQSharpLoadAutomatically>true</IQSharpLoadAutomatically>` property in the .csproj to load the required package(s) at initialization time. See https://docs.microsoft.com/en-us/quantum/user-guide/host-programs?tabs=tabid-python#using-q-code-from-other-projects-or-packages-1 for more details.

Issue #67, opened 2019-12-02
Import Q# files _not_ located in same directory


crazy4pi314:
**Is your feature request related to a problem? Please describe.**
It seems that when importing functions and operations from a *.qs file namespace to Python, everything must be in the same directory. There also doesn't seem to be a way to change where `qsharp` is looking for those files. 

**Describe the solution you'd like**
It would be really helpful to be able to import things from other locations, so that code need not be duplicated local to each project I am working on. I would like to have a repo with common Q# code I want to reuse, that I can then in multiple other Python projects import stuff from the common Q# repo. That requires that there be a way I can tell Q# where to look for stuff to import, other than the default local directory (which is a fine default).

**Describe alternatives you've considered**
Currently, I can duplicate the Q# code everywhere, but that is really not maintainable. 

**Additional context**
n/a

cgranade:
If I understand correctly, this is related to #24?

crazy4pi314:
Yeah @cgranade this is totes related to #24.

One other use case/example where this feature would be really useful that I just ran into is the following. I have a bunch of sample code arranged in separate directories (one for each chapter of a book). I want to also have a separate directory of solutions to exercises (in Jupyter notebooks) for each chapter that reference the sample code for the chapter itself. Without referencing source files outside the current directory, this is not possible afaik.

rmshaffer:
This is now possible via the `%project` magic command. See https://docs.microsoft.com/qsharp/api/iqsharp-magic/project for details.

Issue #24, opened 2019-07-12
Project references


cgranade:
**Is your feature request related to a problem? Please describe.**
Similar to the current `%package` magic in IQ# (exposed as `qsharp.packages` in the Python client), it would be nice to have a `%project` magic that adds a reference to a `.csproj` file containing Q# code.
This would make it easier to use Visual Studio or Visual Studio Code intelliSense features with Python hosts by providing the appropriate information to the language server about Q# compilation units.
This feature would also make it easier to package Q# code along with Python hosts without having to depend on assumptions about the current working directory.

**Describe the solution you'd like**
A magic command of the form `%project ../path/to/project.csproj` would cause the following to occur:
- `project.csproj` would be added as a project reference to the current IQ# session.
- If `project.csproj` is not up to date, it would be rebuilt and its build artifact (assembly) added as a reference to all future snippets.
- Upon `%workspace reload`, referenced projects would be rebuilt if out-of-date.

glassnotes:
Are there any updates on this functionality? It would be very helpful for testing to be able to use Python/Jupyter notebooks during the development process. Thanks!

rmshaffer:
@glassnotes this is still very much on the radar, just not something we've found time to do yet.

Note that you can already use Python/Jupyter notebooks to develop with Q#. Currently all of your .qs source files must live in the same folder as the .py/.ipynb file you're working with. When loading the IQ# kernel (i.e., either when starting a Q# notebook, or running `import qsharp` from Python), the operations/functions in these .qs files are automatically made available for use. If you are actively changing the source files, you can use `%workspace reload` in Jupyter or `qsharp.reload()` in Python to force recompilation.

crazy4pi314:
So if I can add some more context to what @glassnotes was was saying we are working on developing a library for Q# here: [qRAM](https://github.com/qsharp-community/qram).

For nuget packaging purposes, we have the source for the library in a separate directory from the samples which is what @glassnotes was working on at the time. We can use the ` %package` command and adding a custom nuget feed for our GitHub published nuget package (will go to nuget.org once closer to finished) but then if you are debugging or working on the library you have to push a new packaged version to be able to use it from Python. I don't think having it in the same folder is really a solution that will work here, so in the meantime, I have written a C# host that can have the project reference. I would much rather do it in Python as we want to use it to generate the testing data/plot results but I also don't want to keep forgetting that I have to re-publish the library to use it from `%package`.

[...]

PR #233, opened 2020-07-24
Read package and project references from .csproj file

rmshaffer:
Preliminary support for package and project references in IQ#. Begins to address issues #24, #67, #139, #171.

This PR adds the following behavior during IQ# workspace load:

1. Read the `.csproj` in the current folder, if there is exactly one such file.

    - Here, "current folder" refers to the workspace folder, which by default is the folder containing the opened `.ipynb` file, but can be overridden via command line.

       > NOTE: A `--skipAutoLoadProject <true|false>` kernel command line option (and corresponding `IQSHARP_SKIP_AUTO_LOAD_PROJECT` environment variable) is also added to optionally skip reading the `.csproj` in the current folder. The new behavior should be an enhancement in most scenarios, so it is enabled by default, but there may be scenarios in which it is not desirable.

    - If we use the `.csproj` in the current folder:
    
        1. Recursively enumerate all `.csproj` project references specified by `ProjectReference` elements in the loaded `.csproj`. (To avoid circular references leading to infinite recursion, we skip any `.csproj` files that have already been opened.)

        1. Load any packages specified by `PackageReference` in any of the enumerated `.csproj` files.

        1. Enumerate all `.qs` files in folders containing the enumerated `.csproj` project references.

            - We also look for `.qs` files in subfolders, in order to match the `Sdk`'s default `QsharpCompile` source path of `**/*.qs`).

            - We catch any exceptions in enumerating files (e.g. access denied), report the error to the user, and attempt to continue on.

    - If we *do not* use `.csproj` in the current folder:
    
        1. Simply enumerate all `.qs` files in the current folder.
    
            - In this case we do *not* look for `.qs` files in subfolders, to maintain backward compatibility (that's the existing behavior) and to avoid potential issues when the `.ipynb` file may be in some location with very large and/or protected subfolders, e.g. `%USERPROFILE%`, `C:\`, etc.

1. Compile all enumerated `.qs` files if necessary (i.e., if any of the enumerated `.qs` files or `.csproj` files have changed since the last compilation).

    - All `.qs` files will be compiled using the Q# compiler that matches the current IQ# version. We ignore the QDK version specified in the `.csproj` file. This avoids potential issues that we could run into if each project were compiled using the QDK version specified in its `.csproj` file, which could be different from each other and also different from the current IQ# version.

        > NOTE: We will emit a warning if the `Microsoft.Quantum.Sdk` version specified in any of the `.csproj` files doesn't match the current IQ# version.

    - The local `.qs` files (i.e., corresponding to the `.csproj` in the current folder, or just the enumerated files if we are not using a `.csproj`) are built into a "workspace" assembly.

    - Each project obtained via `ProjectReference` will have its `.qs` files built into separate assemblies. This is necessary in order to properly support Q# features such as the `internal` access modifier. To support this, code using the "workspace" assembly throughout the IQ# codebase needs to change to allow an enumeration of assemblies.

    - The built project assemblies will be saved on disk alongside the "snippets" and "workspace" assemblies used by IQ#, and *not* inside the folders of the referenced projects. Likewise, reloading the workspace (e.g., via `%workspace reload`) will rebuild all of the project assemblies in addition to the workspace assembly.

1. Enable file system watching for `.qs` and `.csproj` in all of the above folders, rather than only the current workspace folder.

This work will be merged into `feature/project-references` feature branch. Future work in this feature branch will include adding a new `%project` magic command which will allow adding project references at runtime.


crazy4pi314:
Let me just say I am so freaking excited for this 😍

src/Core/Workspace/Workspace.cs
```diff
@@ -77,10 +88,74 @@ public string CacheFolder
...
+                        var missingFiles = projectFilesToResolve.Where(projectFile => !File.Exists(projectFile));
+                        if (missingFiles.Any())
+                        {
+                            Logger?.LogError($"Skipping invalid project references: {string.Join(";", missingFiles)}");
+                            projectFilesToResolve = projectFilesToResolve.Except(missingFiles).ToArray();
+                        }
+
+                        Logger?.LogInformation($"Adding project references: {string.Join(";", projectFilesToResolve)}");
+
+                        projectFiles = projectFiles.Concat(projectFilesToResolve);
```
cgranade, line 132: Very minor suggestion, as this is unlikely to be a performance bottleneck, but given that you're growing `projectFiles` in a loop it may be good to use a `List` instead of an array, and then replace this by `AddRange`.

src/Core/Workspace/Workspace.cs
```diff
@@ -77,10 +88,74 @@ public string CacheFolder
...
+                .Select(projectFile => Path.GetDirectoryName(projectFile))
+                .Append(Root)
+                .Distinct();
+
         /// <summary>
         /// Gets the source files to be built for this Workspace.
         /// </summary>
-        public IEnumerable<string> SourceFiles => Directory.EnumerateFiles(Root, "*.qs", SearchOption.TopDirectoryOnly);
+        public IEnumerable<string> SourceFiles =>
+            SourceFolders.SelectMany(folder => Directory.EnumerateFiles(folder, "*.qs", SearchOption.TopDirectoryOnly));
```
cgranade, line 158: I think this makes sense given that we currently have a `<QsCompile Include="**/*.qs"/>` element by default, but as a future enhancement it may be good to resolve logical projects with MSBuild to handle when that implicit compile item is modified in a particular project.

src/Tests/Workspace.ProjectReferences/Operation.qs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Tests.ProjectReferences {
+    
+    open Microsoft.Quantum.Intrinsic;
+    open Tests.ProjectReferences.ProjectA;
+
+    operation MeasureSingleQubit() : Result
+    {
```
cgranade, line 10:
```suggestion
    operation MeasureSingleQubit() : Result {
```

src/Tests/Workspace.ProjectReferences/ProjectA/ProjectA.qs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Tests.ProjectReferences.ProjectA {
+    
+    open Microsoft.Quantum.Intrinsic;
+
+    operation RotateAndMeasure(q : Qubit) : Result
+    {
```
cgranade, line 9:
```suggestion
    operation RotateAndMeasure(q : Qubit) : Result {
```

src/Core/Workspace/Workspace.cs
rmshaffer: Yes, this is a good point. Using MSBuild to resolve this will be more robust. And this also points out another inconsistency here, since I'm using `SearchOption.TopDirectoryOnly`, but the default include directive `**/*.qs` would find source files in all subdirectories.

rmshaffer:
Moving this PR back to draft status. Per offline conversation with @cgranade, in order to properly support the `internal` access modifier in Q#, we will need to compile each project reference into its own assembly. Need to rework this PR a bit in order to support that, and will republish once that's ready.

src/Core/Resolver/OperationResolver.cs
```diff
@@ -39,12 +37,8 @@ public OperationResolver(ISnippets snippets, IWorkspace workspace, IReferences r
         private IEnumerable<AssemblyInfo> RelevantAssemblies()
         {
             if (snippets?.AssemblyInfo != null) yield return snippets.AssemblyInfo;
-            if (workspace?.AssemblyInfo != null) yield return workspace.AssemblyInfo;
-
-            foreach (var asm in references.Assemblies)
-            {
-                yield return asm;
-            }
+            foreach (var asm in workspace?.Assemblies) yield return asm;
```
cgranade, line 40: Completely tangential, so please disregard, but this just makes me wish C# had an equivalent to Python's `yield from` statement.

src/Core/Snippets/Snippets.cs
```diff
@@ -115,9 +115,9 @@ private void OnWorkspaceReloaded(object sender, ReloadedEventArgs e)
             ? AssemblyInfo?.Operations
             : AssemblyInfo?.Operations
             .Concat(
-                Workspace
-                ?.AssemblyInfo
-                ?.Operations
+                Workspace?
```
cgranade, line 118: I didn't know you could split the `?.` operator up like that, neat!

src/Core/Workspace/Project.cs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    internal class ProjectFileComparer : EqualityComparer<Project>
+    {
+        public override bool Equals(Project p1, Project p2) => p1.ProjectFile == p2.ProjectFile;
+
+        public override int GetHashCode(Project project) => project.ProjectFile.GetHashCode();
+    }
+
+    public class Project
```
cgranade, line 32: I think you added this in the other PR, but in case not, should this have a public doc comment?

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -75,15 +91,38 @@ public IEntryPoint Generate(string operationName, string? executionTarget)
             if (workspaceFiles.Any())
             {
                 Logger?.LogDebug($"{workspaceFiles.Length} files found in workspace. Compiling.");
-                WorkspaceAssemblyInfo = Compiler.BuildFiles(
-                    workspaceFiles, compilerMetadata, logger, Path.Combine(Workspace.CacheFolder, "__entrypoint__workspace__.dll"), executionTarget);
-                if (WorkspaceAssemblyInfo == null || logger.HasErrors)
+
+                var workspaceAssemblies = new List<AssemblyInfo>();
+                if (Workspace is Workspace workspace)
```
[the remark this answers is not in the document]
rmshaffer: Yes, makes sense! I've added everything needed here to the `IWorkspace` interface.

src/Core/Resolver/OperationResolver.cs
rmshaffer: Yes, I was looking for that...

src/Core/Workspace/Project.cs
rmshaffer: Good point, yes, I missed this. I'll go add it in the other PR.

PR #238, opened 2020-07-29
Allow overriding builtin packages and namespaces

cgranade:
This feature helps with local development of IQ# and libraries used within IQ# by allowing for automatic loading of packages and namespaces to either be disabled, modified, or extended by the use of environment variables. Note that since on Windows, environment variables cannot be empty strings, the special string `"$null"` is recognized in both cases as disabling all built-in defaults (useful for manually setting versions of Microsoft.Quantum.Standard on a per-instance basis).

### DIFF
--- a/src/Core/References/References.cs
+++ b/src/Core/References/References.cs
@@ -68,13 +68,19 @@ namespace Microsoft.Quantum.IQSharp
             eventService?.TriggerServiceInitialized<IReferences>(this);
 
             var referencesOptions = configuration.Get<ReferencesOptions>();
-            if (!(referencesOptions?.BuiltInPackages is null))
+            if (referencesOptions?.BuiltInPackages is string pkgs)
             {
-                BUILT_IN_PACKAGES = referencesOptions
-                    .BuiltInPackages
-                    .Split(",")
-                    .Select(pkg => pkg.Trim())
-                    .ToImmutableList();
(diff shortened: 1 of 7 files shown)


src/Core/References/References.cs
```diff
@@ -41,25 +47,42 @@ public class References : IReferences
...
         ///   * Microsoft.Quantum.Standard
         /// </summary>
-        public static readonly string[] BUILT_IN_PACKAGES =
-        {
-            "Microsoft.Quantum.Standard"
-        };
+        public readonly ImmutableList<string> BUILT_IN_PACKAGES =
+            ImmutableList.Create(
+                "Microsoft.Quantum.Standard"
+            );
```
rmshaffer, line 53: Since you're modifying this anyway, is it worth updating the name of this variable to CamelCase? Also perhaps something like `AutoLoadPackages` is a more accurate name, there's not really anything "built-in" about these packages, even less so with this change...

cgranade: Good point! This should now be fixed as of 9dc4953.

rmshaffer: Looks great, thanks for updating that!

cgranade:
Looks like I forgot to adapt a test; would you be willing to take another look, @rmshaffer? Thanks!

PR #240, opened 2020-07-31
Add syntax highlighting for %trace and %lsopen commands

theRoughCode:
This PR adds syntax highlighting for the `%trace` and `%lsopen` command:

image

### DIFF
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -67,7 +67,7 @@ function defineQSharpMode() {
         {
             // built-in magic commands
             token: "builtin",
-            regex: String.raw`(%(config|estimate|lsmagic|package|performance|simulate|toffoli|version|who|workspace|trace))\b`,
+            regex: String.raw`(%(config|estimate|lsmagic|lsopen|package|performance|simulate|toffoli|trace|version|who|workspace))\b`,
             beginWord: true,
         },
         {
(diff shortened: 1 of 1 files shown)


src/Kernel/client/kernel.ts
```diff
@@ -67,7 +67,7 @@ function defineQSharpMode() {
         {
             // built-in magic commands
             token: "builtin",
-            regex: String.raw`(%(config|estimate|lsmagic|package|performance|simulate|toffoli|version|who|workspace))\b`,
+            regex: String.raw`(%(config|estimate|lsmagic|package|performance|simulate|toffoli|version|who|workspace|trace))\b`,
```
rmshaffer, line 70: Silly, but the rest of these are in alphabetical order, so:
```suggestion
            regex: String.raw`(%(config|estimate|lsmagic|lsopen|package|performance|simulate|toffoli|trace|version|who|workspace))\b`,
```
(I also snuck in one other missing one, if you don't mind including that...)

theRoughCode: Sure! I just didn't know if I should add it to this PR or another one straight into `master` haha. Thanks!

rmshaffer: Doesn't matter much; this way avoids any potential merge conflicts... 😊

PR #243, opened 2020-08-04
Log out errors and fail early in CI build

theRoughCode:
More updates to the CI build to fix Solid build errors.

### DIFF
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -87,16 +87,24 @@ function Test-JavaScript {
         Try {
             npm install
         } Catch {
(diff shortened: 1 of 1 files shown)


theRoughCode:
@rmshaffer Do you mean to remove the `$script:all_ok = $False`?

rmshaffer:
> @rmshaffer Do you mean to remove the `$script:all_ok = $False`?

@theRoughCode - yes, and perhaps also changing the error logs in the `catch` block to warnings for now, until the underlying issue is fixed?

PR #244, opened 2020-08-05
Abstract ExecutionPathTracer into its own project

theRoughCode:
As part of future efforts to use the tracer in other projects, this PR takes the first step at decoupling `ExecutionPathTracer` from `IQSharp.Core` and moving it into its own project under the `ExecutionPathTracer` folder.

### DIFF
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -87,16 +87,22 @@ function Test-JavaScript {
         Try {
             npm install
         } Catch {
-            Write-Host "##vso[task.logissue type=error;]Failed to install npm dependencies."
+            Write-Host $Error[0]
+            Write-Host "##vso[task.logissue type=warning;]Failed to install npm dependencies."
         }
     Pop-Location
 
     Write-Host "##[info]Testing JS inside $packageFolder"    
     Push-Location (Join-Path $PSScriptRoot $packageFolder)
-        if (!$options) {
-            npm test
-        } else {
-            npm test -- $options
+        Try {
+            if (!$options) {
+                npm test
+            } else {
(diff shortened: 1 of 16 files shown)


cgranade:
As a quick note, the term "tracer" has often been used informally to refer to the trace simulator.

rmshaffer:
> As a quick note, the term "tracer" has often been used informally to refer to the trace simulator.

Thanks @cgranade, I didn't realize that. Maybe then `TraceVisualizer` would be less ambiguous?

theRoughCode:
@rmshaffer @cgranade I did consider that as well, and I'm open to suggestions! I thought of `TraceVisualizer` but this folder doesn't do any of the visualization, just traces out the operations into an intermediate representation to pass to the Visualizer (JS).

rmshaffer:
> @rmshaffer Ryan Shaffer FTE @cgranade Chris Granade FTE I did consider that as well, and I'm open to suggestions! I thought of `TraceVisualizer` but this folder doesn't do any of the visualization, just traces out the operations into an intermediate representation to pass to the Visualizer (JS).

That's a good point as well. Going back to your original comment, I wonder why `ExecutionPathTracer` doesn't work? e.g. in `AzureClient` project, both the namespace and a public class within the namespace are called `AzureClient` and everything seems to work out.

theRoughCode:
@rmshaffer I'll try it out again haha

theRoughCode:
@rmshaffer ah, it might have been solved in `src\Kernel\Magic\Resolution\MagicResolver.cs` L45 with `AzureClient/AzureClient`.

PR #248, opened 2020-08-06
Implement classically controlled operations

theRoughCode:
This PR is a first step at implementing classically controlled operations and is based off of work in [this PR](https://github.com/microsoft/qsharp-runtime/pull/325).  It is part of the Q# visualization tool as described in #158.

In this PR, we:

1. Use the `IsConditional` flag in the `RuntimeMetadata` class to check if the given operation is a classically-controlled operation.
2. Implement the same logic as we do for composite operations (i.e. recursively trace nested operations). In this case we would get 2 operations (one for each conditional branch). Upon exit from the recursive tracer, we add them as `Children` of the outer condition operation.
4. Override the logic for `ApplyIfElseR` with a custom `TracerApplyIfElse`.

TODO:
- [ ] Implement getting correct classical register
- [ ] Use `IsConditional` flag in visualizer (JS) to determine if the operation is a classically-controlled operation instead of checking `children`.
- [ ] Modify classically-controlled tests in visualizer (JS) to use `IsConditional` instead of `IsControlled`.

If this logic makes sense, our next steps are to:

- [ ] Override conditional logic for all other variants of `ApplyIfElse`
- [ ] Perform an IRewriteStep to convert if/else statements to `ApplyIfElse`
- [ ] Reduce `ApplyIfOne` and `ApplyIfZero` to `ApplyIfElse`.

### DIFF
--- a/src/ExecutionPathTracer/ExecutionPath.cs
+++ b/src/ExecutionPathTracer/ExecutionPath.cs
@@ -118,9 +118,6 @@ namespace Microsoft.Quantum.IQSharp.ExecutionPathTracer
         /// <summary>
         /// Group of operations for each classical branch.
         /// </summary>
-        /// <remarks>
-        /// Currently not used as this is intended for classically-controlled operations.
-        /// </remarks>
         [JsonProperty("children")]
         public IEnumerable<IEnumerable<Operation>>? Children { get; set; }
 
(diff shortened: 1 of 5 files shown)


theRoughCode:
Closing this PR in favour of #290.

PR #250, opened 2020-08-07
Add display hook to pass diagnostics to Python notebooks

cgranade:
This PR adds a new display hook to allow passing diagnostic displays from IQ# all the way through Python notebooks as well as IQ#  notebooks, addressing an issue raised by @guenp in offline discussions about https://github.com/microsoft/QuantumLibraries/pull/302.

image

**NB:** This PR introduces changes to the API surface of the IQ# client for Python, but these changes are fully localized to private methods.

### DIFF
--- a/src/Python/qsharp-core/qsharp/clients/iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/clients/iqsharp.py
@@ -18,6 +18,7 @@ import urllib.parse
 import os
 import jupyter_client
 
+from functools import partial
 from io import StringIO
 from collections import defaultdict
 from typing import List, Dict, Callable, Any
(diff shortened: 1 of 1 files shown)


src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -157,18 +165,46 @@ def capture(msg):
...
 
     ## Internal-Use Methods ##
 
-    def _execute_magic(self, magic : str, raise_on_stderr : bool = False, **params) -> Any:
-        return self._execute(f'%{magic} {json.dumps(map_tuples(params))}', raise_on_stderr=raise_on_stderr)
+    def _execute_magic(self, magic : str, raise_on_stderr : bool = False, _quiet_ : bool = False, **kwargs) -> Any:
+        return self._execute(
+            f'%{magic} {json.dumps(map_tuples(kwargs))}',
+            raise_on_stderr=raise_on_stderr, _quiet_=_quiet_,
+            **kwargs
```
rmshaffer, line 177: Perhaps `**kwargs` shouldn't be passed through here, since it's being converted to JSON as an argument to the magic command? I noticed the CI failure and I assume it's related to this.

cgranade: Good catch, thank you! I must have made that chance accidentally renaming things to be consistent, will revert that line.

PR #257, opened 2020-08-11
Add colours to path visualizer

theRoughCode:
This PR is sets the groundwork for adding colours to the path visualizer and the possibility for customizing the colours. Specifically, this PR accomplishes the following:

- Add colours to unitary (blue) and measurement (yellow) gates
- A small tweak where the gate labels and gate outline of classically controlled gates are no longer red/blue but black
image
- Add `userStyleConfig` as an argument to `executionPathToHtml` to allow users to customize colours in visualization.
- Allow users to choose between the following 3 styles using `%config trace.style="<style>"`:

1. `Default`
<p align="center">

</p>

2.  `BlackAndWhite`
<p align="center">

</p>

3. `Inverted` (e.g. under "high contrast" setting)
<p align="center">

</p>

Note: These colours have been checked to pass the accessibility tests using [this program](https://contrastchecker.com/).

Enhancement feature for #158.

### DIFF
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -96,6 +96,13 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// </summary>
         public int TraceVisualizationDefaultDepth =>
             GetOptionOrDefault("trace.defaultDepth", 1);
+
+        /// <summary>
+        ///     Allows for setting the default visualization style for visualizing Q# operations
+        ///     using the <c>%trace</c> command.
+        /// </summary>
+        public string TraceVisualizationStyle =>
+            GetOptionOrDefault("trace.style", "Default");
     }
 
     /// <summary>
--- a/src/Jupyter/ConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource.cs
@@ -2,11 +2,9 @@
 // Licensed under the MIT License.
 #nullable enable
 
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -101,8 +99,8 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         ///     Allows for setting the default visualization style for visualizing Q# operations
         ///     using the <c>%trace</c> command.
         /// </summary>
-        public string TraceVisualizationStyle =>
-            GetOptionOrDefault("trace.style", "Default");
+        public TraceVisualizationStyle TraceVisualizationStyle =>
+            GetOptionOrDefault("trace.style", TraceVisualizationStyle.Default);
     }
 
     /// <summary>
--- a/src/Kernel/Magic/TraceMagic.cs
+++ b/src/Kernel/Magic/TraceMagic.cs
@@ -24,11 +24,11 @@ namespace Microsoft.Quantum.IQSharp.Kernel
         ///     given <see cref="ExecutionPath"/> (as a <see cref="JToken"/>), HTML div ID,
         ///     and the visualization style of the output circuit.
         /// </summary>
-        public ExecutionPathVisualizerContent(JToken executionPath, string id, string style)
+        public ExecutionPathVisualizerContent(JToken executionPath, string id, TraceVisualizationStyle style)
         {
             this.ExecutionPath = executionPath;
             this.Id = id;
-            this.Style = style;
+            this.Style = style.ToString();
         }
 
         /// <summary>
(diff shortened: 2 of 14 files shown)


src/Kernel/client/ExecutionPathVisualizer/index.ts
```diff
@@ -5,102 +5,17 @@ import { processOperations } from "./process";
...
  * Converts JSON representing an execution path of a Q# program given by the simulator and returns its HTML visualization.
  * 
- * @param json JSON received from simulator.
+ * @param json            JSON received from simulator.
+ * @param userStyleConfig Custom CSS style config for visualization.
  * 
  * @returns HTML representation of circuit.
  */
-export const executionPathToHtml = (json: ExecutionPath): string => {
+export const executionPathToHtml = (json: ExecutionPath, userStyleConfig?: StyleConfig): string => {
```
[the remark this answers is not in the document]
theRoughCode: Yea, that's a good point! Setting it through `%config` makes sense.

src/Kernel/client/ExecutionPathVisualizer/styles.ts
```diff
@@ -0,0 +1,125 @@
...
+        fill: ${config.unitary};
+    }
+    .gate-measure {
+        fill: ${config.measure};
+    }
+    <!-- Classically controlled gates -->
+    .hidden {
+        display: none;
+    }
+    .cls-control-unknown {
```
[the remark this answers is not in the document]
theRoughCode: Right, it's short for classical. The classes with the `cls` prefix are relevant to the styles of the classically-controlled interactivity.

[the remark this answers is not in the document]
theRoughCode: I can change it to `classically-controlled` for better readability?

theRoughCode:
Thanks for catching these accessibility potential cases, @rmshaffer! I'll add these possibilities as a config option :)

src/Kernel/client/ExecutionPathVisualizer/styles.ts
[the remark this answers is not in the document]
rmshaffer: Yes, I think that would be helpful!

src/Jupyter/ConfigurationSource.cs
```diff
@@ -96,6 +96,13 @@ public interface IConfigurationSource
         /// </summary>
         public int TraceVisualizationDefaultDepth =>
             GetOptionOrDefault("trace.defaultDepth", 1);
+
+        /// <summary>
+        ///     Allows for setting the default visualization style for visualizing Q# operations
+        ///     using the <c>%trace</c> command.
+        /// </summary>
+        public string TraceVisualizationStyle =>
+            GetOptionOrDefault("trace.style", "Default");
```
rmshaffer, line 105: Could you make this an enum, similar to `BasisStateLabelingConvention` and `PhaseDisplayStyle` above? A bonus is that this gives you some input validation for free.

src/Kernel/Magic/TraceMagic.cs
```diff
@@ -15,19 +15,20 @@ namespace Microsoft.Quantum.IQSharp.Kernel
...
-        ///     given <see cref="ExecutionPath"/> (as a <see cref="JToken"/>) and HTML div ID.
+        ///     given <see cref="ExecutionPath"/> (as a <see cref="JToken"/>), HTML div ID,
+        ///     and the visualization style of the output circuit.
         /// </summary>
-        public ExecutionPathVisualizerContent(JToken executionPath, string id)
+        public ExecutionPathVisualizerContent(JToken executionPath, string id, string style)
         {
             this.ExecutionPath = executionPath;
             this.Id = id;
+            this.Style = style;
```
rmshaffer, line 31: If you make `style` an enum type, it'd be reasonable to just `.ToString()` it here and keep the message content property `Style` as a string.

src/Kernel/client/ExecutionPathVisualizer/styles.ts
```diff
@@ -14,112 +22,138 @@ export interface StyleConfig {
...
+    measure: "#000000",
+    classicalUnknown: "#000000",
+    classicalZero: "#000000",
+    classicalOne: "#000000",
+};
+
+/**
+ * Set of default styles.
+ */
+export const STYLES: { [name: string]: StyleConfig } = {
```
rmshaffer, line 64: Does the all caps indicate something special here?

theRoughCode: It's just a [JavaScript convention](https://www.w3schools.com/js/js_conventions.asp) for naming constants haha

rmshaffer: Got it. I saw the other `const` things in this file weren't like that, but I guess this one is exported. Just curious, thanks!

theRoughCode: Oh yea, that's a good catch. Yea, I guess the other `const` variables are locally used whereas this is an exported one to be used "externally". Not sure if it's just my own personal style haha!

rmshaffer: My general rule of thumb is that if you are a human, personal style is allowed.

PR #258, opened 2020-08-12
Pass RuntimeCapabilities to for Azure execution targets to compiler

### LINKED ISSUE #185: Improved error messaging for Q# programs unsupported on target hardware
When using `%azure.submit` or `%azure.execute` to execute a Q# program which contains unsupported operations on the target hardware, the error message doesn't include specific details about the line number of the code causing the error. The exception being thrown contains more information which can be extracted and displayed to the user.

Also, the error is not thrown until the job translation stage. Instead, the execution target should be provided to the compiler as part of the `CompilationLoader` configuration so that the error can be discovered as early as possible.

rmshaffer:
Fixes #185 by passing `RuntimeCapabilities` for the specified Azure execution target down to the compiler.

Here's an example of the messaging with this change. Note that we get specific file and line number information from the compiler, as expected.
image

Note that it's not ideal to have the logic mapping the `RuntimeCapabilities` for each provider embedded in IQ# code, since this already exists in the compiler repo here:
https://github.com/microsoft/qsharp-compiler/blob/b77641106b8faaa9046e119dec9f704f200bce46/src/QuantumSdk/DefaultItems/DefaultItems.targets#L50-L56

However, I don't see any easy way to re-use that logic at the moment.

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -163,7 +163,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             IEntryPoint? entryPoint = null;
             try
             {
-                entryPoint = EntryPointGenerator.Generate(submissionContext.OperationName, ActiveTarget.TargetId);
+                entryPoint = EntryPointGenerator.Generate(submissionContext.OperationName, ActiveTarget.TargetId, ActiveTarget.RuntimeCapabilities);
             }
             catch (UnsupportedOperationException e)
             {
--- a/src/AzureClient/AzureExecutionTarget.cs
+++ b/src/AzureClient/AzureExecutionTarget.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System;
+using Microsoft.Quantum.QsCompiler.ReservedKeywords;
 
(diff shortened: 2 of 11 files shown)


rmshaffer:
> It looks good, are there any tests?

These code paths are covered by several tests, but I haven't added tests for the various runtime capabilities explicitly. Do you think it would make sense to actually add tests here for code snippets that should/shouldn't compile for various targets? I was thinking that might be better left to the tests in the compiler repo. But if there's something unlikely enough to change for which I can add a test here, I'd be happy to do that. (Maybe comparing measurement results, and verifying that it compiles on QPRGen1 targets but not on QPRGen0?)

samarsha:
Yeah, I guess I was thinking of tests to make sure that the runtime capabilities IQ# picks are actually going to the compiler, rather than testing that the capabilities themselves are correct (you're right that should be in the compiler tests). I think it makes sense to assume that QPRGen0 doesn't allow any result comparisons and QPRGen1 allows limited result comparisons.

rmshaffer:
Makes sense, thanks @samarsha. I've added a simple test.

PR #260, opened 2020-08-12
Ensure valid access token before each Azure service call

### LINKED ISSUE #218: Azure commands should automatically re-authenticate if necessary
**Describe the bug**
Currently, only the `%azure.connect` magic command attempts to perform authentication and retrieve a token for use with Azure Quantum API calls. If that token expires, other `%azure.*` calls will fail until another call to `%azure.connect` is made.

**To Reproduce**
1. Connect successfully to an Azure Quantum workspace using `%azure.connect`.
2. Wait until the token expires (likely >1 hour).
3. Make another API call, e.g., `%azure.jobs`.
4. Command fails. Error message doesn't indicate what exactly went wrong. It just says that no jobs were retrieved.

**Expected behavior**
Every service call should attempt to re-authenticate automatically. In most cases this will succeed silently, and the user will never need to know about it.

If re-authentication fails, then an error message should be displayed that indicates specifically that there was an authentication failure.

**Additional context**
For reference, here is the existing code snippet inside `AzureClient.ConnectAsync()` that performs the authentication:

https://github.com/microsoft/iqsharp/blob/11d393d19854282502d82bab3c4335827b73329b/src/AzureClient/AzureClient.cs#L76-L92

rmshaffer:
Fixes #218 by running the MSAL authentication code, which is triggered in this code by calling `RefreshConnectionAsync()`, before every Azure service call. This code ensures that the current access token is not expired and fetches a new one if necessary.

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -38,6 +38,16 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             ? "(no execution targets available)"
             : string.Join(", ", ValidExecutionTargets.Select(target => target.Id));
 
+        /// <summary>
+        /// Creates an <see cref="AzureClient"/> object that provides methods for
+        /// interacting with an Azure Quantum workspace.
+        /// </summary>
+        /// <param name="engine">The execution engine for interaction with Jupyter.</param>
+        /// <param name="references">The references to use when compiling Q# code.</param>
+        /// <param name="entryPointGenerator">The generator of entry points for Azure Quantum execution.</param>
+        /// <param name="metadataController">The metadata controller to use when compiling Q# code.</param>
+        /// <param name="logger">The logger to use for diagnostic information.</param>
+        /// <param name="eventService">The event service for the IQ# kernel.</param>
         public AzureClient(
             IExecutionEngine engine,
             IReferences references,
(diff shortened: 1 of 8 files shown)


src/AzureClient/AzureClient.cs
```diff
@@ -72,6 +82,37 @@ public class AzureClient : IAzureClient
...
+        {
+            var connectionResult = await ConnectToWorkspaceAsync(channel, subscriptionId, resourceGroupName, workspaceName, refreshCredentials);
+            if (connectionResult.Status != ExecuteStatus.Ok)
+            {
+                return connectionResult;
+            }
+
+            if (ActiveWorkspace == null)
+            {
+                return AzureClientError.WorkspaceNotFound.ToExecutionResult();
```
ricardo-espinoza, line 94: Is this something that we would like to log as an error?

rmshaffer: Returning an error here causes the magic command `%azure.connect` to return an error status, which will be logged and communicated back to the user.

(This particular case should never really happen, since the current implementation of `ConnectToWorkspaceAsync` should have always set `ActiveWorkspace` if it returns `ExecuteStatus.Ok`, but it seems necessary to keep this here to convince the compiler that `ActiveWorkspace` truly is not `null` when using it below.)

PR #261, opened 2020-08-12
Add MockAzureExecutionTarget for Python tests

### LINKED ISSUE #223: MockAzureWorkspace tests should load provider packages using MockNugetPackages or fixed package version
**Describe the bug**
Currently, MockAzureWorkspace tests load the default version of each of the provider packages. This doesn't work well for end-to-end CI builds, where `iqsharp` is built prior to the provider packages.

**Expected behavior**
When the AzureEnvironment is set to Mock, `ActiveTarget.PackageName` should specify a fixed version such that the tests can successfully load an existing version of these packages.

When fixing this, we should remove the `MockAzureWorkspace` check here:

https://github.com/microsoft/iqsharp/blob/c18ee78db8c9428d5210a4e788aed6b90fddcc08/src/AzureClient/AzureClient.cs#L292-L296

rmshaffer:
Fixes #223 by adding a `MockAzureExecutionTarget` class which uses a fixed version of provider packages. This allows Python tests to successfully load packages in the end-to-end QDK builds where the "current" version of these provider packages is not available.

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -263,7 +263,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         /// <inheritdoc/>
         public async Task<ExecutionResult> SetActiveTargetAsync(IChannel channel, string targetId)
         {
-            if (AvailableProviders == null)
+            if (ActiveWorkspace == null || AvailableProviders == null)
             {
                 channel.Stderr($"Please call {GetCommandDisplayName("connect")} before setting an execution target.");
                 return AzureClientError.NotConnected.ToExecutionResult();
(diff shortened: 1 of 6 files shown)


rmshaffer:
> I guess in this case the package is actually loaded, it just that it loads a specific version?
> I guess that works, except that it might not be able to actually submit jobs or validate, but I guess that's ok.

That's right, and the tests don't actually try to submit jobs or do anything with the package. Note that the C# tests are using `MockNugetPackages` as you suggested. Only the Python tests will use this (because the Python tests currently don't have as much access to the internals of the IQ# services in order to inject the `MockNugetPackages` service). It's basically just validation that the package name is generated correctly.

rmshaffer:
@anpaz-msft FYI I also validated this change on the end-to-end QDK pipeline (both the old one and the new one) before merging.

PR #272, opened 2020-08-15
Add custom metadata to gates

theRoughCode:
This PR is another set of changes working towards extending the functionality of this visualizer and generalizing the use case of the library. Towards that goal, this PR includes the following changes:

1. Rename `ExecutionPath` to `Circuit`. This visualizer does not deal with or needs to know that the input given to it is an execution path of a program. It takes an input some list of qubits and operations and then outputs their visualization. In this sense, the library is just a circuit visualizer to which we pass in an execution path to render as a circuit.
2. Allow users to pass in custom metadata to each `Operation` to be added on as a data attribute on the operation's visualization.
3. Add UUID to SVG to distinguish between multiple visualizations in the same browser.

### DIFF
--- a/src/Kernel/client/ExecutionPathVisualizer/utils.ts
+++ b/src/Kernel/client/ExecutionPathVisualizer/utils.ts
@@ -2,6 +2,18 @@ import { Metadata } from './metadata';
 import { GateType, minGateWidth, labelPadding, labelFontSize, argsFontSize } from './constants';
 
 /**
+ * Generate a UUID using `Math.random`.
+ *
+ * @returns UUID string.
+ */
+const createUUID = (): string =>
+    'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+        const r = (Math.random() * 16) | 0,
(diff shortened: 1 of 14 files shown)


src/Kernel/client/ExecutionPathVisualizer/utils.ts
```diff
@@ -1,6 +1,18 @@
...
+ * Generate a UUID using `Math.random`.
+ *
+ * @returns UUID string.
+ */
+const createUUID = (): string =>
+    'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+        const r = (Math.random() * 16) | 0,
+            v = c == 'x' ? r : (r & 0x3) | 0x8;
+        return v.toString(16);
+    });
```
rmshaffer, line 14: It may be worth adding a link to where this implementation came from ([this post](https://stackoverflow.com/questions/105034/how-to-create-guid-uuid)? 😊) and a note that it's not cryptographically secure. (Of course, it's good enough for this application!)

theRoughCode: That's a good point, thanks!

PR #274, opened 2020-08-17
Fix implementation of splitting multi-qubit unitary gates

theRoughCode:
This PR fixes how the executionPathVisualizer splits up multi-qubit unitary gates. Currently, it just checks if the difference in y-coord of the registers is greater than `registerHeight`. This fails if there is a classical register added to the upper qubit register as in the following diagram:
<p align="center">

</p>

This PR fixes it so that the first `ZZ` gate gets rendered without the dashed line, as follows:
<p align="center">

</p>

Additionally, this PR also includes the following changes:
1. Change to the API of the ExecutionPathTracer such that to use it, we do:
```JavaScript
const html: string = createExecutionPathVisualizer()
  .stylize(userStyleConfig)
  .compose(executionPath)
  .asHtml();
```
2. FIx a bug where the JS within the SVG that is used to perform interactivity (e.g. classical control toggle) does not work if we inject the circuit into the browser via `innerHtml` since the JS script isn't compiled. Now, if you pass in `true` to `asHtml` or `asSvg`, the visualizer will inject the necessary scripts into the DOM.

Note: This should be merged in after #272.


src/Kernel/client/ExecutionPathVisualizer/metadata.ts
```diff
@@ -13,6 +13,9 @@ export interface Metadata {
     controlsY: number[];
     /** Array of y coords of target registers. */
     targetsY: number[];
+    /** y coords of target registers separated by classical registers.
+     *  This is used for rendering multi-qubit unitary gates. */
+    groupedTargetsY?: number[][];
```
rmshaffer, line 18: Is `targetsY` still needed on this interface? It seems like `groupedTargetsY` should contain the same information (plus additional information about the grouping), or is there a case I'm missing?

Are you doing it this way to avoid breaking the existing consumer of this interface?

theRoughCode: Yea, that's a good question which I should have explained in the description. `groupedTargetsY` is only used for gates of type `GateType.Unitary` as they are the only gates that span multiple registers as big unitary box. Other gate types don't need to be split, so they use `targetsY`. `groupedTargetsY` is only computed for `GateType.Unitary` currently. There might be a way to merge them both 🤔

rmshaffer: I see. I am wondering if perhaps `targetsY` could be used for both scenarios. It could be a `number[]`, in which case you assume the simple scenario, or `number[][]`, in which case you treat it as `groupedTargetsY`?

This isn't really such a big deal, just nice to avoid requiring redundant information from callers if possible.

theRoughCode: That makes sense, I'll give it a try :)

src/Kernel/client/ExecutionPathVisualizer/metadata.ts
```diff
@@ -11,11 +11,8 @@ export interface Metadata {
     x: number;
     /** Array of y coords of control registers. */
     controlsY: number[];
-    /** Array of y coords of target registers. */
-    targetsY: number[];
-    /** y coords of target registers separated by classical registers.
-     *  This is used for rendering multi-qubit unitary gates. */
-    groupedTargetsY?: number[][];
+    /** Array of y coords (or groups of y coords) of target registers. */
+    targetsY: (number | number[])[];
```
rmshaffer, line 15: Nice! Is it worth documenting here when `targetsY` should be `number[]` vs. `number[][]`? I think it should be `number[][]` in the case where `type` is `GateType.ControlledUnitary`, and `number[]` in all other cases? (Partially just asking to check my own understanding... 😊)

PR #275, opened 2020-08-17
Make 'children' attribute of 'Operation' a 1D array

theRoughCode:
This change is in preparation for future work to render the whole circuit and allow zoom in/out interactivity rather than specifying a fixed render depth. To do so, we nest inner operations of the parent operation inside the `children` attribute. 

Currently, the `children` attribute is meant to only be used for classical operations (not yet implemented).  This PR changes the implementation of the `children` attribute in `Operation` to a 1D array to allow all operations to use it to represent the nested operations within them.

Note: This should be merged in after #274.

### DIFF
--- a/src/Kernel/client/ExecutionPathVisualizer/circuit.ts
+++ b/src/Kernel/client/ExecutionPathVisualizer/circuit.ts
@@ -19,6 +19,15 @@ export interface Qubit {
     numChildren?: number;
 }
 
+export enum ConditionalRender {
+    /** Always rendered. */
+    None,
+    /** Render classically-controlled operation when measurement is a zero. */
+    OnZero,
+    /** Render classically-controlled operation when measurement is a one. */
+    OnOne,
+}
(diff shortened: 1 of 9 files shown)


src/Kernel/client/ExecutionPathVisualizer/circuit.ts
```diff
@@ -19,6 +19,15 @@ export interface Qubit {
     numChildren?: number;
 }
 
+export enum ConditionalRender {
+    /** Always rendered. */
+    None,
```
rmshaffer, line 24: I understand that this means "none" in the sense of "no condition", but maybe `Always` would be a clearer name for this, since the other values `OnZero` and `OnOne` also describe when to render the operation?

theRoughCode: Yea, that's a good point. Thanks!

PR #276, opened 2020-08-17
Add telemetry datapoints for projects and namespaces

rmshaffer:
This PR adds the following datapoints for collection through the existing telemetry pipeline:

- List of opened namespaces (as part of existing `SnippetCompiled` event)
- Count of loaded projects (as part of existing `WorkspaceReloaded` event)
- New `ProjectLoaded` event, triggered when compiling a project, which contains:
   - Project identifier (based on the project file path, which will be treated as PII)
   - Count of source files
   - Count of project references
   - Count of package references
   - Elapsed time while compiling the project

### DIFF
--- a/src/Core/Snippets/ISnippets.cs
+++ b/src/Core/Snippets/ISnippets.cs
@@ -13,10 +13,11 @@ namespace Microsoft.Quantum.IQSharp
     /// </summary>
     public class SnippetCompiledEventArgs : EventArgs
     {
-        public SnippetCompiledEventArgs(string status, string[] errors, TimeSpan duration)
+        public SnippetCompiledEventArgs(string status, string[] errors, string[] namespaces, TimeSpan duration)
         {
             this.Status = status;
             this.Errors = errors;
+            this.Namespaces = namespaces;
             this.Duration = duration;
         }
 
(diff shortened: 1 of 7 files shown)


src/Core/Snippets/ISnippets.cs
```diff
@@ -13,10 +13,11 @@ namespace Microsoft.Quantum.IQSharp
     /// </summary>
     public class SnippetCompiledEventArgs : EventArgs
```
cgranade, line 14: Not directly relevant, but goodness, I can't wait for C# 9 record classes.

src/Core/Workspace/Project.cs
```diff
@@ -37,6 +37,46 @@ internal class ProjectFileComparer : EqualityComparer<Project>
             project.ProjectFile.GetHashCode(StringComparison);
     }
 
+    /// <summary>
+    /// List of arguments that are part of the project loaded event.
+    /// </summary>
+    public class ProjectLoadedEventArgs : EventArgs
+    {
+        public ProjectLoadedEventArgs(Uri? projectUri, int sourceFileCount, int projectReferenceCount, int packageReferenceCount, TimeSpan duration)
```
cgranade, line 45: Is there a way to track if a project load happened explicitly, or as the result of a transitive dependency from another load event?

rmshaffer: Yes, that's a good idea, I'll add a `UserAdded` field here, which will track whether the project was added directly via `%project` or `qsharp.projects.add()`.

PR #279, opened 2020-08-18
Properly handle npm install or test errors in CI

rmshaffer:
Instead of catching and ignoring all errors in `npm install` and `npm test`, this PR properly redirects the errors to stdout in the PowerShell script and allows the code to continue, only reporting an error if the installation or tests did indeed fail (rather than just something being output to stderr).

This fix uses the same workaround as is used here: 
https://github.com/microsoft/iqsharp/blob/cb821411572f266c0d4c9f842cd014e6beb7def8/build/pack-conda.ps1#L31-L47

Using this PR to test impact on CI. I also have an end-to-end QDK build running to ensure that the CI passes there also.

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -36,6 +36,7 @@ Push-Location (Join-Path $PSScriptRoot ../src/Kernel)
         $ErrorActionPreference = 'Continue'
         npm install 2>&1 | ForEach-Object { "$_" }
     } Catch {
+        Write-Host $Error[0]
         Write-Host "##vso[task.logissue type=error;]Failed to install npm dependencies."
         $script:all_ok = $False
     } Finally {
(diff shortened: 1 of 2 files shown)


build/build.ps1
```diff
@@ -29,9 +29,23 @@ function Build-One {
...
     Try {
-        npm install
+        # npm install writes some output to stderr, which causes PowerShell to throw
+        # unless $ErrorActionPreference is set to 'Continue'.
+        # We also redirect stderr output to stdout to prevent an exception being incorrectly thrown.
+        $OldPreference = $ErrorActionPreference;
+        $ErrorActionPreference = 'Continue'
+        npm install 2>&1 | ForEach-Object { "$_" }
     } Catch {
+        Write-Host $Error[0]
```
theRoughCode, line 39: Would the catch block ever run in this case since we're redirecting all output to stderr?

rmshaffer: Probably not? I don't know if there's another way for an exception to be thrown here. I thought it wouldn't hurt to keep it there.

theRoughCode: I guess there's no good way to separate warnings and errors :P

PR #280, opened 2020-08-18
ExecutionPathTracer: Render whole circuit + zoom in/out

### LINKED ISSUE #271: No image produced for %trace magic
**Describe the bug**
If an operation that generates a valid circuit image with the `%trace` command is wrapped with another operation, the output image is empty.

**To Reproduce**
1. Define the following operations in your Q# notebook: 
```
operation ApplyH () : Unit {
    using (q = Qubit()) {
        H(q);
        H(q);
    }
}

operation ApplyHWrapper() : Unit {
    return ApplyH();
}
```
2. Run the trace command for both of the operations defined above:
```
%trace ApplyH
```
and
```
%trace ApplyHWrapper
```
**Expected behavior**
I would expect the traced image output of `ApplyHWrapper` to be the same as `ApplyH`.

**Screenshots**
image

**System information**
-Windows 10, EdgeDev, IQSharp version in screenshot

theRoughCode:
This PR introduces a (minor?) overhaul with how the `ExecutionPathTracer` works. Currently, the `ExecutionPathTracer` renders operations at or above a given depth. There was little flexibility in render operations at difference depths. This PR rethinks that implementation to allow the more flexibility with features such as zooming in/out on individual gates.

As a bonus, this fixes #271 .

Specifically, this PR introduces the following changes:

- Overhauls (somewhat complicated) logic of determining which operations to trace and instead traces all operations and creates a DAG (tree) structure of operations.
- Adds child target registers to parent's targets so that parent operations will render on the appropriate targets (this is what fixes #271).
- Implements zoom in/out feature (click to zoom in, ctrl-click to zoom out) with visual indicators (i.e. cursor changes to zoom-in on hover) to signal to users that a gate can be zoomed in.

Also, this PR is intended to be compatible with our current v1 `ExecutionPathVisualizer` and the ongoing work with v2, aka [Sqore](https://dev.azure.com/ms-hiddenfalls-interns-20/_git/KohRaphael2020). The specs with `%trace` and the `--depth` flags are intended to work as before, so no changes to samples/docs are needed. 

Note: This PR is needed by the State Visualizer sample integration.


src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -97,33 +65,26 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
...
+            Operation? parentOp = this.operations.Peek();
+            if (parentOp == null) return;
+
+            // CNOTs are Controlled X under the hood, so we don't need to render the nested CNOT
+            if ((currentOperation.Gate == "X" && currentOperation.IsControlled) &&
+                (parentOp.Gate == "X" && parentOp.IsControlled)) return;
+
+            // Add operation to parent operation's children
+            if (parentOp.Children == null) parentOp.Children = new List<Operation>() { currentOperation };
+            else parentOp.Children = parentOp.Children.Append(currentOperation);
```
rmshaffer, line 82: [nit] May be slightly more readable this way:
```suggestion
            parentOp.Children = parentOp.Children?.Append(currentOperation) ?? new List<Operation>() { currentOperation };
```

src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -97,33 +65,26 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
...
+
+            // Add target qubits to parent
+            var parentTargets = parentOp.Targets.ToList();
+            parentTargets.AddRange(currentOperation.Targets.Where(reg => reg is QubitRegister));
+            parentOp.Targets = parentTargets.Distinct();
```
rmshaffer, line 87: [nit] Another small readability suggestion (please double-check that I haven't changed the behavior here!):
```suggestion
            parentOp.Targets = parentOp.Targets
                .Concat(currentOperation.Targets.Where(reg => reg is QubitRegister))
                .Distinct();
```

src/Kernel/client/visualizer.ts
```diff
@@ -0,0 +1,138 @@
...
+        circuit.operations = this.selectOpsAtDepth(circuit.operations, renderDepth);
+        this.renderCircuit(circuit);
+    }
+
+    // Depth-first traversal to assign unique ID to operation.
+    // The operation is assigned the id `id` and its `i`th child is recursively given
+    // the id `${id}-${i}`.
+    fillGateRegistry(operation: Operation, id: string) {
+        if (operation.dataAttributes == null) operation.dataAttributes = {};
+        operation.dataAttributes["id"] = id;
```
rmshaffer, line 57: [nit] Just for consistency...
```suggestion
        operation.dataAttributes['id'] = id;
```

src/ExecutionPathTracer/ExecutionPathTracer.cs
rmshaffer: or, another option:
```suggestion
            parentOp.Children = (parentOp.Children ?? new List<Operation>()).Append(currentOperation);
```

theRoughCode: @rmshaffer thanks for these suggestions! I love these tips on how to make C# code more compact and neater, so keep em coming 😉

src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -55,39 +46,16 @@ public class ExecutionPathTracer
         /// </summary>
         public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
         {
-            this.currentDepth++;
+            RuntimeMetadata? metadata = null;
 
-            // If `compositeTracer` is initialized, pass operations into it instead
-            if (this.compositeTracer != null)
+            var parentOp = (this.operations.Count > 0) ? this.operations.Peek() : null;
```
cgranade, line 51: You could make this a bit more compact and readable by directly using the condition that you're testing here:

```C#
var metadata = (this.operations.Count > 0) && !this.operations.Peek().IsMeasurement
    ? operation.GetRuntimeMetadata(arguments)
    : null;
```
That works because `&&` short-circuits, such that if `this.operations.Count == 0`, the `?:` already falls through.

src/ExecutionPathTracer/ExecutionPathTracer.cs
cgranade: There's a subtle issue that may come up here in that `Enumerable.Append(this IEnumerable<TSource>, TSource)` returns an `IEnumerable<TSource>` rather than a collection; I'd suggest using `.ToList` to explicitly save that new enumerable as a list. Even better, you've effectively treated `parentOp.Children` as an _immutable_ list here, since you make a new list from it and save that back to the mutable variable without modifying the list itself. That speaks to using `System.Collections.Immutable.ImmutableList` and its `.Add` method instead.

src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -97,33 +65,26 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
...
+            var currentOperation = this.operations.Pop();
+            if (currentOperation == null) return;
 
-            // Add latest parsed operation to list of operations, if not null
-            if (this.currentOperation != null)
-            {
-                this.operations.Add(this.currentOperation);
-                this.currentOperation = null;
-            }
+            Operation? parentOp = this.operations.Peek();
```
cgranade, line 73: As noted above, it would help to clarify under what conditions the element popped from the stack is expected to be null.

src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -97,33 +65,26 @@ public void OnOperationStartHandler(ICallable operation, IApplyData arguments)
...
-                    this.currCompositeOp = null;
-                    this.compositeTracer = null;
-                    return;
-                }
-
-                // Pass operations down to it for handling
-                this.compositeTracer.OnOperationEndHandler(operation, result);
-                return;
-            }
+            var currentOperation = this.operations.Pop();
```
cgranade, line 70: This pattern happens enough in your PR, it might help to make it an internal extension method:

```
internal bool TryPop<TSource>(this Stack<TSource> source, out TSource? element)
{
    if (source.Count == 0)
    {
       element = null;
       return false;
    }
    else
    {
        element = source.Pop();
        return true;
    }
}
```
That would allow for replacing this line with something similar to:
```C#
if (!this.operations.TryPop(out var currentOperation) || currentOperation == null) return;
```

src/Kernel/client/ExecutionPathVisualizer/styles.ts
```diff
@@ -109,6 +109,9 @@ export const style = (customStyle: StyleConfig = {}): string => {
     .oplus {
         fill: ${styleConfig.oplus};
     }
+    [data-zoom-in="true"] {
```
cgranade, line 112: Out of curiosity, what motivates the use of data- attributes over CSS classes for marking zoomable elements?

src/Kernel/client/visualizer.ts
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/Kernel/client/ExecutionPathVisualizer/styles.ts
theRoughCode: No particular reason other than that the current implementation of the visualizer allows configurability through data attributes. So, instead of also allowing to pass in extra classes to add to the element, I chose to just use the already exposed configurability. Classes might've worked a little better in this case since it was just a boolean value but I felt that it was similar enough that it didn't warrant that extra change to the API of the visualizer. Open for suggestions though!

src/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Thanks @cgranade! That's a really neat way of writing it :D

src/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: The `TSource?` gave the error: `A nullable type parameter must be known to be a value type or non-nullable reference type. Consider adding a 'class', 'struct', or type constraint.`

Would it also make sense if  I use `Operation` instead of `TSource` specifically since we know that in this namespace we'll only be using a `Stack<Operation>`?

theRoughCode: Wait, it's an actual method? https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1.trypeek?view=netcore-3.1

cgranade: Oh, goodness, I should have known that... sorry about the confusion there. Anyway, for that error, `T?` doesn't make sense for an open `T` without a `where` clause to tell it if `T?` means a nullable reference type or shorthand for `Nullable<T>`. This is the danger of writing code without a compiler, sorry.

src/Kernel/client/ExecutionPathVisualizer/styles.ts
cgranade: Either way definitely works, just was a little surprising to see an attribute selector instead of a class selector, since the latter are generally a bit easier to work with. That said, I figured there was likely a good reason such that I didn't want to flag the use of attribute selectors as a problem per se, so much as to understand. Thanks for the explanation!

src/ExecutionPathTracer/ExecutionPathTracer.cs
theRoughCode: Gotcha, and no worries! Thanks Chris!

PR #283, opened 2020-08-19
Updating Notice.txt with a more recent list of third party libraries.

ricardo-espinoza:
Through a recent component governance audit we're extending the list of third party libraries used by components in this repository.


ricardo-espinoza:
> Makes sense to me, thanks for doing this! I notice there are a lot of licenses for Microsoft products included in the list -- is that expected? Seems fine, just curious.

Thanks for the review, Ryan. Yes, this is the expectation currently; however, I'll follow up on this to confirm as we automate this process in the future. This is an auto-generated file, produced by a tool.

Issue #277, opened 2020-08-17
Use MSBuild logic to compile .csproj references


rmshaffer:
**Is your feature request related to a problem? Please describe.**
As of https://github.com/microsoft/iqsharp/pull/262, IQ# supports adding references to `.csproj` project files that reference `Microsoft.Quantum.Sdk`. However, the implementation of this feature simply parses the `.csproj` for `ProjectReference` and `PackageReference` elements and does not actually use MSBuild logic to build the project. This means that any custom build logic (for example, custom `QsharpCompile` specifications) would not be honored, and therefore the project may not compile as intended.

**Describe the solution you'd like**
IQ# should use MSBuild logic so that project assemblies are built identically to the way `dotnet build` would build them.

**Describe alternatives you've considered**
One alternative would be for IQ# to simply call `dotnet build` out-of-process and load the resulting assembly. But since IQ# is delivered via a conda package, this would require the .NET CLI to also be included as a dependency of that conda package. The .NET CLI is not currently available as a conda package, so that may result in a bit of additional work. It seems preferable to use the MSBuild logic directly inside the IQ# process.

rmshaffer:
Wanted to add here, from discussion in #285, that it would also be desirable to support arbitrary project references (e.g. C# projects), not only projects based on `Microsoft.Quantum.Sdk`. For example, custom magic commands or display encoders can currently only be loaded as package references, but in the context of a large project it may certainly be desirable to load those via project references as well.

PR #285, opened 2020-08-19
Add support for IQSharpLoadAutomatically property

rmshaffer:
This PR disables the new `.csproj` auto-load feature by default. Instead, we will look for an `<IQSharpLoadAutomatically>` property in the `.csproj`, and only load the `.csproj` during workspace reload if that property is set to `true`.

Logging is also included to emit warnings that the default behavior may change in the future.

### DIFF
--- a/src/Core/Workspace/Project.cs
+++ b/src/Core/Workspace/Project.cs
@@ -208,7 +208,7 @@ namespace Microsoft.Quantum.IQSharp
             var defaultProject = new Project(string.Empty, rootFolder, cacheFolder);
             if (skipAutoLoadProject)
             {
-                logger?.LogInformation("Not looking for .csproj to load automatically, because skipAutoLoadProject was specified " +
+                logger?.LogDebug("Not looking for .csproj to load automatically, because skipAutoLoadProject was specified " +
                     "via --skipAutoLoadProject command line switch or via IQSHARP_SKIP_AUTO_LOAD_PROJECT environment variable.");
                 return defaultProject;
             }
@@ -216,7 +216,7 @@ namespace Microsoft.Quantum.IQSharp
             var projectFiles = Directory.EnumerateFiles(rootFolder, "*.csproj", SearchOption.TopDirectoryOnly);
             if (!projectFiles.Any())
             {
-                logger?.LogInformation($"No .csproj file found in workspace root {rootFolder}. Will compile only .qs files in this folder.");
+                logger?.LogDebug($"No .csproj file found in workspace root {rootFolder}. Will compile only .qs files in this folder.");
                 return defaultProject;
             }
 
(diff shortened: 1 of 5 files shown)


src/Core/Workspace/Project.cs
```diff
@@ -184,13 +201,66 @@ public class Project
...
         /// <param name="skipAutoLoadProject">Whether to skip auto-loading the project file.</param>
+        /// <param name="logger">Used for reporting log messages.</param>
         /// <returns>The created <see cref="Project"/> object.</returns>
-        internal static Project FromWorkspaceFolder(string rootFolder, string cacheFolder, bool skipAutoLoadProject)
+        internal static Project FromWorkspaceFolder(string rootFolder, string cacheFolder, bool skipAutoLoadProject, ILogger? logger = null)
         {
+            var defaultProject = new Project(string.Empty, rootFolder, cacheFolder);
+            if (skipAutoLoadProject)
+            {
+                logger?.LogInformation("Not looking for .csproj to load automatically, because skipAutoLoadProject was specified " +
```
anpaz-msft, line 211: Most of these seem more like debug than info messages.

src/Core/Workspace/Project.cs
```diff
@@ -184,13 +201,66 @@ public class Project
...
+                logger?.LogWarning($"Multiple .csproj files referencing Microsoft.Quantum.Sdk found in workspace root {rootFolder}. " +
+                    "Project auto-load is only supported when a single Microsoft.Quantum.Sdk project is present. " +
+                    "Will compile only .qs files in this folder.");
+                return defaultProject;
+            }
+
+            var project = projects.Single();
+            if (!project.ShouldAutoLoad.HasValue)
+            {
+                logger?.LogWarning($"Future deprecation warning: Found .csproj referencing Microsoft.Quantum.Sdk at {project.ProjectFile}, " +
```
anpaz-msft, line 244: This is a nice warning 👍

src/Core/Workspace/Project.cs
```diff
@@ -143,6 +144,22 @@ public class Project
         internal bool UsesQuantumSdk =>
             Sdk.StartsWith("Microsoft.Quantum.Sdk");
 
+        private const string AutoLoadPropertyName = "IQSharpLoadAutomatically";
+        private bool? ShouldAutoLoad
+        {
+            get 
+            {
+                if (string.IsNullOrEmpty(ProjectFile)) return false;
```
anpaz-msft, line 152: This, of course, expects that the `IQSharpLoadAutomatically` is defined in this file. MSBuild, though, supports importing props files and it's common practice to set values there. It would also be nice to be able to enable this property via include a nuget package.
Just highlighting that our current support for csproj files is very limited. Good enough for a v1 version but I truly hope we keep working on it.

src/Core/Workspace/Project.cs
```diff
@@ -184,13 +201,66 @@ public class Project
...
+                .Select(projectFile => Project.FromProjectFile(projectFile, cacheFolder))
+                .Where(project => project.UsesQuantumSdk);
+            if (!projects.Any())
+            {
+                logger?.LogInformation($"No .csproj file referencing Microsoft.Quantum.Sdk found in workspace root {rootFolder}. " +
+                    "Will compile only .qs files in this folder.");
+                return defaultProject;
+            }
+
+            if (projects.Count() != 1)
```
anpaz-msft, line 233: If we're actually supporting multiple projects. Why enable this when there is only one?

src/Tests/WorkspaceTests.cs
```diff
@@ -80,6 +80,26 @@ public void ProjectReferencesWorkspace()
...
+        {
+            // Loading this workspace should fail because the .csproj does not specify <IQSharpLoadAutomatically>,
+            // which prevents the code that depends on Workspace.ProjectReferences.ProjectB from compiling correctly.
+            var ws = Startup.Create<Workspace>("Workspace.ProjectReferences.ProjectA");
+            ws.Reload();
+            Assert.IsTrue(ws.HasErrors);
+
+            // Loading this workspace should succeed, and its Q# operations should be available, but the .csproj
+            // reference should not be loaded because the .csproj does not specify <IQSharpLoadAutomatically>.
+            ws = Startup.Create<Workspace>("Workspace.ProjectReferences.ProjectB");
```
anpaz-msft, line 94: It would be nice to add a test for a project that explicitly disable iQ# loading (i.e. has `IQSharpLoadAutomatically` set to `false`)

src/Core/Workspace/Project.cs
rmshaffer: I totally agree. The real solution is https://github.com/microsoft/iqsharp/issues/277, to use MSBuild to understand the project semantics rather than parsing the file ourselves. Your earlier suggestion of using `dotnet build`, rather than encoding the build logic ourselves, seems to me like one of the best candidates, since it gets us support for all types of projects including Q# and C# projects (e.g., this would also give us support for projects that define custom magic commands or display encoders).

src/Core/Workspace/Project.cs
rmshaffer: Again it's a limitation of the way we treat `.csproj` files at the moment. We assume that any project compiles source files with scope `**\*.qs`, so two in the same folder would conflict.

But, this is a good point, I think this check should be moved _below_ the check for the auto-load property. (It's fine if multiple projects are in a folder, as long as we only auto-load one of them.)

Issue #255, opened 2020-07-23
iqsharp-base Docker image tag should use new SemVer-style version


vivanwin:
### Problem
When you run the updateQDKVersion.ps1 script, the update for the Dockerfile gives a wrong version format for the container repository. 
The new version code for the rest of the repo is in this format `mcr.microsoft.com/quantum/iqsharp-base:x.xx.xxxxxxxx` however the container register has this format `mcr.microsoft.com/quantum/iqsharp-base:x.xx.xxxx.xxxx`. 

### Solution
The update script should be updated to make sure the right format is used in the Docker file.

tcNickolas:
(Transferring the issue to the iqsharp repo, since the fix will involve changing the version of the entry to match the versions in the rest of places rather than modifying the update script to accommodate the version difference.)

rmshaffer:
The iqsharp-base Docker image tag is being set here:
https://github.com/microsoft/iqsharp/blob/10726ae2af98942dc7104a745fb6105046088766/build/pack.ps1#L89

rmshaffer:
This work was completed as of the 0.12.20082513 QDK release. The Docker image is now published in the format `mcr.microsoft.com/quantum/iqsharp-base:x.xx.xxxxxxxx`.

PR #287, opened 2020-08-19
WIP: Use semantic version as identifier of Docker IQSharp images instead of build number

ricardo-espinoza:
This change intends to add consistency in the version numbers used for different components of the QDK.

Using the semantic version is better aligned with the names of the SDK, the NuGet packages and the VS Code extension.

### DIFF
--- a/build/pack-docs.ps1
+++ b/build/pack-docs.ps1
@@ -30,7 +30,7 @@ if ("$Env:BUILD_RELEASETYPE" -eq "release") {
 # easily inject the right base image into the FROM line. In doing so,
 # the build context should include the build_docs.py script that we need.
 $dockerfile = @"
-FROM ${Env:DOCKER_PREFIX}iqsharp-base:${Env:BUILD_BUILDNUMBER}
+FROM ${Env:DOCKER_PREFIX}iqsharp-base:${Env:SEMVER_VERSION}
 
 USER root
 RUN pip install click ruamel.yaml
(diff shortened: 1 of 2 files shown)


rmshaffer:
I just realized looking at the CI result that the CI in this repo doesn't actually use the semantic version anywhere. I think perhaps we're fine leaving this as `BUILD_BUILDNUMBER`, since that's consistent with the assemblies that get published from this repo. And the issue https://github.com/microsoft/iqsharp/issues/255 can be closed as a result of updating the official QDK pipeline to publish the images using the semantic version.

ricardo-espinoza:
We're doing this instead in the official QDK pipeline instead.

ricardo-espinoza:
I'm reopening the pull request, but this time changing the version identifier for the docs only.

build/pack-docs.ps1
```diff
@@ -30,7 +30,7 @@ if ("$Env:BUILD_RELEASETYPE" -eq "release") {
 # easily inject the right base image into the FROM line. In doing so,
 # the build context should include the build_docs.py script that we need.
 $dockerfile = @"
-FROM ${Env:DOCKER_PREFIX}iqsharp-base:${Env:BUILD_BUILDNUMBER}
+FROM ${Env:DOCKER_PREFIX}iqsharp-base:${Env:SEMVER_VERSION}
```
rmshaffer, line 33: I think the problem you'll run into here is that `Env:SEMVER_VERSION` is not defined in `set-env.ps1`; that's a variable that only exists in the end-to-end QDK pipeline.

Probably instead you'd want to make the image version a parameter to `pack-docs.ps1`. Then the CI build here can pass `Env:BUILD_BUILDNUMBER`, and the QDK pipeline can pass the SemVer version.

ricardo-espinoza: Makes sense. I'll start that approach as a separate PR.

PR #289, opened 2020-08-20
Specify version of iqsharp docs as a parameter

ricardo-espinoza:
The pack-docs script is shared by the CI build of this repository as well as the QDK internal build.

However, we're using different version formats in those builds. With this change, the version string is specified as a parameter so each build system can provide the expected version string.

### DIFF
--- a/build/pack-docs.ps1
+++ b/build/pack-docs.ps1
@@ -8,7 +8,7 @@
 #>
 
 param(
-    [string] $Version
+    [string] $DockerImageTag
 );
 
 $ErrorActionPreference = 'Stop';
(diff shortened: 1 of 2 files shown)


build/pack-docs.ps1
```diff
@@ -8,7 +8,7 @@
 #>
 
 param(
-
+    [string] $Version
```
cgranade, line 11: At this point, it's not even a version number, but a Docker image tag. Renaming the parameter accordingly to make it clear that it is directly used as an image tag may help alleviate future confusion.

cgranade (dismissed):

build/pack-docs.ps1
rmshaffer: Agreed, perhaps something like `TagSuffix`?

rmshaffer (dismissed):

build/pack-docs.ps1
ricardo-espinoza: Thanks @cgranade . I just did, and the change did a reset on your vote. :)

PR #304, opened 2020-09-02
Fix all C# compilation warnings

rmshaffer:
This PR ensures that `iqsharp.sln` builds without warnings.

### DIFF
--- a/src/AzureClient/Visualization/HistogramEncoders.cs
+++ b/src/AzureClient/Visualization/HistogramEncoders.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.Extensions.Logging;
 using Microsoft.Jupyter.Core;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -19,7 +20,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
 
     internal static class HistogramExtensions
     {
-        internal static Histogram ToHistogram(this Stream stream)
+        internal static Histogram ToHistogram(this Stream stream, ILogger? logger = null)
         {
             var output = new StreamReader(stream).ReadToEnd();
             var deserializedOutput = JsonConvert.DeserializeObject<Dictionary<string, JToken>>(output);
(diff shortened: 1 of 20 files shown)


src/AzureClient/Visualization/HistogramEncoders.cs
```diff
@@ -23,16 +23,17 @@ internal static Histogram ToHistogram(this Stream stream)
...
-            var deserializedHistogram = deserializedOutput["Histogram"] as JArray;
 
             var histogram = new Histogram();
-            for (var i = 0; i < deserializedHistogram.Count - 1; i += 2)
+            if (deserializedOutput["Histogram"] is JArray deserializedHistogram)
             {
-                var key = deserializedHistogram[i].ToObject<string>();
-                var value = deserializedHistogram[i + 1].ToObject<double>();
-                histogram[key] = value;
+                for (var i = 0; i < deserializedHistogram.Count - 1; i += 2)
```
ricardo-espinoza, line 30: Should we assert or validate that the de-serialized histogram has an even number of entries representing key/value pairs, or is that a given that doesn't need validation?

src/AzureClient/Visualization/TargetStatusEncoders.cs
```diff
@@ -21,7 +21,7 @@ internal static class TargetStatusExtensions
             {
                 ["id"] = target.Id,
                 ["current_availability"] = target.CurrentAvailability,
```
ricardo-espinoza, line 23: Does this one require a default value as done in line 24 below?

src/Tests/AzureClientMagicTests.cs
```diff
@@ -268,6 +268,7 @@ public class MockAzureClient : IAzureClient
         internal List<string> SubmittedJobs = new List<string>();
         internal List<string> ExecutedJobs = new List<string>();
 
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
ricardo-espinoza, line 271: Would it make sense to wrap this method in a sync method with a wait to avoid changing compiler pragmas?

ricardo-espinoza (approved):

src/Tests/AzureClientMagicTests.cs
rmshaffer: Since this is test code, I think disabling this warning is acceptable to improve readability of the code. There are several other test files in this project that already disable this particular warning; we have many interfaces with async methods, and the tests often have stubs of these that aren't actually asynchronous. (I definitely agree with you that it'd be better not to do this in product code.)

src/AzureClient/Visualization/TargetStatusEncoders.cs
rmshaffer: Yes, that's a good catch, we should be consistent here.

src/AzureClient/Visualization/HistogramEncoders.cs
rmshaffer: Yes, that's a good point. We should at least log a warning so that we detect this case, if it ever happens.

ricardo-espinoza (approved):

Issue #310, opened 2020-09-08
Frequent pytest failures in CI build


rmshaffer:
**Describe the bug**
CI builds have started failing more and more frequently with a pytest failure:
```
##[error]Failed to test Python inside ../src/Python/qsharp-core/qsharp/tests
```

The failure occurs due to an `abort` response status from the IQ# kernel while processing the `%azure.connect` magic command, e.g.:

`DEBUG    qsharp.clients.iqsharp:iqsharp.py:213 sending:`
`%azure.connect {"storage": "test", "subscription": "test", "resourceGroup": "test", "workspace": "test"}`

`DEBUG    qsharp.clients.iqsharp:iqsharp.py:256 received:`
`{'header': {'msg_id': 'c15aee67-cbce-4700-a9b1-6a3acf8a87c3', 'username': 'vsts', 'session': '13b75548-ac95ab017606a8fdfecdc565', 'msg_type': 'execute_reply', 'version': '5.2.0', 'date': datetime.datetime(2020, 9, 4, 23, 20, 7, 436112, tzinfo=datetime.timezone.utc)}, 'msg_id': 'c15aee67-cbce-4700-a9b1-6a3acf8a87c3', 'msg_type': 'execute_reply', 'parent_header': {'msg_id': '13b75548-ac95ab017606a8fdfecdc565_2', 'username': 'vsts', 'session': '13b75548-ac95ab017606a8fdfecdc565', 'msg_type': 'execute_request', 'version': '5.3'}, 'metadata': None, 'content': {'status': 'abort', 'execution_count': None, 'payload': [], 'user_expressions': None}, 'buffers': []}`

I have not yet been successful in attempting to reproduce this locally.

**Sample log**
For example, see this log from a recent CI build:
https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=14187&view=logs&j=9e36a536-c213-5644-c2ab-401b20f4b829&t=57baf6f6-621f-5161-f318-42bbd94a0177&l=279

```
2020-09-04T23:21:06.5390751Z =================================== FAILURES ===================================
2020-09-04T23:21:06.5391716Z _____________________________ test_empty_workspace _____________________________
2020-09-04T23:21:06.5392615Z 
2020-09-04T23:21:06.5393041Z     def test_empty_workspace():
2020-09-04T23:21:06.5393499Z         """
2020-09-04T23:21:06.5393988Z         Tests behavior of a mock workspace with no providers.
[...]

PR #317, opened 2020-09-14
End-to-end build

anpaz-msft:


rmshaffer:
FYI, the pytest failures in your CI run are due to issue https://github.com/microsoft/iqsharp/issues/310, which I'm actively working on (I believe it requires a fix in jupyter-core).

build/e2e.yml
```diff
@@ -0,0 +1,33 @@
...
+  - minimal
+  - normal
+  - full
+
+trigger: none
+
+pr:
+- main
+- feature/*
+- features/*
```
rmshaffer, line 18: Just curious, why does this one contains `features/*` but `ci.yml` doesn't?

build/ci.yml
```diff
@@ -1,6 +1,11 @@
 name: $(Build.Major).$(Build.Minor).$(date:yyMM).$(DayOfMonth)$(rev:rr)
-trigger:
-- master
+
+trigger: none
```
rmshaffer, line 3: Why not trigger on commits to `main` here? Will that happen somewhere else?

rmshaffer (dismissed):

build/ci.yml
anpaz-msft: There is not much value to trigger another build on master, and we have limited number of agents for public projects.
Release builds on push to master are indeed triggered on a different build.

rmshaffer (approved):

Issue #318, opened 2020-09-14
Make `QSharpCallable` callable


guenp:
**Is your feature request related to a problem? Please describe.**
A common use case for researchers is to use the Q# Libraries from Python. Currently, to call any library operation or function from Python, one way to do this is the following:

```python
import qsharp
qsharp.packages.add("Microsoft.Quantum.Chemistry")
qsharp.reload()
estimate_energy = qsharp.QSharpCallable("Microsoft.Quantum.Chemistry.JordanWigner.VQE.EstimateEnergy", "")
estimate_energy.simulate(...)
```

However, when I (accidentally) run
```python
estimate_energy(...)
```

This throws the following error:
```
---------------------------------------------------------------------------
 TypeError Traceback (most recent call last)  in  ----> 1 estimate_energy()
 TypeError: 'QSharpCallable' object is not callable
```

This is a bit counter intuitive to me since something with the name "callable" seems like it would be callable.

**Describe the solution you'd like**
I suggest to override the `__call__` method on `QSharpCallable` and have it run `simulate()`.

**Describe alternatives you've considered**
Other ways to run this would be `toffoli_simulator` or `resource_estimator`, but I think it makes most sense to make `simulate()` the default.

PR #321, opened 2020-09-16
override QSharpCallble.__call__

guenp:
See issue #318

### DIFF
--- a/src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
@@ -11,7 +11,8 @@ def test_simulate():
     """
     from Microsoft.Quantum.SanityTests import HelloQ, HelloAgain
     assert HelloQ() == HelloQ.simulate() == ()
-    assert HelloAgain(1, "Ada") == HelloAgain.simulate(1, "Ada")
+    assert HelloAgain(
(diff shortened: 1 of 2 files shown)


src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
```diff
@@ -6,11 +6,11 @@
...
 def test_simulate():
     """
-    Checks that a simple simulate works correctly
+    Checks that a simple simulate works correctly, both using callable() and 
+    callable.simulate()
     """
     from Microsoft.Quantum.SanityTests import HelloQ
-    r = HelloQ.simulate()
-    assert r == ()
+    assert HelloQ() == HelloQ.simulate() == ()
```
rmshaffer, line 13: Would it be worth also adding a line here using a simple operation that takes arguments (e.g. `HelloAgain`) to verify that arguments are handled correctly in both cases?

rmshaffer (approved):
Thanks, this is a nice usability improvement!

src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
guenp: Sure

rmshaffer (approved):

PR #322, opened 2020-09-16
Usability improvements to Python interoperability for Q#

### LINKED ISSUE #19: Support passing NumPy arrays from Python host library
Often times, it will be convienent for Q# operations called from Python to directly accept NumPy arrays as input. While this doesn't confer any performance advantage, given the serialization to JSON and back, it is a significant enough usability improvement that it might be worth adopting.

### LINKED ISSUE #21: Handle missing dependency errors during Python init
Currently, there is a couple of dependencies needed for the `qsharp` Python module, namely:
- .net core 2.2.0
- IQ#
- Jupyter

So, it's not too hard to miss any and have problems when importing the qsharp module. During init, we should catch errors and point users to the installation instructions where they can verify that they followed all steps.

### LINKED ISSUE #126: Allow configuration of DumpMachine output format from IQ# and Python
The new DumpMachine outputs (visual version in Q# Notebooks and different text version in Python-driven invocations) look really neat, but the difference of formats across different mediums can be confusing. It can also cause issues if any [automated tools](https://github.com/microsoft/QuantumKatas/tree/master/utilities/DumpUnitary) or any [educational content](https://github.com/microsoft/QuantumKatas/pull/330) rely on the specific output format.

1. It would be great to document the different available formats of the output at https://docs.microsoft.com/quantum. 
2. If would be great to make different formats configurable from the user interface (for example, so that a Q# Notebook user can choose whether they want a visual format, a neat text format or an old-style text format). I believe some parts of the output are already configurable (such as little endian vs big endian vs bit string), but it's really hard to discover that without documentation.
3. It would also be helpful to keep numeric representation of state phase (possibly also configurable) - arrows look nice but without the base information about which direction corresponds to what phase they are less informative than I'd like them to be.

### LINKED ISSUE #212: Simulation returns 'None' in Python for Q# operations returning tuple of length greater than 7
**Describe the bug**
When the total number of return values from any _operation_ is more than 7 in qs code, then the python's _operation.simulate()_ returns `None`

**To Reproduce**
1. Can take any operation with total number of return items more than 7 : Example
`
	operation Ret8Fail() : (Int, Int, Int, Int, Int, Int, Int, Int)
    {
        body
        {
            return (1, 0, 1, 1, 0, 1, 1, 0);
        }
    }`

2. Run a simple python script to read above values : 
  `fail_8 = Ret8Fail.simulate()
print(f'When total no. of returns >7 : {fail_8}')`
3. Check the Result : `None` is returned

**Expected behavior**
Should get the correct values from the .qs operation : in above case `(1, 0, 1, 1, 0, 1, 1, 0)`

**Screenshots**

image

**System information**
 - OS: Windows 10
 - Python 3.8.3
 - Version : 'iqsharp': '0.11.2006.403', 'Jupyter Core': '1.3.60623.0', '.NET Runtime': '.NETCoreApp,Version=v3.1', 'qsharp':  '0.12.2007.124'

**Additional context**
Python: 
```import qsharp
from TestingReturns import Ret8Fail, Ret7Pass

print(qsharp.component_versions())

fail_8 = Ret8Fail.simulate()
print(f"When total no. of returns >7 : {fail_8}")

pass_7 = Ret7Pass.simulate()
print(f"When total no. returns are <=7 : {pass_7}")
```
Q#:
```
namespace TestingReturns
{
    open Microsoft.Quantum.Primitive;
    open Microsoft.Quantum.Canon;

	operation Ret8Fail() : (Int, Int, Int, Int, Int, Int, Int, Int)
    {
        body
        {
            return (1, 0, 1, 1, 0, 1, 1, 0);
        }
    }

	operation Ret7Pass() : (Int, Int, Int, Int, Int, Int, Int)
    {
        body
        {
            return (1, 1, 0, 0, 11, 1, 1);
        }
    }
}
```

### LINKED ISSUE #256: Cannot pass List[qsharp.Pauli] to .simulate()
**Describe the bug**
When I pass a list of Pauli operators to the `simulate` method, I get the following error:
```
In [2]: %run syndrome.py -q 5                                                                                                                                                      
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/workspace/Quantum/samples/error-correction/syndrome/syndrome.py in <module>
     33         input_values=input_values,
     34         encoding_bases=encoding_bases,
---> 35         qubit_indices=qubit_indices)
     36     ancilla, data = result
     37 

/usr/local/lib/python3.7/site-packages/qsharp/loader.py in simulate(self, **kwargs)
     82         machine, returning its output as a Python object.
     83         """
---> 84         return qsharp.client.simulate(self, **kwargs)
     85 
     86     def toffoli_simulate(self, **kwargs) -> Any:

(...)

/usr/local/lib/python3.7/json/encoder.py in default(self, o)
    177 
    178         """
--> 179         raise TypeError(f'Object of type {o.__class__.__name__} '
    180                         f'is not JSON serializable')
    181 
TypeError: Object of type int64 is not JSON serializable
```

**To Reproduce**
1. Create a Q# namespace with a function that takes `Pauli[]` as input
2. Import said namespace into Python
3. Create list of Pauli enums e.g. `pauli_list = [qsharp.Pauli.X, qsharp.Pauli.Y, qsharp.Pauli.Z]`
4. Run `MyNameSpace.simulate(pauli_list)`
5. See above for error traceback

**Expected behavior**
I would expect the list of `Pauli` enums to be passed to the simulated Q# function in the same way as a single `Pauli` enum (I would expect the function to run without any errors)

**System information**
 - OS: Unix-based Docker container (from `samples` image)

**Additional context**
See code [here](https://github.com/microsoft/Quantum/pull/399/files#diff-5c4b206990ac84956e99febf22b04139)

### LINKED ISSUE #324: TypeError when Q# errors are propagated to Python
**Describe the bug**
Whenever there is a bug in a Q# program that I'm trying to access from Python, the REPL returns the following error:
```python
TypeError: string indices must be integers
In [4]: %debug
> /usr/local/lib/python3.7/site-packages/qsharp/clients/iqsharp.py(225)log_error()
    223 
    224         def log_error(msg):
--> 225             if msg['msg_type'] == 'stream' and msg['content']['name'] == 'stderr':
    226                 errors.append(msg['content']['text'])
    227
```

When I enter debug mode (via `%debug`):
```python
ipdb> msg
'/workspaces/ieee-quantum-week-2020/session2/Chemistry/Trotterization.qs(12,16): error QS3238: Expecting a argument tuple of compile time constants.\n'
```

**To Reproduce**
1. Create a new Q# project
2. Create a new .qs file with an entrypoint that does something invalid (e.g. has an unnamed `Tuple` as input argument)
3. From Python, run `import qsharp` and `qsharp.reload()`

**Expected behavior**
Still an error but not a `TypeError` with said traceback

**System information**
Running in iqsharp-based Docker container (`mcr.microsoft.com/quantum/iqsharp-base:0.12.20082513`)

CC @rmshaffer

rmshaffer:
This PR contains a number of usability improvements to Python interoperability for Q#:

1. Support for passing NumPy arrays and scalar types as arguments to Q# callables. (see #19, #256)
1. Support Q# callables which have long tuples (>7 items) as arguments and/or return values. (see #212)
1. Add a config option to force plain-text encoding for state visualization in HTML-enabled clients. (see #126)
1. Add Python APIs for getting and setting IQ# configuration options. (see #126)
1. Fix display of error messages from IQ# kernel. (see #324)
1. Better error message with link to documentation when IQ# kernel is not installed or fails to initialize. (see #21)

Fixes #19, fixes #21, fixes #126, fixes #212, fixes #256, fixes #324.

### DIFF
--- a/src/Python/qsharp-core/qsharp/__init__.py
+++ b/src/Python/qsharp-core/qsharp/__init__.py
@@ -21,6 +21,7 @@ from distutils.version import LooseVersion
 from qsharp.clients import _start_client
 from qsharp.clients.iqsharp import IQSharpError
 from qsharp.loader import QSharpCallable, QSharpModuleFinder
+from qsharp.config import Config
 from qsharp.packages import Packages
 from qsharp.projects import Projects
 from qsharp.types import Result, Pauli
@@ -35,7 +36,7 @@ __all__ = [
     'compile', 'reload',
     'get_available_operations', 'get_available_operations_by_namespace',
     'get_workspace_operations',
-    'get_config', 'set_config', 'save_config',
+    'config',
     'packages',
     'projects',
     'IQSharpError',
@@ -89,34 +90,6 @@ def get_workspace_operations() -> List[str]:
     """
     return client.get_workspace_operations()
 
-def get_config() -> Dict[str, object]:
-    """
-    Returns a dictionary containing the names and values of the configuration
-    options. Options can be set by calls to qsharp.set_config() or by loading
-    a file previously created by qsharp.save_config().
-    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
-    of supported configuration setting names and values.
-    """
-    return client.get_config()
-
-def set_config(name : str, value : object) -> None:
-    """
-    Sets a specified configuration option with the specified value.
-    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
-    of supported configuration setting names and values.
-    """
-    client.set_config(name, value)
-
-def save_config() -> None:
-    """
-    Saves all current configuration options to a file named .iqsharp-config.json
-    in the current working directory. This file is automatically loaded
-    by the IQ# kernel at initialization time.
-    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
-    of supported configuration setting names and values.
-    """
-    client.save_config()
-
 def get_available_operations_by_namespace() -> Dict[str, List[str]]:
     """
     Returns a dictionary from namespaces to lists of operations and functions
@@ -148,6 +121,7 @@ def component_versions() -> Dict[str, LooseVersion]:
 ## STARTUP ##
 
 client = _start_client()
+config = Config(client)
 packages = Packages(client)
 projects = Projects(client)
 
--- a/src/Python/qsharp-core/qsharp/__init__.py
+++ b/src/Python/qsharp-core/qsharp/__init__.py
@@ -91,19 +91,29 @@ def get_workspace_operations() -> List[str]:
 
 def get_config() -> Dict[str, object]:
     """
-    TODO: document
+    Returns a dictionary containing the names and values of the configuration
+    options. Options can be set by calls to qsharp.set_config() or by loading
+    a file previously created by qsharp.save_config().
+    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
+    of supported configuration setting names and values.
     """
     return client.get_config()
 
 def set_config(name : str, value : object) -> None:
     """
-    TODO: document
+    Sets a specified configuration option with the specified value.
+    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
+    of supported configuration setting names and values.
     """
     client.set_config(name, value)
 
 def save_config() -> None:
     """
-    TODO: document
+    Saves all current configuration options to a file named .iqsharp-config.json
+    in the current working directory. This file is automatically loaded
+    by the IQ# kernel at initialization time.
+    See https://docs.microsoft.com/qsharp/api/iqsharp-magic/config for the list
+    of supported configuration setting names and values.
     """
     client.save_config()
 
--- a/src/Python/qsharp-core/qsharp/__init__.py
+++ b/src/Python/qsharp-core/qsharp/__init__.py
@@ -35,6 +35,7 @@ __all__ = [
     'compile', 'reload',
     'get_available_operations', 'get_available_operations_by_namespace',
     'get_workspace_operations',
+    'get_config', 'set_config', 'save_config',
     'packages',
     'projects',
     'IQSharpError',
@@ -87,6 +88,24 @@ def get_workspace_operations() -> List[str]:
     in the current workspace, excluding dynamically compiled snippets.
     """
     return client.get_workspace_operations()
+
+def get_config() -> Dict[str, object]:
+    """
+    TODO: document
+    """
+    return client.get_config()
+
+def set_config(name : str, value : object) -> None:
+    """
+    TODO: document
+    """
+    client.set_config(name, value)
+
+def save_config() -> None:
+    """
+    TODO: document
+    """
+    client.save_config()
 
 def get_available_operations_by_namespace() -> Dict[str, List[str]]:
     """
--- a/src/Python/qsharp-core/qsharp/clients/iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/clients/iqsharp.py
@@ -241,8 +241,7 @@ class IQSharpClient(object):
         errors = []
 
         def log_error(msg):
-            if msg['msg_type'] == 'stream' and msg['content']['name'] == 'stderr':
-                errors.append(msg['content']['text'])
+            errors.append(msg)
 
         handlers = {
             'execute_result': (lambda msg: results.append(msg))
--- a/src/Python/qsharp-core/qsharp/clients/iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/clients/iqsharp.py
@@ -127,6 +127,25 @@ class IQSharpClient(object):
     def reload(self) -> None:
         return self._execute(f"%workspace reload", raise_on_stderr=True)
 
+    def get_config(self) -> Dict[str, object]:
+        raw = self._execute(f"%config", raise_on_stderr=True)
+        config_settings = {}
+        for row in raw["rows"]:
+            config_settings[row['Key']] = row['Value']
+        return config_settings
+
+    def set_config(self, name : str, value : object) -> None:
+        from numbers import Number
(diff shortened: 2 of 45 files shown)


src/Python/qsharp-core/qsharp/__init__.py
```diff
@@ -88,6 +89,34 @@ def get_workspace_operations() -> List[str]:
     """
     return client.get_workspace_operations()
 
+def get_config() -> Dict[str, object]:
```
cgranade, line 92: We may want to consider exposing the config as a dict-like object, similar to how matplotlib exposes `rcParams`. E.g.:
```
class QSharpConfig(object):
    def __init__(self, client):
        self.client = client
    def __getitem__(self, item : str) -> Any:
        return self.client.get_config()[item]
    def __setitem__(self, item : str, value : object) -> None:
        self.client.set_config(self, item, value)
    def save(self):
        self.client.save_config() 

config = Config(client)
```

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -127,6 +127,25 @@ def get_workspace_operations(self) -> List[str]:
...
+        return config_settings
+
+    def set_config(self, name : str, value : object) -> None:
+        from numbers import Number
+        if isinstance(value, bool):
+            self._execute(f"%config {name}={'true' if value else 'false'}", raise_on_stderr=True)
+        elif isinstance(value, Number):
+            self._execute(f"%config {name}={value}", raise_on_stderr=True)
+        else:
+            self._execute(f"%config {name}='{value}'", raise_on_stderr=True)
```
cgranade, line 144: We may want to consolidate this with existing support for preparing values for JSON serialization (`map_tuples` and `unmap_tuples`).

src/Python/qsharp-core/qsharp/serialization.py
```diff
@@ -7,6 +7,11 @@
 # Licensed under the MIT License.
 ##
 
+try:
+    import numpy as np
+except ImportError:
```
cgranade, line 12: We should check, but it may be possible for importing numpy to raise an exception other than `ImportError` if NumPy has been installed, but is broken. In that case, we may still want to fail hard, but it may be worth catching that exception so to give a better error message.

src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
```diff
@@ -85,6 +163,28 @@ def test_multi_compile():
     assert r == qsharp.Result.One
 
 
+def test_config():
+    """
+    Verifies get and set of config settings
+    """
+    config = qsharp.get_config()
+    assert len(config) == 0
```
cgranade, line 171: Will those go badly if the test environment somehow has a `.iqsharp-config.json` file?

cgranade (approved):

src/Python/qsharp-core/qsharp/__init__.py
rmshaffer: Yes, that's an excellent idea.

src/Python/qsharp-core/qsharp/tests/test_iqsharp.py
rmshaffer: Yes, good point. With the refactoring into a separate `Config` object, I don't think these checks really add much value, so I'll just remove them.

src/Python/qsharp-core/qsharp/clients/iqsharp.py
rmshaffer: Yes, actually today `%config` doesn't support JSON-style arguments. It needs to be updated to share the argument-parsing code with `%simulate`, etc., so that we can pass arguments in JSON format as with the other magic commands. I've opened #323 to track this.

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -127,6 +127,25 @@ def get_workspace_operations(self) -> List[str]:
     def reload(self) -> None:
         return self._execute(f"%workspace reload", raise_on_stderr=True)
 
+    def get_config(self) -> Dict[str, object]:
+        raw = self._execute(f"%config", raise_on_stderr=True)
```
guenp, line 131: is there a reason the functions in this module don't have docstrings?

src/Python/qsharp-core/qsharp/serialization.py
```diff
@@ -17,11 +25,17 @@ def map_tuples(obj):
         result = {
             '@type': 'tuple'
         }
-        for i in range(len(obj)):
+        
+        # For tuples of more than 7 items, the .NET type is ValueTuple<T1,T2,T3,T4,T5,T6,T7,TRest>.
+        # Items beyond Item7 must be nested inside a key called "Rest".
+        maxTupleLength = 7
```
guenp, line 31: PEP8 recommends using `snake_case`
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

rmshaffer: Good catch! I think the comment here about .NET types must have kicked my mind out of Python mode... 😄

src/Python/qsharp-core/qsharp/clients/iqsharp.py
cgranade: That was my intent with initial versions, yeah; that separation allows for the mock client to be substituted into the package easily during unit tests and such.

cgranade (approved):

src/Python/qsharp-core/qsharp/clients/iqsharp.py
guenp: Thanks both! I would still argue we could at some point add docstrings here, but happy to discuss :)

cgranade: I don't think there's an argument against, so much as just that they're being internal-only means we didn't get to it yet.

PR #330, opened 2020-09-29
New simulator performance events.

cgranade:
This PR opens for early feedback an enhancement that logs performance data from the `%simulate` command as events for later analysis. To accomplish this, this PR also splits `PerformanceMonitor` into an interface and implementation as per best practices for other services.

TODO:
- [ ] API comments
- [ ] Unit tests (not sure of the right practice for testing these events; happy for suggestions)


src/Kernel/Magic/Simulate.cs
```diff
@@ -94,12 +97,26 @@ public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
...
+            {
+                maxNQubits = System.Math.Max(qsim.QubitManager.GetAllocatedQubitsCount(), maxNQubits);
+            };
+            var stopwatch = new Stopwatch();
             var value = await symbol.Operation.RunAsync(qsim, inputParameters);
-            return value.ToExecutionResult();
+            stopwatch.Stop();
+            var result = value.ToExecutionResult();
+            (Monitor as PerformanceMonitor)?.ReportSimulatorPerformance(new SimulatorPerformanceArgs(
+                simulatorName: typeof(SimulateMagic).FullName,
```
rmshaffer, line 115: Would it make sense to report the name of the simulator itself (e.g., `qsim.GetType().FullName`) rather than the name of the magic command?

rmshaffer (commented):
I know this is still a draft, but it makes sense to me! Just one comment.

src/Kernel/Magic/Simulate.cs
cgranade: Good catch, thanks; I'll modify accordingly. That should also be a lot easier to use if we generalize how we handle different simulators in the future.

cgranade:
I always forget to adapt the unit tests when DI configuration changes, but that should be fixed now.

rmshaffer:
Regarding unit tests: perhaps you could add something to `JupyterActions` in `TelemetryTests.cs`?

https://github.com/microsoft/iqsharp/blob/762a822fde713dcea896f78cbad299f3ba31de18/src/Tests/TelemetryTests.cs#L265-L270

That test already validates that `%simulate` produces a particular event; if your new event should also be produced at that point, should validation of the `SimulatorPerformance` event be added there?

As for the `KernelPerformance` event, that's a bit trickier since it just happens in the background. Maybe you can manually trigger it in a test... e.g., perhaps expose an `internal` method such as `PerformanceMonitor.ReportKernelPerformance()`, which you could invoke from a test?

cgranade:
> Regarding unit tests: perhaps you could add something to `JupyterActions` in `TelemetryTests.cs`?
> 
> https://github.com/microsoft/iqsharp/blob/762a822fde713dcea896f78cbad299f3ba31de18/src/Tests/TelemetryTests.cs#L265-L270
> 
> That test already validates that `%simulate` produces a particular event; if your new event should also be produced at that point, should validation of the `SimulatorPerformance` event be added there?
> 
> As for the `KernelPerformance` event, that's a bit trickier since it just happens in the background. Maybe you can manually trigger it in a test... e.g., perhaps expose an `internal` method such as `PerformanceMonitor.ReportKernelPerformance()`, which you could invoke from a test?

That actually raises another subtle issue, namely that the number of events captured by the logger in that test is now non-deterministic, since it depends on how many `KernelPerformance` events are raised during the interim. I'd expect new test failures as a result... hm.

cgranade:
I think this should work; it's a bit of a refactoring of `PerformanceMonitor`, but it should be a lot simpler to control now.

src/Tool/Telemetry.cs
```diff
@@ -226,6 +232,27 @@ public static EventProperties AsTelemetryEvent(this ExecutedEventArgs info)
...
             return evt;
         }
+
+        public static EventProperties AsTelemetryEvent(this SimulatorPerformanceArgs info)
+        {
+            var evt = new EventProperties() { Name = "SimulatorPerformance".WithTelemetryNamespace() };
+
+            evt.SetProperty("SimulatorName".WithTelemetryNamespace(), info.SimulatorName);
+            evt.SetProperty("NQubits".WithTelemetryNamespace(), info.NQubits);
+            evt.SetProperty("Duration".WithTelemetryNamespace(), info.Duration);
```
rmshaffer, line 242: Looks like this will be necessary for the CI to succeed:
```suggestion
            evt.SetProperty("Duration".WithTelemetryNamespace(), info.Duration.ToString());
```

src/Tests/TelemetryTests.cs
```diff
@@ -264,10 +269,22 @@ public void JupyterActions()
...
             Assert.AreEqual("%simulate", logger.Events[count].Properties["Quantum.IQSharp.Command"]);
             Assert.AreEqual("Ok", logger.Events[count].Properties["Quantum.IQSharp.Status"]);
+            count++;
+            Assert.AreEqual("Quantum.IQSharp.SimulatorPerformance", logger.Events[count].Name);
+            Assert.AreEqual(typeof(QuantumSimulator).FullName, logger.Events[count].Properties["Quantum.IQSharp.SimulatorName"]);
+            Assert.AreEqual("0", logger.Events[count].Properties["Quantum.IQSharp.NQubits"]);
+            // NB: Not testing Duration, since that is non-determinstic.
+
+            // Make sure that forcing a performance report works.
+            performanceMonitor.Report();
```
rmshaffer, line 285: I think incrementing `count` is necessary here?
```suggestion
            count++;
            performanceMonitor.Report();
```

rmshaffer (approved):
Looks good to me! Happy to quickly approve again once the E2E CI issues are resolved.

rmshaffer (approved):

cgranade:
It looks like the e2e builds failed, but with no error message such that I suspect a transient failure; reran tests, and hopefully that should work.

rmshaffer (dismissed):

vxfield (approved):
Made some minor fixes:
1) Started StopWatch on the Simulator performance tracking
2) Changed the format of telemetry duration TimeSpans to include milliseconds
3) Added unit tests asserts for Duration>0

PR #331, opened 2020-10-01
Initialize workspace asynchronously

rmshaffer:
In cases where we load a significant number of packages and/or compile a significant amount of Q# code at kernel initialization time, it's possible to exceed Jupyter's default 60-second timeout for kernel_info_reply, which results in an error message being displayed to the user (even though the initialization does eventually succeed).

In addition, client-side features such as Q# syntax highlighting don't light up until the kernel_info_reply is received.

This change moves the workspace initialization to be asynchronous, such that the kernel_info_reply can be sent prior to loading the packages or compiling the Q# code in the workspace. Cell execution will now wait for the workspace initialization to complete before attempting to execute the cell.

This PR is intended to address issues discovered in CI runs and local testing for https://github.com/microsoft/QuantumKatas/pull/468 and https://github.com/microsoft/QuantumKatas/pull/494.

Here is a diagram showing how this changes the behavior as compared to the August IQ# release (0.12.20082513) and September IQ# release (0.12.20092803):
image

The change in this PR should fix two problems:

1. Loading M.Q.Standard.Visualization in the background caused some non-deterministic behavior depending on when the package finished loading because the package-loaded listeners that register display encoders were not thread-safe.
   - I believe this was the cause of the failures in the CI tests in https://github.com/microsoft/QuantumKatas/pull/494.
	
1. When loading project references, it's likely that loading package references and Q# compilation will take much longer than it would take otherwise. So those should be async to avoid delaying the kernel info reply too long.		
   - This was the cause of the CI failures in https://github.com/microsoft/QuantumKatas/pull/468. This is needed to enable project/package references in the Katas projects, without running the risk of Jupyter Notebook potentially showing the kernel info timeout message.

### DIFF
--- a/src/Core/Workspace/IWorkspace.cs
+++ b/src/Core/Workspace/IWorkspace.cs
@@ -138,5 +138,11 @@ namespace Microsoft.Quantum.IQSharp
         /// Triggers the workspace to be reloaded from disk.
         /// </summary>
         void Reload(Action<string> statusCallback = null);
+
+        /// <summary>
+        /// Waits for the initial workspace load to complete, including
+        /// package loads and project compilation.
+        /// </summary>
+        void WaitForInitialization();
     }
 }
--- a/src/Core/Workspace/IWorkspace.cs
+++ b/src/Core/Workspace/IWorkspace.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-
+using System.Threading.Tasks;
 using Microsoft.Quantum.IQSharp.Common;
 
 namespace Microsoft.Quantum.IQSharp
@@ -140,9 +140,10 @@ namespace Microsoft.Quantum.IQSharp
         void Reload(Action<string> statusCallback = null);
 
         /// <summary>
-        /// Waits for the initial workspace load to complete, including
-        /// package loads and project compilation.
+        /// Task that will be completed when the initial workspace
+        /// initialization has finished, including package loads and
+        /// project compilation.
         /// </summary>
-        void WaitForInitialization();
+        Task Initialization { get; }
     }
 }
--- a/src/Core/Workspace/Workspace.cs
+++ b/src/Core/Workspace/Workspace.cs
@@ -183,11 +183,11 @@ namespace Microsoft.Quantum.IQSharp
             logger?.LogInformation($"Starting IQ# Workspace:\n----------------\nRoot: {Root}\nCache folder:{CacheFolder}\nMonitoring changes: {MonitorWorkspace}\nUser agent: {metadata?.UserAgent ?? "<unknown>"}\nHosting environment: {metadata?.HostingEnvironment ?? "<unknown>"}\n----------------");
 
             // Initialize the workspace asynchronously
-            Task.Run(async () =>
+            Task.Run(() =>
             {
                 try
                 {
-                    await GlobalReferences.LoadDefaultPackages();
+                    GlobalReferences.LoadDefaultPackages();
                     ResolveProjectReferences();
 
                     if (!LoadFromCache())
--- a/src/Core/Workspace/Workspace.cs
+++ b/src/Core/Workspace/Workspace.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Xml.Linq;
 using System.Xml.XPath;
 using Microsoft.CodeAnalysis;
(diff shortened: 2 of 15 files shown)


src/Core/Workspace/Workspace.cs
```diff
@@ -382,6 +422,8 @@ private void LoadReferencedPackages(Action<string> statusCallback = null)
         /// <inheritdoc/>
         public void AddProject(string projectFile)
```
cgranade, line 423: It looks like both this and `Reload` are blocking due to calls to `WaitForInitialization`; are they called from async methods further up?

cgranade (commented):
I think this looks good, just had the one main question about `WaitForInitialization`. Thanks!

src/Core/Workspace/Workspace.cs
rmshaffer: In IQ#, those methods are currently called only from the `Run()` method of magic command implementations (`%workspace reload` and `%project`), and yes, those are being called in an async call via `AbstractMagic.SafeExecute()`.

I take your point though that both `Reload()` and `AddProject()` could be long-running calls and should be async APIs themselves. Actually, `Reload()` was already long-running since it is synchronously waiting for compilation and package downloads. So, while maybe it should be async, I would like to avoid introducing that API change at the moment, if possible.

And upon further thought, there is no reason for calling `WaitForInitialization()` from inside `AddProject()`, since that method is not actually doing any compilation (and the magic command entry point will anyway not be invoked until the initialization has completed). I'll remove that one.

src/Core/Workspace/Workspace.cs
```diff
@@ -172,22 +182,52 @@ public string CacheFolder
...
+                            StartFileWatching();
+                        }
+                    }
+
+                    eventService?.TriggerServiceInitialized<IWorkspace>(this);
+                }
+                finally
+                {
+                    initialized.Set();
+                }
```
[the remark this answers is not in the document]
rmshaffer: Yes, I believe it should be in a separate thread (that's what `Task.Run()` does, if I'm understanding correctly). Previously this was just blocking in the constructor, which meant that all of this was happening synchronously while the `Workspace` object was being created.

Is that what you were asking, or did I misunderstand your question?

src/Core/Workspace/IWorkspace.cs
```diff
@@ -138,5 +138,11 @@ public interface IWorkspace
         /// Triggers the workspace to be reloaded from disk.
         /// </summary>
         void Reload(Action<string> statusCallback = null);
+
+        /// <summary>
+        /// Waits for the initial workspace load to complete, including
+        /// package loads and project compilation.
+        /// </summary>
+        void WaitForInitialization();
```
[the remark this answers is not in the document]
rmshaffer: Good suggestion! That sounds more appropriate.

anpaz:
>                 }

Should we not wait here then, if we want to load this async?

---
Refers to: src/Core/References/References.cs:135 in ff9efa1. [](commit_id = ff9efa10d186957d34d319a3d68c1edd8f5c54c2, deletion_comment = False)

rmshaffer:
> > ```
> >             }
> > ```
> 
> Should we not wait here then, if we want to load this async?
> 
> Refers to: src/Core/References/References.cs:135 in [ff9efa1](https://github.com/microsoft/iqsharp/commit/ff9efa10d186957d34d319a3d68c1edd8f5c54c2). [](commit_id = [ff9efa1](https://github.com/microsoft/iqsharp/commit/ff9efa10d186957d34d319a3d68c1edd8f5c54c2), deletion_comment = False)

So `AutoLoadPackages` here is typically just `Microsoft.Quantum.Standard`. That package has always been loaded synchronously at startup, and I wanted to avoid changing that behavior here. Anything in `DeferredLoadPackages` (which typically just contains `Microsoft.Quantum.Standard.Visualization`) will be loaded asynchronously. 

Although I certainly agree with your point, and we could consider moving `Microsoft.Quantum.Standard` into `DeferredLoadPackages`. But, for example, in some places we rely on `python -c "import qsharp"` to help populate NuGet caches, etc., and changing the load strategy for `Microsoft.Quantum.Standard` may "break" that.

src/Core/Workspace/Workspace.cs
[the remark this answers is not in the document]
cgranade: Maybe I was missing something with GitHub's web UI, sorry, but this looked like it was outside of the `Task.Run` call for some reason, sorry. If it is all in `Task.Run`, I agree entirely with this change, then.

src/Core/Workspace/Workspace.cs
cgranade: Makes sense!

src/Core/Workspace/Workspace.cs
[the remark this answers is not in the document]
rmshaffer: Got it! Yes, the diff here does look a little bit weird, but it's definitely all inside `Task.Run`.

cgranade (approved):

src/Core/Workspace/Workspace.cs
```diff
@@ -146,17 +148,25 @@ public string CacheFolder
...
         public string CacheFolder { get; set; }
 
+        /// <summary>
+        /// An event that is set after the workspace initialization has completed.
+        /// </summary>
+        private ManualResetEvent initialized = new ManualResetEvent(false);
+
+        /// <inheritdoc/>
+        public Task Initialization => Task.Run(() => initialized.WaitOne());
+
```
anpaz, line 158: Is this created before calling the constructor or after? Just making sure.

anpaz (approved):
:shipit:

src/Core/Workspace/Workspace.cs
rmshaffer: The `initialized` event above is a `ManualResetEvent` object which is created _before_ the constructor is called, and the event does not get set until the asynchronous initialization has completed. It never gets reset; once it's set, it's set forever.

As for `Initialization`, since it uses this expression-bodied property accessor via `=>`, it's actually creating a new `Task` for each caller who accesses `Workspace.Initialization`. The task is very lightweight; it just waits for the `initialized` event to be set (or, if it is already set, it just completes immediately).

anpaz (approved):

PR #334, opened 2020-10-06
Use documentation attributes to present more structured summaries of symbols

cgranade:
This PR uses the new attributes introduced in https://github.com/microsoft/qsharp-runtime/pull/390 to present more structured documentation summaries than can be generated from API documentation comments directly.


src/Kernel/Extensions.cs
```diff
@@ -23,5 +29,61 @@ public static void AddIQSharpKernel(this IServiceCollection services)
...
+            this OperationInfo operation, string attributeName,
+            string namespaceName = "Microsoft.Quantum.Documentation"
+        ) =>
+            operation.Header.Attributes.Where(
+                attribute =>
+                    attribute.TypeId.Item.Namespace.Value == namespaceName &&
+                    attribute.TypeId.Item.Name.Value == attributeName
+            );
+
+        internal static bool TryAsStringLiteral(this TypedExpression expression, [NotNullWhen(true)] out string value)
```
rmshaffer, line 43: Cool, wasn't aware of these post-condition attributes!

rmshaffer (approved):
Looks good!

src/Kernel/Extensions.cs
cgranade: C# 8 has some neat features, if a bit hidden!

PR #340, opened 2020-10-13
Fix DumpMachine to allow location-based output

swernli:
This fixes a bug in `DumpMachine` and `DumpRegister` where providing a location would fallback to the no-op implementation instead of the simulator implementation. I tested this by simulating an operation that included `DumpMachine("Foo.txt")` and saw that the file gets created with the expected content.

### DIFF
--- a/src/Tests/IQsharpEngineTests.cs
+++ b/src/Tests/IQsharpEngineTests.cs
@@ -262,6 +262,22 @@ namespace Tests.IQSharp
         }
 
         [TestMethod]
+        public async Task DumpToFile()
+        {
+            var engine = Init();
+
+            // Compile DumpMachine snippet.
+            await AssertCompile(engine, SNIPPETS.DumpToFile, "DumpToFile");
+
+            // Run, which should produce files in working directory.
+            await AssertSimulate(engine, "DumpToFile", "Dumped to file!");
+
+            // Ensure the expected files got created.
+            Assert.IsTrue(System.IO.File.Exists("DumpMachine.txt"));
+            Assert.IsTrue(System.IO.File.Exists("DumpRegister.txt"));
+        }
+
+        [TestMethod]
         public async Task UpdateDependency()
         {
             var engine = Init();
--- a/src/Tests/IQsharpEngineTests.cs
+++ b/src/Tests/IQsharpEngineTests.cs
@@ -273,8 +273,20 @@ namespace Tests.IQSharp
             await AssertSimulate(engine, "DumpToFile", "Dumped to file!");
 
             // Ensure the expected files got created.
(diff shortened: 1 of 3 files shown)


swernli:
@cgranade @rmshaffer This definitely feels like something I should add a test for, since we didn't notice for a long time that it was broken, but I'm not sure where the right place in the tree to add this kind of test is. It will need to execute a `DumpMachine` and a `DumpRegister` with a path provided and then confirm that the file gets created, ideally also validating the content. Any suggestions on where to put that?

cgranade (commented):
Looks good to me, thanks for finding that! Commenting only to make sure @rmshaffer gets a chance to look as well.

rmshaffer:
Thanks for finding and fixing this, @swernli!

There are a number of tests under `src/Tests/IQSharpEngineTests.cs` that compile and simulate operations from Q# snippets. (The snippets are defined in `src/Tests/SNIPPETS.cs`.)

You can probably define a snippet there calling `DumpMachine` with an argument, and then add a test method very similar to this one:
https://github.com/microsoft/iqsharp/blob/9deca20a374eeeff9c6ab1b679e170a6fc7c571f/src/Tests/IQsharpEngineTests.cs#L188-L199

Of course, after calling `AssertSimulate`, the test method could then verify that the expected file was created.

rmshaffer (dismissed):
Approving the change as-is, but if you'd like to also add a test, that's even better. Thanks!

swernli:
@rmshaffer Thanks for the tips on where to test! I added a test case for this that is working locally, and hopefully should do the right thing in the pipeline as well.

Looks like it dismissed your sign off with the new PR, so I'll need another one before this can merge. Thanks!

src/Tests/IQsharpEngineTests.cs
```diff
@@ -261,6 +261,22 @@ public async Task UpdateSnippet()
...
+
+            // Compile DumpMachine snippet.
+            await AssertCompile(engine, SNIPPETS.DumpToFile, "DumpToFile");
+
+            // Run, which should produce files in working directory.
+            await AssertSimulate(engine, "DumpToFile", "Dumped to file!");
+
+            // Ensure the expected files got created.
+            Assert.IsTrue(System.IO.File.Exists("DumpMachine.txt"));
+            Assert.IsTrue(System.IO.File.Exists("DumpRegister.txt"));
```
rmshaffer, line 277: Minor suggestion - could you then delete the files after verifying that they exist?

swernli: Sure thing, should be easy enough to add!

rmshaffer (approved):
Looks great, and thanks for adding the test!

PR #341, opened 2020-10-14
Remove unused open directives

swernli:
The tests for type loading from libraries have leftover open directives from an earlier version that are not used. These can cause problems in the e2e builds if contents of those libraries are changing. Removing the unused directives more clearly shows what the test depends on and does not incur any extra compilation of unused libraries.


cgranade:
Looks good to me, thanks for fixing and for the explanation!

cgranade (approved):

PR #345, opened 2020-10-20
%debug UX improvements and integration with ExecutionPathTracer

rmshaffer:
This PR brings a few improvements to the `feature/visualization-improvements` branch:

1. Initial integration of `ExecutionPathTracer` into the `%debug` flow. (This is still a little bit buggy while stepping through nested operations.)
1. Some minor UX design updates for `%debug` experience and related controls.
1. Improved experience when `%debug` finishes, either naturally or via kernel interrupt.
1. Some code refactoring to reduce duplication.
1. C# unit tests for `%debug`.
1. Selenium tests for `%debug` and `%trace`.

Here's a screenshot of a simple `%debug` call in action:

image


rmshaffer:
@cgranade, I need to do some more validation on this PR, but wanted to raise it now to give you a chance to take an early look, if you'd like.

theRoughCode:
Wow, this looks great! I must admit, this was a pretty huge PR so I tried to read as much as I could, but overall, lgtm. I like the refactoring you did with extracting the rendering logic for the execution path into its own extension method.

Just out of curiosity, what were the bugs you encountered when stepping through nested operations?

src/ExecutionPathTracer/ExecutionPathTracer.cs
```diff
@@ -37,7 +37,7 @@ public class ExecutionPathTracer
                         ? this.classicalRegisters[k].Count
                         : 0
                     )),
-                this.operations.ToList().WhereNotNull()
+                this.operations.Reverse().ToList().WhereNotNull()
```
theRoughCode, line 40: What was this change for? Was the operations returned in reverse order before? :O

theRoughCode:
Also, I really like the buttons and overall UI for `%debug`. Looks really neat and clean. Good job @rmshaffer!

rmshaffer:
Thanks for taking a look, @theRoughCode!

> Just out of curiosity, what were the bugs you encountered when stepping through nested operations?

I don't know that they are even "bugs", but just quirks of how the step-through is implemented. For example, stepping through nested operations first renders the child operations, and then once the parent operation completes, those child operations disappear and the parent operation appears.

And also, sometimes taking a step performs some operation that has no visual impact on either the state visualization _or_ the execution path, so that also feels weird.

But overall things really worked well! Didn't mean for that to sound negative!

src/ExecutionPathTracer/ExecutionPathTracer.cs
rmshaffer: Typically `operations` here would contain exactly one operation, because we are rendering the execution path for a single top-level operation. So the order is meaningless in that case.

But while stepping through nested operations, some child operations are basically orphaned because their parent operations haven't completed yet. Those show up in this `operations` list as well, but since it's a stack, the most recent item is first. So `Reverse()` just puts it into chronological order, as expected.

theRoughCode: Oh that makes sense. Yea, I guess that's because when I implemented the tracer, I didn't mean for the execution path to be retrieved in the middle of the execution like what we're doing with the state visualizer.

Thanks for the clarification!

rmshaffer: Yeah, I may be stretching the intended purpose of the execution path tracer a bit here... but the nice thing is that it works!

theRoughCode:
> For example, stepping through nested operations first renders the child operations, and then once the parent operation completes, those child operations disappear and the parent operation appears.

Ah I see. Is this for both the state visualizer and tracer or tracer only? If it's the tracer, it might be due to the fact that we render the execution path `onOperationEnd` which only gets called once the tracer has already traversed the whole stack of children and the deepest, last child is popped. Maybe the quick workaround for this is to render the top-most operation in the stack?

> sometimes taking a step performs some operation that has no visual impact on either the state visualization or the execution path, so that also feels weird.

I think I've encountered this before where there's some internal operation that's happening but is not visualized externally to the user 🤔

cgranade (approved):
Looks great to me, thanks for doing this! Just had a couple minor comments.

src/Kernel/client/plotting.ts
```diff
@@ -8,7 +8,7 @@
 // Using this hack, we can split the type import and
 // the runtime import apart, then glue them back
 // together using a declare global to solve TS2686.
-/// <amd-dependency path="/kernelspecs/iqsharp/chart.js" name="Chart" />
+/// <amd-dependency path="chart" name="Chart" />
```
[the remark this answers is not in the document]
rmshaffer: Yes, and I actually found this by running the Selenium test... 👀

src/Jupyter/Visualization/StateDisplayEncoders.cs
```diff
@@ -296,24 +296,23 @@ public StateVectorToHtmlResultEncoder(IConfigurationSource configurationSource)
                             "),
                             PhaseDisplayStyle.ArrowAndNumber => FormattableString.Invariant($@"
                                 <td>
-                                 <div style=""{StyleForAngle(amplitude.Phase)}""> ↑ </div> <div style=""{StyleForNumber()}"">{amplitude.Phase}</div>
+                                 <div style=""{StyleForAngle(amplitude.Phase)}""> ↑ </div>
+                                 <div style=""{StyleForNumber()}"">{amplitude.Phase:F4}</div>
```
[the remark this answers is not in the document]
rmshaffer: Yes, it's `dump.measurementDisplayPrecision`. I thought about that, but the name and documentation for that implies that it's used specifically for measurement probabilities. We actually use a hard-coded `F4` format string for displaying the amplitudes as well, so at least this is consistent with that...

src/Kernel/client/kernel.ts
```diff
@@ -52,55 +51,35 @@ class Kernel {
...
+                        // Create buttons to change plot style
+                        addToolbarButton(toolbarContainer, "Measurement Probability", event => update(debugSession, "amplitude-squared"));
+                        addToolbarButton(toolbarContainer, "Amplitude and Phase", event => update(debugSession, "amplitude-phase"));
+                        addToolbarButton(toolbarContainer, "Real and Imaginary", event => update(debugSession,  "real-imag"));
 
+                        // Create debug toolbar
+                        let debugContainer = createToolbarContainer("Debug controls:");
+                        debugContainer.className = "iqsharp-debug-toolbar";
+                        stateDiv.insertBefore(debugContainer, stateDiv.firstChild);
+                        addToolbarButton(debugContainer, "▶️ Next step", event => this.advanceDebugger(debugSession));
```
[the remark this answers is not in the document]
rmshaffer: Yeah, I looked around for something more "step"-like, but I am hoping the words "Next step" should make it clear what the function is...

rmshaffer:
Thanks @cgranade! I'm merging this now in order to get the `feature`->`main` PR going (and its CI runs), but I'm happy to follow up with additional changes if needed - please take a look at my responses to your comments and let me know what you think.

PR #346, opened 2020-10-20
Support manifest with no validation.

anpaz:
To support signing dlls on e2e builds, we need to get the list of potential assemblies from a module even if they were not built. The reason is that all built dlls need to be signed, but we don't build all assemblies in all submodules all the time (only a subset).
With this change the module can report the list of dlls it may generate and in the QDK we will check if the dll was built and sign it if necessary.
We will still validate that the dlls for the module getting built are generated.


rmshaffer (approved):

rmshaffer (approved):

build/manifest.ps1
```diff
@@ -1,23 +1,42 @@
...
 # Licensed under the MIT License.
 
-#!/usr/bin/env pwsh
-#Requires -PSEdition Core
+<#
+    .SYNOPSIS
+        Provides the list of artifacts (Packages and Assemblies) generated by this repository.
+    
+    .PARAMETER OutputFormat
+        Specifies if the output of this script should be a hashtable with the artificats
```
rmshaffer, line 9:
```suggestion
        Specifies if the output of this script should be a hashtable with the artifacts
```

rmshaffer (approved):

build/manifest.ps1
```diff
@@ -1,18 +1,30 @@
...
 
-@{
+$artifacts = @{
     Packages = @(
         "Microsoft.Quantum.IQSharp.Core",
         "Microsoft.Quantum.IQSharp.ExecutionPathTracer",
         "Microsoft.Quantum.IQSharp.Jupyter",
         "Microsoft.Quantum.IQSharp"
-    );
+    ) | ForEach-Object { Join-Path $Env:NUGET_OUTDIR "$_.$Env:NUGET_VERSION.nupkg" };
```
rmshaffer, line 26: Similar comment as in the other repos about `.snupkg`, though that's not currently applicable to this repo since it doesn't produce any `.snupkg` artifacts.

rmshaffer (approved):

build/manifest.ps1
anpaz: I think they do. Practically it doesn't matter as for nuget packages, as opposed to dlls, we submit for signing everything that we found in the NUGET_OUTDIR folder (which includes the .snupkgs). So, this is not really needed for signing.
We do use it to validate that the package is created on local builds. So for now, I'm just going to leave it like this.

PR #358, opened 2020-10-29
Load package perf improvements

anpaz:
Doing a couple of chagnes to improve the load of nuget packages:
1. Keeping alive the list of AvailablePackages. This was getting calculated every time that we want to add a package, so it was recalculating and redownloading the same package dependencies for common things like M.Q.Std and M.Q.Sim.
2. Remove the optimization to not download system packages. If the package is not installed better byte the bullet and download it, otherwise nuget will go to the server to fetch its dependencies.


rmshaffer (approved):

rmshaffer:
@anpaz is there anything else needed before we can merge this?

PR #359, opened 2020-10-30
Fixes for a few build/test issues

rmshaffer:
This PR contains fixes for a few small issues that should be addressed for build/test purposes:

1. Sets `IQSHARP_AUTO_LOAD_PACKAGES` to `$null` in HTTP server tests and in Python tests. This should avoid issues encountered when trying to perform E2E builds locally, since the default auto-load packages (Microsoft.Quantum.Standard and Microsoft.Quantum.Standard.Visualization) are built from the `QuantumLibraries` repo, which is downstream of this repo.

1. Avoid sending the `iqsharp_state_dump` IOPub message if the interactive histogram feature is not enabled. This is harmless in practice, but it produces ugly error messages when running `jupyter nbconvert` with notebooks that contain `DumpMachine()` calls.

1. Fixes a couple of build warnings in `SymbolResolver.cs`.

### DIFF
--- a/src/Jupyter/Visualization/StateDisplayOperations.cs
+++ b/src/Jupyter/Visualization/StateDisplayOperations.cs
@@ -46,6 +46,11 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// </summary>
         public BasisStateLabelingConvention BasisStateLabelingConvention { get; set; } = BasisStateLabelingConvention.Bitstring;
 
+        /// <summary>
+        ///     Whether the measurement probabilities will be displayed as an
+        ///     interactive histogram.
+        /// </summary>
+        private bool ShowMeasurementDisplayHistogram { get; set; } = false;
 
         /// <summary>
         ///     Sets the properties of this display dumper from a given
@@ -54,6 +59,7 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         public JupyterDisplayDumper Configure(IConfigurationSource configurationSource)
         {
             configurationSource
+                .ApplyConfiguration<bool>("dump.measurementDisplayHistogram", value => ShowMeasurementDisplayHistogram = value)
                 .ApplyConfiguration<bool>("dump.truncateSmallAmplitudes", value => TruncateSmallAmplitudes = value)
                 .ApplyConfiguration<double>("dump.truncationThreshold", value => TruncationThreshold = value)
                 .ApplyConfiguration<BasisStateLabelingConvention>(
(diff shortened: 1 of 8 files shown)


src/Jupyter/Visualization/StateDisplayOperations.cs
```diff
@@ -46,6 +46,11 @@ public class JupyterDisplayDumper : QuantumSimulator.StateDumper
         /// </summary>
         public BasisStateLabelingConvention BasisStateLabelingConvention { get; set; } = BasisStateLabelingConvention.Bitstring;
 
+        /// <summary>
+        ///     Whether the measurement probabilities will be displayed as an
+        ///     interactive histogram.
+        /// </summary>
+        private bool ShowMeasurementDisplayHistogram { get; set; } = false;
```
cgranade, line 53: Does this imply that users need to explicitly opt-in with `%config` to get the new functionality? If so, where should we document that? We should likely also adapt samples as part of this to set that `%config` option.

cgranade (approved):

src/Jupyter/Visualization/StateDisplayOperations.cs
rmshaffer: Well, this particular `bool` doesn't actually have that effect. This one simply gets set inside `Configure()` below with whatever the value of `dump.measurementDisplayHistogram` is. So the initial `false` value here does not impact the behavior.

The current default of `dump.measurementDisplayHistogram` is indeed `false`, though, as documented at https://docs.microsoft.com/qsharp/api/iqsharp-magic/config. As you point out, I believe we should flip this to `true` when ready (i.e. once the visualization features have been more fully finished). For the last QDK release I kept it at `false` in order to minimize potential impact on the release until we had done further testing.

cgranade: > Well, this particular `bool` doesn't actually have that effect. This one simply gets set inside `Configure()` below with whatever the value of `dump.measurementDisplayHistogram` is. So the initial `false` value here does not impact the behavior.

Ah, my apologies for misunderstanding!

> The current default of `dump.measurementDisplayHistogram` is indeed `false`, though, as documented at https://docs.microsoft.com/qsharp/api/iqsharp-magic/config. As you point out, I believe we should flip this to `true` when ready (i.e. once the visualization features have been more fully finished). For the last QDK release I kept it at `false` in order to minimize potential impact on the release until we had done further testing.

Makes perfect sense, thanks for the context!

PR #360, opened 2020-11-02
Adapting for API change in qsharp-runtime/#370

bettinaheim:


cgranade (approved):

rmshaffer (dismissed):
Looks good, thanks! In order for the CI to pass (and to avoid breaking the E2E QDK build), please also update IQ# to use a `beta` QDK build which includes the necessary `qsharp-runtime` changes.

build/ci.yml
```diff
@@ -10,7 +10,7 @@ pr:
 
 variables:
   Build.Major: 0
-  Build.Minor: 13
+  Build.Minor: 12
```
rmshaffer, line 13: Was this intentional?

images/iqsharp-base/Dockerfile
```diff
@@ -109,7 +109,7 @@ ENV PATH=$PATH:${HOME}/dotnet:${HOME}/.dotnet/tools \
 # Install IQ# and the project templates, using the NuGet packages from the
 # build context.
 ARG IQSHARP_VERSION
-RUN dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.13.20102604" && \
+RUN dotnet new -i "Microsoft.Quantum.ProjectTemplates::0.12.20101601-beta" && \
```
rmshaffer, line 112: Was this intentional?

build/ci.yml
bettinaheim: No, it was not, thanks! Should I change it to 14 while I am at it, since we will have breaking changes with the next release?

images/iqsharp-base/Dockerfile
bettinaheim: No, thanks!

build/ci.yml
rmshaffer: I think keeping it in sync with the official builds from the `qdk` repo is most desirable, so we could probably wait on that (unless that is also imminent).

rmshaffer (approved):

rmshaffer (approved):

PR #367, opened 2020-11-14
Update Azure Quantum data plane URL for multi-region

rmshaffer:
This change updates the base URL used by IQ# to connect to the Azure Quantum data plane.

It also adds a `location` parameter to the `%azure.connect` magic command (and therefore also the `qsharp.azure.connect()` Python API) allowing the user to specify the region in which their Azure Quantum workspace is provisioned.

### DIFF
--- a/src/AzureClient/AzureEnvironment.cs
+++ b/src/AzureClient/AzureEnvironment.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         private string ClientId { get; set; } = string.Empty;
         private string Authority { get; set; } = string.Empty;
         private List<string> Scopes { get; set; } = new List<string>();
-        private Uri? BaseUri { get; set; }
+        private Func<string?, Uri> BaseUriForLocation { get; set; } = (location) => new Uri($"https://{location}.quantum.azure.com/");
 
         private AzureEnvironment()
         {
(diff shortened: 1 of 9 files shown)


src/AzureClient/AzureEnvironment.cs
```diff
@@ -58,17 +58,39 @@ public static AzureEnvironment Create(string subscriptionId)
             return Production(subscriptionId);
         }
 
+        private string GetNormalizedLocation(string location)
+        {
+            if (Type == AzureEnvironmentType.Canary)
+            {
+                return "eastus2euap";
+            }
```
ricardo-espinoza, line 66: Instead of adding a hard coded condition here to check for for Canary, should we instead by passing a different BaseUriForLocation function when constructing Canary below (L224)?

ricardo-espinoza (approved):

src/AzureClient/AzureEnvironment.cs
```diff
@@ -58,17 +58,39 @@ public static AzureEnvironment Create(string subscriptionId)
...
+
+            // Convert user-provided location into names recognized by Azure resource manager.
+            // For example, a customer-provided value of "West US" should be converted to "westus".
+            var normalizedLocation = location.ToLowerInvariant().Replace(" ", "");
+            if (UriHostNameType.Unknown == Uri.CheckHostName(normalizedLocation))
+            {
+                // If provided location is invalid, "westus" is used.
+                normalizedLocation = "westus";
```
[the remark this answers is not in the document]
rmshaffer: Yes, I agree, that'd be very helpful.

src/AzureClient/AzureEnvironment.cs
rmshaffer: Yes, I think that's a better way to do this.

vxfield (approved):

PR #368, opened 2020-11-16
Updates to compilation serialization

rmshaffer:
Updates to IQ# compilation serialization to consume the changes from https://github.com/microsoft/qsharp-compiler/pull/721.

### DIFF
--- a/src/Core/Compiler/CompilerService.cs
+++ b/src/Core/Compiler/CompilerService.cs
@@ -249,7 +249,11 @@ namespace Microsoft.Quantum.IQSharp
                 var syntaxTree = new QsCompilation(fromSources.ToImmutableArray(), qsCompilation.EntryPoints);
 
                 using var serializedCompilation = new MemoryStream();
-                QsCompiler.BondSchemas.Protocols.SerializeQsCompilationToSimpleBinary(syntaxTree, serializedCompilation);
+                if (!CompilationLoader.WriteBinary(syntaxTree, serializedCompilation))
+                {
+                    logger.LogError("IQS005", "Failed to write compilation to binary stream.");
+                    return null;
+                }
 
                 var resourceDescription = new ResourceDescription
                 (
--- a/src/Core/Compiler/CompilerService.cs
+++ b/src/Core/Compiler/CompilerService.cs
@@ -245,52 +245,52 @@ namespace Microsoft.Quantum.IQSharp
                     options);
 
                 // Generate the assembly from the C# compilation:
-                using (var ms = new MemoryStream())
+                var fromSources = qsCompilation.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.EndsWith(".qs")));
+                var syntaxTree = new QsCompilation(fromSources.ToImmutableArray(), qsCompilation.EntryPoints);
+
+                using var serializedCompilation = new MemoryStream();
+                QsCompiler.BondSchemas.Protocols.SerializeQsCompilationToSimpleBinary(syntaxTree, serializedCompilation);
+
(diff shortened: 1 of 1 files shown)


src/Core/Compiler/CompilerService.cs
```diff
@@ -248,54 +245,52 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
...
 
                 // Generate the assembly from the C# compilation:
-                using (var ms = new MemoryStream())
-                using (var bsonStream = new MemoryStream())
-                {
-                    using var writer = new BsonDataWriter(bsonStream) { CloseOutput = false };
-                    var fromSources = qsCompilation.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.EndsWith(".qs")));
-                    Json.Serializer.Serialize(writer, new QsCompilation(fromSources.ToImmutableArray(), qsCompilation.EntryPoints));
+                var fromSources = qsCompilation.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.EndsWith(".qs")));
+                var syntaxTree = new QsCompilation(fromSources.ToImmutableArray(), qsCompilation.EntryPoints);
```
rmshaffer, line 249: Apologies for the ugly-looking diff here. It's mostly whitespace/indentation changes, since I moved to inline `using` statements. Choosing "Hide whitespace changes" should make it easier to review.

src/Core/Compiler/CompilerService.cs
```diff
@@ -248,54 +245,52 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
...
 
-                    var resourceDescription = new ResourceDescription
-                    (
-                        resourceName: QsCompiler.ReservedKeywords.DotnetCoreDll.ResourceName,
-                        dataProvider: () => new MemoryStream(bsonStream.ToArray()), 
-                        isPublic: true
-                    );
+                using var serializedCompilation = new MemoryStream();
+                QsCompiler.BondSchemas.Protocols.SerializeQsCompilationToSimpleBinary(syntaxTree, serializedCompilation);
```
cesarzc, line 252: `CompilationLoader` has a public API that handles the serialization of the syntax tree into a Stream:
`public static bool WriteBinary(QsCompilation syntaxTree, Stream stream)`

It would be great if it could be used here since it ensures that the protocol used is in sync with the version of the compiler that is referenced.

rmshaffer: I definitely agree in principle. The only issue I see is that a few lines below, we have to explicitly specify `DotnetCoreDll.ResourceNameQsDataBondV1` when creating the `ResourceDescription`. So if we switch to using `CompilationLoader.WriteBinary()` here, those could potentially get out-of-sync, right? Or is there some other way around this?

cesarzc: Currently I don't think there's other way around it. Something I can think of is to add a new property to `DotnetCoreDll` which always provides the current name of the serialized syntax tree resource. That way, this part of the code would always be in sync even if the underlying protocols change.

I would be in favor of using `WriteBinary()` now, and later adding the `CurrentSyntaxTreeResourceName` to the compiler and change it here. Since the plan is not to change the protocol, I don't think there's going to be a problem if `ResourceNameQsDataBondV1` is used for this PR.

cesarzc: I'll start working on a PR on the compiler side.

rmshaffer: Makes sense! Updating this now to use `WriteBinary()`.

cesarzc (approved):
Thanks for making this change Ryan!

PR #369, opened 2020-11-17
GDPR Cookie Compliance update

vxfield:
This PR updates our telemetry code to use the latest Microsoft Cookie Compliance API to satisfy latest GDPR requirements.

### DIFF
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -228,10 +228,10 @@ class Kernel {
             || location.hostname == "127.0.0.1"
             || this.hostingEnvironment == null
             || this.hostingEnvironment == "";
-        const enableClientTelemetry =
-            this.hostingEnvironment == "ENABLE_CLIENT_TELEMETRY";
+        const forceEnableClientTelemetry =
+            this.hostingEnvironment == "FORCE_ENABLE_CLIENT_TELEMETRY";
 
-        if (!enableClientTelemetry && isLocalEnvironment) {
+        if (!forceEnableClientTelemetry && isLocalEnvironment) {
             console.log("Client telemetry disabled on local environment");
             return;
         }
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -223,13 +223,15 @@ class Kernel {
             return;
         }
 
-        var isLocalEnvironment =
+        const isLocalEnvironment =
             location.hostname == "localhost"
             || location.hostname == "127.0.0.1"
             || this.hostingEnvironment == null
             || this.hostingEnvironment == "";
+        const enableClientTelemetry =
+            this.hostingEnvironment == "ENABLE_CLIENT_TELEMETRY";
 
-        if (isLocalEnvironment) {
+        if (!enableClientTelemetry && isLocalEnvironment) {
             console.log("Client telemetry disabled on local environment");
             return;
         }
--- a/src/Kernel/client/telemetry.ts
+++ b/src/Kernel/client/telemetry.ts
@@ -4,6 +4,8 @@
 //[SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="This function access is supposed to be public.")]
 const CLIENT_INFO_API_URL = 'https://iqsharp-telemetry.azurewebsites.net/api/GetClientInfo?code=1kIsLwHdwLlH9n5LflhgVlafZaTH122yPK/3xezIjCQqC87MJrkdyQ==';
 
+const COOKIE_CONSENT_JS = 'https://wcpstatic.microsoft.com/mscc/lib/v2/wcp-consent.js';
+
 /*
 There are npm packages for this but I preferred to keep it small
 and simple with no additional dependencies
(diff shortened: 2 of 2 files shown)


src/Kernel/client/telemetry.ts
```diff
@@ -148,18 +152,19 @@ class TelemetryHelper {
...
         console.log("Getting ClientInfo");
-        var url =
+        const url =
             CLIENT_INFO_API_URL
             + (setHasConsent ? "&hasconsent=1" : "")
-            + ((this.origin != null && this.origin != "") ? "&origin=" + this.origin : "");
+            + ((this.origin !== null && this.origin !== "") ? "&origin=" + this.origin : "");
         try {
-            var clientInfo = await this.fetchExt<ClientInfo>(url);
+            const clientInfo = await this.fetchExt<ClientInfo>(url);
```
rmshaffer, line 167: From what I can tell, this GetClientInfo API returns `HasConsent: true` even if I have no cookies set locally. I presume that is because my request is coming from the US? What's the best way to test this locally... is there some URL parameter I can add to the request to override the country code?

src/Kernel/client/kernel.ts
```diff
@@ -223,13 +223,15 @@ class Kernel {
             return;
         }
 
-        var isLocalEnvironment =
+        const isLocalEnvironment =
             location.hostname == "localhost"
             || location.hostname == "127.0.0.1"
             || this.hostingEnvironment == null
             || this.hostingEnvironment == "";
+        const enableClientTelemetry =
```
rmshaffer, line 231: [nit] variable naming - I'd suggest something like `forceEnableClientTelemetry`, since this is not the only way for client telemetry to be enabled (i.e., a value of `false` does not necessarily mean that client telemetry is disabled).

src/Kernel/client/telemetry.ts
```diff
@@ -34,94 +36,97 @@ class LiteEvent<T> implements ILiteEvent<T> {
...
+        const userLanguage = navigator.language;
+        WcpConsent && WcpConsent.init(userLanguage, div.id, (err, siteConsent) => {
+            if (!err) {
+                this.siteConsent = siteConsent;
+                console.log("WcpConsent initialized.");
+                this.checkRequiredConsent();
+            } else {
+                console.log("Error initializing WcpConsent: " + err);
+            }
+        }, this.onConsentChanged, Themes.light);
```
rmshaffer, line 129: Should we detect if the user is in high-contrast mode and, if so, pass the appropriate theme here for accessibility? This may be outside the scope of this change, but it would be good to create an issue so we don't forget about it.

src/Kernel/client/telemetry.ts
```diff
@@ -4,6 +4,8 @@
 //[SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="This function access is supposed to be public.")]
 const CLIENT_INFO_API_URL = 'https://iqsharp-telemetry.azurewebsites.net/api/GetClientInfo?code=1kIsLwHdwLlH9n5LflhgVlafZaTH122yPK/3xezIjCQqC87MJrkdyQ==';
 
+const COOKIE_CONSENT_JS = 'https://wcpstatic.microsoft.com/mscc/lib/v2/wcp-consent.js';
```
rmshaffer, line 7: Does this JS file have documentation anywhere? Could we add a link to it in a comment here?

rmshaffer (commented):
Are there any tests that exercise the new TypeScript code? At least unit tests, although ideally we could also add a UI test via Selenium. This is a critical feature and it would be rarely exercised in day-to-day use, which means that regressions might go unnoticed if there are no tests.

src/Kernel/client/telemetry.ts
vxfield: This is kind of hard to test in an automated way. The Microsoft cookie API has all the logic in it and it about when it shows or not the cookie banner and response if the user has consented, and it could change in the future becoming more or less conservative depending on the user's country (via reverse IP) and locale (browser's primary language).

The way I tested it was to set the "ENABLE_CLIENT_TELEMETRY" and use a VPN to emulate traffic coming from Belgium, for example.

src/Kernel/client/kernel.ts
vxfield: Makes sense. I renamed it.

src/Kernel/client/telemetry.ts
vxfield: The theme parameter is a new cookie API parameter to have the banner to better blend in the webpage look & feel. For that we would need to understand/support Jupyter themes work and that's out of scope.
From the accessibility standpoint, the banner will display correctly, matching the OS high-contrast settings (I just tested on Mac and Windows).

src/Kernel/client/telemetry.ts
vxfield: This is a Microsoft internal library and there is no public documentation to link  to.

src/Kernel/client/telemetry.ts
cgranade, line 129: This could work poorly if the user has dark mode set in Windows 10 or Android preferences. Using a [CSS media query](https://usefulangle.com/post/318/javascript-check-dark-mode-activated) should allow for detecting and conditioning on whether the dark-mode preference is set.

```suggestion
        }, this.onConsentChanged, window.matchMedia('prefers-color-scheme: dark').matches ? Themes.dark : Themes.light);
```

vxfield:
> Are there any tests that exercise the new TypeScript code? At least unit tests, although ideally we could also add a UI test via Selenium. This is a critical feature and it would be rarely exercised in day-to-day use, which means that regressions might go unnoticed if there are no tests.

Unfortunately there is no easy way to test this code as the logic that decides whether or not to show the banner and the banner contents rest in an internal Microsoft Cookie API library that is a black-box for us.
That library is current using the client's IP to do a country lookup and decide when to show the banner, but that could change in the future.

cgranade (commented):
As a side note, since some of the changes made in this PR are stylistic / linting issues, would it be good at this point to enable tslint in CI?

src/Kernel/client/telemetry.ts
rmshaffer: Agreed, we wouldn't want automated tests that actually rely on the behavior of the cookie API.

One idea; Could we have an override here (e.g., maybe via another special `hostingEnvironment` value, or perhaps something else similar) that skips the call to the service and always uses some hard-coded `ClientInfo` object which has `HasConsent: false`?

After our discussion, I understand that we can't easily get automated tests to actually show the banner, since that code is outside our control. But it could be useful to at least allow us to exercise our own portion of the code.

src/Kernel/client/telemetry.ts
rmshaffer: Makes sense. Could you add a comment here indicating that?

src/Kernel/client/telemetry.ts
```diff
@@ -34,94 +36,97 @@ class LiteEvent<T> implements ILiteEvent<T> {
...
-    CookieConsentMarkup: ConsentMarkup;
     HasConsent: boolean;
     FirstOrigin: string;
 }
 
-interface ConsentMarkup {
-    Markup: string;
-    Javascripts: string[];
-    Stylesheets: string[];
+enum consentCategories {
```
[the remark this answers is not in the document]
vxfield: Done.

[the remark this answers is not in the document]
vxfield: I reverted it to enum because I really want to use it as an enum in 
this call `this.siteConsent.getConsentFor(consentCategories.Required);` instead of a string literal.

src/Kernel/client/telemetry.ts
rmshaffer: As discussed, unit tests can be added in a separate PR, since this is only a modification to pre-existing code. It would be great if you could create an issue to track adding unit tests for this code.

rmshaffer (approved):

src/Kernel/client/telemetry.ts
vxfield: Created issue #370 to add unit tests for this and other typescript code.

vxfield:
> stylistic / linting

Created issue #372 for this.

PR #382, opened 2020-12-09
Skip serialization for non-targeted assemblies

rmshaffer:
PR #368 updated IQ# to use the new compiler serialization format for Q# assemblies. There is some perf cost associated with loading the serializers/deserializers that we would like to avoid. Fortunately, it looks like the serialization is only necessary for targeted assemblies (i.e., assemblies being built for submission to hardware targets). The bulk of IQ# scenarios do not compile to a hardware target, and so we can get performance improvement by skipping serialization in these cases.

This PR skips the Q# compilation serialization when the execution target is not set. It also sets `LoadReferencesBasedOnGeneratedCsharp=true` in the `CompilationLoader` configuration, which should allow the compiler to make a similar optimization.

Keeping this PR as a draft until verification is complete.

### DIFF
--- a/src/Core/Compiler/CompilerService.cs
+++ b/src/Core/Compiler/CompilerService.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Quantum.IQSharp
             var loadOptions = new CompilationLoader.Configuration
             {
                 GenerateFunctorSupport = true,
-                LoadReferencesBasedOnGeneratedCsharp = string.IsNullOrEmpty(executionTarget), // serialization is unnecessary if no execution target
+                LoadReferencesBasedOnGeneratedCsharp = string.IsNullOrEmpty(executionTarget), // deserialization of resources in references is only needed if there is an execution target
                 IsExecutable = compileAsExecutable,
                 AssemblyConstants = new Dictionary<string, string> { [AssemblyConstants.ProcessorArchitecture] = executionTarget ?? string.Empty },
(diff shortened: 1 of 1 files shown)


src/Core/Compiler/CompilerService.cs
```diff
@@ -132,6 +132,7 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
             var loadOptions = new CompilationLoader.Configuration
             {
                 GenerateFunctorSupport = true,
+                LoadReferencesBasedOnGeneratedCsharp = true,
```
cesarzc, line 135: I think this should depend on whether compilation is done for an execution target or not.

rmshaffer: Yes, good point, that makes sense to me as well.

src/Core/Compiler/CompilerService.cs
```diff
@@ -132,6 +132,7 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
             var loadOptions = new CompilationLoader.Configuration
             {
                 GenerateFunctorSupport = true,
+                LoadReferencesBasedOnGeneratedCsharp = string.IsNullOrEmpty(executionTarget), // serialization is unnecessary if no execution target
```
cesarzc, line 135:
```suggestion
                LoadReferencesBasedOnGeneratedCsharp = string.IsNullOrEmpty(executionTarget), // deserialization of resources in references is only needed if there is an execution target
```

cesarzc (approved):

PR #385, opened 2020-12-23
Remove assertion that compilation had no warnings in engine tests

samarsha:
microsoft/qsharp-compiler#791 adds deprecation warnings when parentheses are used for `for`, `using` and `borrowing` statements. IQ# has test snippets that trigger these warnings, for example:

https://github.com/microsoft/iqsharp/blob/afeb47a8b2957ba72e150a542f25d5d1aaedadad/src/Tests/SNIPPETS.cs#L319-L321

The IQ# tests also assert that no warnings were generated, which caused the end-to-end build to fail. Since the warnings were designed to not introduce any breaking changes yet, it would be nice if IQ# tests didn't rely on there being no warnings.

The other option is to update the tests to remove the deprecated parentheses and avoid the warnings - let me know which option seems better.

Cc: @bettinaheim

### DIFF
--- a/src/Tests/IQsharpEngineTests.cs
+++ b/src/Tests/IQsharpEngineTests.cs
@@ -52,7 +52,6 @@ namespace Tests.IQSharp
             var response = await engine.ExecuteMundane(source, channel);
             PrintResult(response, channel);
             Assert.AreEqual(ExecuteStatus.Ok, response.Status);
-            Assert.AreEqual(0, channel.msgs.Count);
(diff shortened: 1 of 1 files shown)


rmshaffer (approved):

samarsha:
@rmshaffer It looks like this PR and #386 are blocked by a component governance alert for `y18n` 4.0.0.

cgranade (approved):

cgranade:
With the alert for `y18n` being resolved by #388, I think this should now be ready to merge in. Thanks for fixing this, @samarsha!

PR #389, opened 2021-01-14
Start engine services in Start() instead of constructor.

cgranade:
This PR delays starting services required by `IQSharpEngine` until the `Engine.Start` method is called, so as to not block proper handling of shell messages for reporting startup status to the client. This PR also follows the advice of @cesarzc, and initializes Bond protocol serialization and deserialization in the engine, so that the Q# compilation loader is ready earlier.

(Marked as draft until a dependency on https://github.com/microsoft/jupyter-core/pull/71 can be taken.)

### DIFF
--- a/src/Kernel/IQSharpEngine.cs
+++ b/src/Kernel/IQSharpEngine.cs
@@ -78,6 +78,10 @@ namespace Microsoft.Quantum.IQSharp.Kernel
             // and deserializers. Since that runs in the background, starting
             // the engine should not be blocked, and other services can
             // continue to initialize while the Q# compilation loader works.
+            //
+            // For more details, see:
+            //     https://github.com/microsoft/qsharp-compiler/pull/727
+            //     https://github.com/microsoft/qsharp-compiler/pull/810
             logger.LogDebug("Loading serialization and deserialziation protocols.");
             Protocols.Initialize();
 
--- a/src/Kernel/IQSharpEngine.cs
+++ b/src/Kernel/IQSharpEngine.cs
@@ -44,6 +44,9 @@ namespace Microsoft.Quantum.IQSharp.Kernel
 
         internal IWorkspace? Workspace { get; private set; } = null;
 
+        private TaskCompletionSource<bool> initializedSource = new TaskCompletionSource<bool>();
+        private Task initialized => initializedSource.Task;
+
         /// <summary>
         /// The main constructor. It expects an `ISnippets` instance that takes care
         /// of compiling and keeping track of the code Snippets provided by users.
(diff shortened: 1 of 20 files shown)


cgranade:
This turned into a lot more of a nullability improvement PR than I intended, as adapting a test wound up involving chasing down everywhere that uninitialized services could result in null reference exceptions. Apologies for the noisy PR as a result of that.

src/Kernel/IQSharpEngine.cs
```diff
@@ -41,20 +58,38 @@ public class IQSharpEngine : BaseEngine
...
+        /// <inheritdoc />
+        public override void Start()
+        {
+            base.Start();
 
+            // Before anything else, make sure to start the right background
+            // thread on the Q# compilation loader to initialize serializers
+            // and deserializers. Since that runs in the background, starting
+            // the engine should not be blocked, and other services can
+            // continue to initialize while the Q# compilation loader works.
```
guenp, line 80: If there is a GH issue, perhaps link it here to explain what problem this is solving

cgranade: I don't think there's an issue that was created for this already, so much as that @cesarzc and I noticed this while walking through the code yesterday to understand the initialization performance better. Perhaps linking to the PR that introduced the serializer initialization methods could be useful, though? Let me go on and do that, then.

cesarzc: Here's the PR that introduced serializer/deserializer initialization methos: https://github.com/microsoft/qsharp-compiler/pull/727

src/Kernel/Magic/Resolution/MagicResolver.cs
```diff
@@ -102,59 +102,76 @@ public MagicSymbol Resolve(string symbolName)
...
-        /// <summary>
-        /// Finds the MagicSymbols inside an assembly, and returns an instance of each.
-        /// </summary>
+        /// <inheritdoc />
         public IEnumerable<MagicSymbol> FindMagic(AssemblyInfo assm)
         {
             var result = new MagicSymbol[0];
 
-            if (cache.TryGetValue(assm, out result))
+            lock (cache)
```
cesarzc, line 110: Before adding this `lock` statement, was this part not thread-safe?

cgranade: I think that's correct, yes. That said, it did not used to be called from background threads. This PR changes so that the engine starts searching for magic symbols in the background once started, rather than waiting for the first cell before searching.

src/Kernel/IQSharpEngine.cs
guenp: Thanks both!

guenp (approved):

cgranade:
It looks like something else is going wrong with CI, such that tests are hanging without an explicit failure. I'm unable to reproduce locally yet; it seems like this only happens in CI, but I will keep investigating.

cgranade:
In investigating recent CI failures, it appears that https://github.com/microsoft/jupyter-core/pull/71 may have introduced a subtle Jupyter messaging protocol bug. With 566b5b9, I'm currently testing a fix in jupyter-core; if it works, will open a PR for that fix shortly and will push a new release.

cgranade:
It looks like this resolved the issue, so I will open a PR on jupyter-core, and then will update this PR once that is resolved.

guenp (approved):

guenp (approved):

PR #392, opened 2021-01-14
Updating test case to recent change in qsharp-runtime

msoeken:
This PR fixes a failing test case due to microsoft/qsharp-runtime#458.

### DIFF
--- a/src/Tests/ExecutionPathTracerTests.cs
+++ b/src/Tests/ExecutionPathTracerTests.cs
@@ -67,11 +67,12 @@ namespace Tests.IQSharp
         }
 
         // Helper functions for tests
-        public Operation ControlledX(int[] controlId, int targetId) =>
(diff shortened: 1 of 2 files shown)


src/Tests/ExecutionPathTracerTests.cs
```diff
@@ -67,11 +67,12 @@ private void PruneNonDeterministicGates(IEnumerable<Operation>? operations)
         }
 
         // Helper functions for tests
-        public Operation ControlledX(int[] controlId, int targetId) =>
+        public Operation ControlledX(int[] controlId, int targetId, bool isAdjoint = false) =>
```
cgranade, line 70: Thanks for fixing this! I'm slightly surprised that the execution path tracer is sensitive to the `Controlled X` operation being adjointed, though; seems like a reasonable thing to fix going forward after this is merged in, perhaps?

guenp (approved):

cgranade (approved):

PR #395, opened 2021-01-25
Create Magic command azure.quotas in IQSharp

ricardo-espinoza:
This change introduces a new magic command: `%azure.quotas` to IQSharp. This pull request is dependent on the following PR, since it exposes the required classes and interface methods:
https://github.com/microsoft/qsharp-runtime/pull/480


src/AzureClient/AzureClient.cs
```diff
@@ -502,6 +502,30 @@ public async Task<ExecutionResult> GetJobListAsync(IChannel channel, string filt
...
+            if (connectionResult.Status != ExecuteStatus.Ok)
+            {
+                return connectionResult;
+            }
+
+            var quotas = await ActiveWorkspace.ListQuotasAsync() ?? new List<QuotaInfo>();
+
+            if (quotas.Count() == 0)
+            {
+                channel.Stderr("No quota information found in current Azure Quantum workspace.");
```
cgranade, line 523: This may not be an error condition, as it correctly reports that there are no quotas to list?

src/AzureClient/AzureClient.cs
```diff
@@ -502,6 +502,30 @@ public async Task<ExecutionResult> GetJobListAsync(IChannel channel, string filt
             return jobs.ToExecutionResult();
         }
 
+        public async Task<ExecutionResult> GetQuotaListAsync(IChannel channel)
+        {
+            if (ActiveWorkspace == null)
+            {
+                channel.Stderr($"Please call {GetCommandDisplayName("connect")} before reading quota information.");
+                return AzureClientError.NotConnected.ToExecutionResult();
```
cgranade, line 510: IIRC, you'll want to pass `ExecuteStatus.Error` as an optional parameter to `ToExecutionResult` to mark this as an error.

src/AzureClient/Magic/QuotaMagic.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        ///     Lists all jobs in the active workspace, optionally filtered by a provided parameter.
+        /// </summary>
+        public override async Task<ExecutionResult> RunAsync(string input, IChannel channel, CancellationToken cancellationToken)
+        {
+            return await AzureClient.GetQuotaListAsync(channel);
+        }
+    }
+}
```
cgranade, line 65:
```suggestion
}

```

cgranade (approved):

src/AzureClient/AzureClient.cs
ricardo-espinoza: For this one, I'm following the pattern of the other methods in the client, and they appear to be leaving the default when converting errors to execution results. Keeping as is for consistency for now, but we can chat offline about this.

src/AzureClient/AzureClient.cs
ricardo-espinoza: You're right. Thanks for the catch!

cgranade (approved):

ricardo-espinoza:
This pull request is being force merged since it is interdependent with [qsharp-runtime:480](https://github.com/microsoft/qsharp-runtime/pull/480).

PR #399, opened 2021-01-27
Fix incorrect construction of magic command azure.quotas

ricardo-espinoza:
Quotas command was constructed using an incorrect name.

Updating to the intended name.


cgranade (approved):
It looks like the API documentation comment at L17 may also be good to update, but otherwise LGTM.

ricardo-espinoza:
> It looks like the API documentation comment at L17 may also be good to update, but otherwise LGTM.

Thanks for the catch, @cgranade . To avoid delaying the pending PR we have tonight, I'll leave as is, but I'll correct the comment in the next few days.

cgranade:
Force merging to restore build.

PR #411, opened 2021-02-17
Add horizontal scroll for large debug visuals

anjbur:
Add horizontal scroll bar for large visualizations.

### DIFF
--- a/src/Kernel/client/kernel.ts
+++ b/src/Kernel/client/kernel.ts
@@ -65,7 +65,16 @@ class Kernel {
                 if (stateDivId != null) {
                     let stateDiv = document.getElementById(stateDivId);
                     if (stateDiv != null) {
-                        let { chart: chart } = createNewCanvas(stateDiv);
+                        // Create necessary elements so a large chart will scroll
+                        let chartWrapperDiv = document.createElement("div");
+                        chartWrapperDiv.style.overflowX = "scroll";
+                        let innerChartWrapperDiv = document.createElement("div");
+                        innerChartWrapperDiv.style.height = "350px";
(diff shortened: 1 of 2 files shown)


src/Kernel/client/kernel.ts
```diff
@@ -65,7 +65,16 @@ class Kernel {
                 if (stateDivId != null) {
                     let stateDiv = document.getElementById(stateDivId);
                     if (stateDiv != null) {
-                        let { chart: chart } = createNewCanvas(stateDiv);
+                        // Create necessary elements so a large chart will scroll
+                        let chartWrapperDiv = document.createElement("div");
+                        chartWrapperDiv.style.overflowX = "scroll";
+                        let innerChartWrapperDiv = document.createElement("div");
+                        innerChartWrapperDiv.style.height = "350px";
+                        innerChartWrapperDiv.id = "innerChartWrapper";
```
cgranade, line 73: There may in general be more than one such `<div>` element, such that assigning the same ID to each could cause problems later. Would it be possible to either pass the `HTMLElement` object directly to `fitChart`, or to generate a unique ID for each chart wrapper div?

cgranade (approved):
Just had the one question about uniqueness of div ids, but otherwise LGTM. Approving so as to leave to your best judgment about how to resolve div uniqueness.

src/Kernel/client/kernel.ts
anjbur: ah good point thank you. Fixed it so it refers to this element through `chart.canvas.parentElement` instead

cgranade (approved):

PR #437, opened 2021-05-19
Target both Net5.0 and 3.1 for core compiler and runtime assemblies.

ricardo-espinoza:
Cross-repo change to add limited support for users that only have .Net5 installed.


cgranade:
@ricardo-espinoza: Is this PR still needed given .NET 6, or should we bring the changes into a different feature branch? Thanks!

ricardo-espinoza:
This work has been superseded by a different plan to move into .NET 6 instead later this year.

PR #442, opened 2021-05-29
Expose trace as a command of a Q# operation

anpaz:
Working on migrating our execution trace visualizer to its own repo so it can be a stand-alone tool.
These changes expose the response of the `%trace` magic as a `trace` method on a Q# operation. This makes it trivial to get the trace output needed by the tool:

```python
import qsharp
from Teleport import TeleportCircuit
result = TeleportCircuit.trace()
print(json.dumps(result))
```

### DIFF
--- a/src/Python/qsharp-core/qsharp/clients/iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/clients/iqsharp.py
@@ -178,6 +178,9 @@ class IQSharpClient(object):
             counts[row["Metric"]] = int(row["Sum"])
         return counts
 
+    def trace(self, op, **kwargs) -> Any:
+        return self._execute_callable_magic('trace', op, _quiet_ = True, **kwargs)
+
     def component_versions(self, **kwargs) -> Dict[str, LooseVersion]:
         """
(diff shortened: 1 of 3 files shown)


src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -242,9 +245,11 @@ def _execute(self, input, return_full_result=False, raise_on_stderr : bool = Fal
 
         def log_error(msg):
             errors.append(msg)
-
+        
         handlers = {
-            'execute_result': (lambda msg: results.append(msg))
+            'execute_result': (lambda msg: results.append(msg)),
+            'render_execution_path':  (lambda msg: results.append(msg)),
+            'display_data': lambda msg: ...
```
cgranade, line 252: Is the `...` here intended?

anpaz: Yes, it's the equivalent of a no-op. If I don't include it the default handler of display_data is to print a message on the screen.

anjbur (approved):
Looks good to me!

PR #447, opened 2021-06-04
Using quantum-viz.js npm module.

anpaz:
The tracer has now been forked into its own OSS project, `quantum-viz.js`. This removes the local copy and uses the corresponding npm package.


build/test.ps1
```diff
@@ -125,8 +125,6 @@ if ($Env:ENABLE_PYTHON -eq "false") {
     Test-Python '../src/Python/qsharp-core' '../src/Python/qsharp-core/qsharp/tests'
 }
 
-Test-JavaScript '../src/Kernel'
```
anjbur, line 128: Should the function definition be removed as well as the call here? It looks like this was the only place `Test-JavaScript` was used.

anjbur (approved):

build/test.ps1
anpaz: We could, but I left it in case we want to use it in the future.

PR #449, opened 2021-06-04
Start experimental simulators feature

cgranade:
This PR contains the IQ# components needed to support https://github.com/microsoft/qsharp-runtime/pull/709. Please see that PR for full details.

### DIFF
--- a/src/Kernel/Magic/SimulateNoise.cs
+++ b/src/Kernel/Magic/SimulateNoise.cs
@@ -84,6 +84,7 @@ namespace Microsoft.Quantum.Experimental
             if (engine is IQSharpEngine iQSharpEngine)
             {
                 iQSharpEngine.RegisterDisplayEncoder(new MixedStateToHtmlDisplayEncoder());
+                iQSharpEngine.RegisterDisplayEncoder(new StabilizerStateToHtmlDisplayEncoder(configurationSource));
             }
         }
 
@@ -116,8 +117,8 @@ namespace Microsoft.Quantum.Experimental
             if (symbol == null) throw new InvalidOperationException($"Invalid operation name: {name}");
 
             var qsim = new OpenSystemsSimulator(
-                ConfigurationSource.OpenSystemsSimulatorCapacity,
-                ConfigurationSource.OpenSystemsSimulatorRepresentation
+                ConfigurationSource.ExperimentalSimulatorCapacity,
+                ConfigurationSource.ExperimentalSimulatorRepresentation
             );
             if (NoiseModelSource.NoiseModel != null)
             {
--- a/src/Kernel/Magic/SimulateNoise.cs
+++ b/src/Kernel/Magic/SimulateNoise.cs
@@ -43,6 +43,7 @@ namespace Microsoft.Quantum.Experimental
 
                     #### See also
 
+                    - [`%config`](https://docs.microsoft.com/qsharp/api/iqsharp-magic/config)
                     - [`%experimental.noise_model`](https://docs.microsoft.com/qsharp/api/iqsharp-magic/experimental.noise_model)
 
                     #### Required parameters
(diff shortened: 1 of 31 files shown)


src/Kernel/Magic/SimulateNoise.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var inputParameters = ParseInputParameters(input, firstParameterInferredName: ParameterNameOperationName);
+
+            var name = inputParameters.DecodeParameter<string>(ParameterNameOperationName);
+            var symbol = SymbolResolver.Resolve(name) as dynamic; // FIXME: Should be IQSharpSymbol.
+            if (symbol == null) throw new InvalidOperationException($"Invalid operation name: {name}");
+
+            var qsim = new OpenSystemsSimulator(
+                ConfigurationSource.OpenSystemsSimulatorCapacity //,
+                // ConfigurationSource.OpenSystemsSimulatorRepresentation
```
swernli, line 120: Is this functionality that is not ready yet or a hold-over from a previous design?

src/Tests/IQsharpEngineTests.cs
```diff
@@ -245,6 +278,50 @@ public async Task Estimate()
...
+        {
+            var engine = await Init();
+            var channel = new MockChannel();
+
+            // Compile it:
+            await AssertCompile(engine, SNIPPETS.FailIfOne, "FailIfOne");
+
+            // Try running again:
+            // Note that noiseModel: null sets the noise model to be ideal.
+            await AssertNoisySimulate(engine, "FailIfOne", noiseModel: null);
```
swernli, line 308: This might be a bit silly, but is it worth having the inverse test where a noise model is specified that is so noisy that we expect `FailIfOne` to fail? Perhaps that's too probabilistic of a test such that it would not be a good candidate for CI, but is there any way the noise model could be set such that freshly allocated qubits in the ground state have a 100% chance to measure a One?

swernli (approved):
The changes look good to me, based on my understanding. Just a few random questions/ideas, but seems like these updates are good to go.

src/Kernel/Magic/SimulateNoise.cs
cgranade: Ah, this was due to a cross-repo change in the API not having propagated correctly; that should be fixed now such that I can uncomment. Thanks for catching this!

src/Tests/IQsharpEngineTests.cs
cgranade: I think that could be a good test, but it'd be difficult to write for exactly the reason you mention: such a test would be inherently probabilistic. Given that probabilistic tests have been a problem for our build already, I'm reticent to add more, but I think the question you raise is a good one to address as a part of improving test coverage more generally. Perhaps something like a completely amplitude damping channel?

cgranade:
I've made some updates to this PR to track updates in https://github.com/microsoft/qsharp-runtime/pull/709, to fix serialization issues in the original form of this PR. This PR also now adds the `%%qsharp` _Python_ magic command suggested by @guenp to help make it easier to [document the experimental feature](https://nbviewer.jupyter.org/github/microsoft/qsharp-runtime/blob/1b9aa59508c6e06b65d665820420842b25fac549/documentation/examples/experimental-simulators-from-python.ipynb).

image

It looks like the build is broken until https://github.com/microsoft/qsharp-runtime/pull/709 completes, but I'm a bit surprised that it's failing at the QIR step since that comes strictly before this repo in build order, such that it shouldn't be possible for _any_ change in this repo to affect how QIR support builds. @swernli, have you seen that error before? Thanks!

anjbur (approved):
LGTM!

cgranade:
Apologies for bothering again, @anpaz, but was wondering if you had any ideas on this one, as the local build is failing, but the package works locally on my machine and the e2e build is passing. From the error log, it looks like the `NativeInterface` class in qsharp-runtime can't find `Microsoft.Quantum.Experimental.Simulators.Runtime.dll`, but that DLL shows up in `native` directory of each runtime ID folder in the NuGet package.

Thanks for the help!

cgranade:
It looks like cc52fb0 should have fixed the build. The root cause was that the native runtime for the new experimental simulators feature requires glibc version 2.29, which is not available by default on older versions of Ubuntu (16.04). Updating the build agent seems to have resolved that issue.

cgranade:
The e2e failure seems to be similar as the one we saw earlier, namely an accidental circular dependency between the compiler and runtime repositories unrelated to the changes in this PR. Since earlier e2e builds passed and the local build is passing, and since this PR is against a feature branch, my vote would be to merge this PR anyway.

cgranade:
e2e build is passing at https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=27983&view=results, but that build isn't getting picked up by the status above. Merging to feature branch as per approval and passing e2e build.

PR #454, opened 2021-06-15
Capture and log exceptions in magic resolver.

cgranade:
This PR makes a minor change to magic resolution to better capture diagnostics about why magic commands may fail to load.


anjbur (approved):

cgranade:
The e2e failure looks to be that the build agent disconnected midway through. I've gone on and restarted the job that disconnected, and that should finish soon.

PR #456, opened 2021-06-16
Adding telemetry to %azure.connect

anpaz:
Adding telemetry to %azure.connect so we can keep track of whether the users as using a custom storage and what type of credentials they are using.


cgranade (approved):
Looks good to me!

vxfield (approved):
Look good to me.

PR #457, opened 2021-06-16
Fix versioning issues

anjbur:


cgranade (dismissed):

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -242,7 +242,7 @@ def _handle_message(self, msg, handlers=None, error_callback=None, fallback_hook
             else:
                 fallback_hook(msg)
 
-    def _execute(self, input, return_full_result=False, raise_on_stderr : bool = False, output_hook=None, _quiet_ : bool = False, **kwargs):
+    def _execute(self, input, return_full_result=False, raise_on_stderr : bool = False, output_hook=None, display_data_handler=None, _quiet_ : bool = False, **kwargs):
```
cgranade, line 245: Ah, it looks like the default of `None` is causing the build failure here. Will need to adapt lines 238 and 239 above:

```
if msg_type in handlers:
    # None is used to indicate that no action should be taken
    # for that given message type, such that we should not
    # call the fallback hook when msg_type is explicitly in
    # handlers, and is None.
    if handlers[msg_type] is not None:
        handlers[msg_type](msg)
else:
```

As an alternative, can adapt line 264 as follows:

```
        handlers = {
            'execute_result': (lambda msg: results.append(msg)),
            'render_execution_path':  (lambda msg: results.append(msg)),
            'display_data': display_data_handler if display_data_handler is not None else lambda msg: ...
        }
```

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -261,7 +261,7 @@ def log_error(msg):
         handlers = {
             'execute_result': (lambda msg: results.append(msg)),
             'render_execution_path':  (lambda msg: results.append(msg)),
-            'display_data': lambda msg: ...
+            'display_data': display_data_handler
```
anpaz: I'm trying this change locally, and it appears to me that this makes `import qsharp` fail from Python.

cgranade (approved):

anpaz (approved):

PR #458, opened 2021-06-19
Update Dockerfile

guenp:
Fixes msal_extensions.libsecret runtime dependency issue on Linux
See: https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/e889c44ab438501c32ad1763f49d20e29df0b031/msal_extensions/libsecret.py#L24

### DIFF
--- a/images/iqsharp-base/Dockerfile
+++ b/images/iqsharp-base/Dockerfile
@@ -29,7 +29,12 @@ RUN apt-get update && \
                        # given what we have already installed.
                        git \
                        # Used to retrieve node version information.
-                       curl && \
+                       curl \
+                       # Used to support PyGObject dependency of
+                       # msal_extensions package
+                       python3-gi \
+                       python3-gi-cairo \
+                       gir1.2-secret-1 && \
     # Upgrade optional dependencies brought in by the previous step.
     apt-get -y upgrade libidn2-0 && \
     # We clean the apt cache at the end of each apt command so that the caches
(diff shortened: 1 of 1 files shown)


xinyi-joffre (dismissed):
Thanks, Guen!

cgranade (approved):

guenp:
This fix unfortunately did not work; `apt-get` installs `gi` in the folder `/usr/lib/python3/dist-packages/` so I need to push another fix to add it to the path.

images/iqsharp-base/Dockerfile
```diff
@@ -124,7 +129,8 @@ RUN chown -R ${USER}:${USER} /home/${USER}/.azure
 USER ${USER}
 # Make sure that .NET Core is on the notebook users' path.
 ENV PATH=$PATH:${HOME}/dotnet:${HOME}/.dotnet/tools \
-    DOTNET_ROOT=${HOME}/dotnet
+    DOTNET_ROOT=${HOME}/dotnet \
+    PYTHONPATH=$PYTHONPATH:/usr/lib/python3/dist-packages/
```
cgranade, line 133: If possible, I'd recommend against doing this; the FROM line for this image bases on the `python:slim-buster` image, which maintains a copy of Python separate from the one used by the OS itself. Adding dist-packages could cause conflicts or other unexpected behavior by making all OS-level packages available to the separate install of Python added by the base image on the FROM line.

guenp: I'm not sure that is a real risk here; it doesn't look like there's any overlap between `dist-packages` and `site-packages`.
An alternative solution would be to add a `.pth` file; e.g. `dist-packages.pth` to `/usr/local/lib/python3.7/site-packages` containing the path to `/gi`. Would that work for you?

guenp:
Closing this, can reopen if this feature is still needed

PR #460, opened 2021-06-28
Pick default value for %azure.connect parameters from environment

anpaz:


src/AzureClient/Magic/ConnectMagic.cs
```diff
@@ -195,9 +215,9 @@ public override async Task<ExecutionResult> RunAsync(string input, IChannel chan
             else
             {
                 // look for each of the parameters individually
-                subscriptionId = inputParameters.DecodeParameter<string>(ParameterNameSubscriptionId, defaultValue: string.Empty);
-                resourceGroupName = inputParameters.DecodeParameter<string>(ParameterNameResourceGroupName, defaultValue: string.Empty);
-                workspaceName = inputParameters.DecodeParameter<string>(ParameterNameWorkspaceName, defaultValue: string.Empty);
+                subscriptionId = inputParameters.DecodeParameter<string>(ParameterNameSubscriptionId, defaultValue: System.Environment.GetEnvironmentVariable(EnvironmentSubscriptionId));
```
cgranade, line 218: Do these also need `?? string.Empty`?

cgranade (dismissed):
Approving as it looks like this approach would definitely work, but I'm curious what motivated looking at environment variables directly in the `%azure.connect` magic command rather than using the `ConfigurationBuilder` in `Program.cs` that merges command line arguments, appsettings.json, and environment variables.

https://github.com/microsoft/iqsharp/blob/460860292b15a55c56283a185b374b90004d8591/src/Tool/Program.cs#L69

src/Jupyter/ConfigurationSource/IConfigurationSource.cs
```diff
@@ -203,8 +203,32 @@ public T GetOptionOrDefault<T>(string optionName, T defaultValue, Func<string, T
 
         /// <summary>
         ///     Specifies the format used in dumping stabilizer states.
-        /// <summary>
+        /// </summary>
         public StabilizerStateVisualizationStyle ExperimentalSimulatorStabilizerStateVisualizationStyle =>
             GetOptionOrDefault("experimental.simulators.stabilizerStateStyle", StabilizerStateVisualizationStyle.MatrixWithDestabilizers);
+
+        /// <summary>
+        ///     Default SubscriptionId to use when connecting to Azure Quantum. Returns null if not configured.
```
cgranade, line 211: It looks like this returns string.Empty?

src/Jupyter/ConfigurationSource/IConfigurationSource.cs
```diff
@@ -203,8 +203,32 @@ public T GetOptionOrDefault<T>(string optionName, T defaultValue, Func<string, T
...
+        /// </summary>
+        public string SubscriptionId =>
+            GetOptionOrDefault("azurequantum.subscription.id", string.Empty);
+
+        /// <summary>
+        ///     Default Workspace's Resource Group to use when connecting to Azure Quantum. Returns null if not configured.
```
cgranade, line 217: Similarly here.

cgranade (approved):

cgranade (approved):

PR #464, opened 2021-07-07
Cgranade/capture diagnostics

### LINKED ISSUE #462: Allow capturing diagnostic data from Python hosts
Often, it's useful to be able to capture diagnostic information such as the states returned by `DumpMachine` and `DumpRegister`. This information is sent as JSON encoded into `'display_data'` messages across Jupyter sessions, such that wavevectors can be extracted by overriding display hooks. For example, the sample at https://github.com/microsoft/quantum/blob/2a541167f844ce03b898bccc784013514e9f6fb0/samples/simulation/gaussian-initial-state/host.py uses this technique to plot initial Gaussian states:

```python
# Add a new message handler to the IQ# client that listens for the output
# of DumpMachine and DumpRegister. This will let us make nicer plots
# in matplotlib.
old_handler = qsharp.client._handle_message
def handle_state_dump(message, **kwargs):
    if message['msg_type'] == 'display_data':
        data = (
            # Check both the old and new MIME types used by the IQ#
            # kernel.
            json.loads(message['content']['data'].get('application/json', "null")) or
            json.loads(message['content']['data'].get('application/x-qsharp-data', "null"))
        )
        if data is not None and 'amplitudes' in data:
            states.append(data['amplitudes'])
            return

    old_handler(message, **kwargs)
qsharp.client._handle_message = handle_state_dump
```

It would be nice if the `qsharp` package offered a hook to add such listeners directly without needing to dynamically replace the definition of the internal `qsharp.client._handle_message` method. This could be done in a variety of different approaches:

```
# As a context manager:
with qsharp.capture_display_data() as diagnostics:
     operation.simulate()
states = [data['amplitudes'] for data in diagnostics if 'amplitudes' in data]

# As a new special keyword argument to .simulate, .simulate_noise, etc.
# NB: need __ to prevent collision with Q# input names.
result, diagnostics = operation.simulate(__capture=True)

# As a function that sets a callback listener.
diagnostics = []
qsharp.add_display_data_callback(lambda display_data: diagnostics.append(display_data))
operation.simulate()
[data['amplitudes'] for data in diagnostics if 'amplitudes' in data]
```

cgranade:
This PR adds a new `qsharp.capture_diagnostics` context manager that can be used to capture `DumpMachine`, `DumpRegister`, and `DumpOperation` calls from `.simulate()`, `.simulate_noise()`, and other simulation methods on Q# callables. The new context manager supports representing captured diagnostics as QuTiP objects as well, so that they can be plotted and reused in research applications.

Fixes #462.


cgranade:
As some quick clarification to 5914d94, CI was failing above due to the test environment for `conda-build --test` requiring package dependencies that were not satisfiable. We had intentionally used older versions of several dependencies to avoid upstream bugs, but these bugs should since have been fixed such that relaxing test environment requirements should allow for satisfying dependencies.

cgranade:
During the course of fixing the CI build, I noticed that the unit tests for `qsharp` were not getting re-run inside the final conda package. The last few commits attempt to turn on at least the tests that are new to this PR.

anjbur (approved):
looks good to me!

PR #466, opened 2021-07-08
Improve error handling for %%qsharp cell magic.

cgranade:


src/Python/qsharp-core/qsharp/__init__.py
```diff
@@ -55,6 +55,10 @@ def compile(code : str) -> Union[QSharpCallable, List[QSharpCallable]]:
     :returns: A list of callables compiled from `code`, or a callable if exactly
         one callable is found.
     """
+    compiled = client.compile(code)
+    if compiled is None:
```
guenp, line 59: I think we discussed this but please remind me - Does `client.compile` raise any compilation errors or are we suppressing them?

guenp (approved):

src/Python/qsharp-core/qsharp/__init__.py
cgranade: Neither exactly; client.compile prints errors to stderr and returns `None` on compilation failure. We could consider capturing and raising as Python errors, but if so I'd like to defer that to a future PR since this one is purely a bugfix for existing behavior.

guenp: Sounds good, thanks!

PR #468, opened 2021-07-08
Fix message when no targets available

anjbur:


src/AzureClient/AzureClient.cs
```diff
@@ -41,8 +41,8 @@ public class AzureClient : IAzureClient
         private IEnumerable<TargetStatusInfo>? AvailableTargets => AvailableProviders?.SelectMany(provider => provider.Targets);
         private IEnumerable<TargetStatusInfo>? ValidExecutionTargets => AvailableTargets?.Where(AzureExecutionTarget.IsValid);
         private string ValidExecutionTargetsDisplayText =>
-            ValidExecutionTargets == null
-            ? "(no execution targets available)"
+            (ValidExecutionTargets == null || ValidExecutionTargets.Count() == 0)
+            ? "(no Q# execution targets available)"
```
cgranade, line 45: We may want to change this to "quantum computing" targets, in that Q# is the language used to write quantum programs, but that's not necessarily a property of the execution target itself. That is, targets see what Q# is compiled to, rather than Q# directly.

cgranade (dismissed):

src/AzureClient/AzureClient.cs
anjbur: Makes sense! I had set it to match the output string from `%azure.connect` if there aren't any computing targets ([here](https://github.com/microsoft/iqsharp/blob/f7090286fa14d739f67ea8b42b19136c7f9e0122/src/AzureClient/AzureClient.cs#L144)). I'll change it in both spots.

cgranade (approved):

PR #471, opened 2021-07-12
Improve performance of magic discovery.

cgranade:
This PR improves the performance with which magic commands and display encoders are discovered, reducing kernel startup times and the amount of time before the first cell can be executed. In particular, this PR:

- Completely removes an expensive logging call.
- Uses parallel LINQ to search assemblies in parallel.
- Resolves an issue where some assemblies were getting twice by changing the key type used for caching from `AssemblyInfo` to `AssemblyName`.
- Adds a blocklist of assemblies that never need to be searched for magic commands or display encoders.

### DIFF
--- a/src/Kernel/Magic/Resolution/MagicResolver.cs
+++ b/src/Kernel/Magic/Resolution/MagicResolver.cs
@@ -1,7 +1,8 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Immutable;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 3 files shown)


src/Kernel/Magic/Resolution/MagicResolver.cs
```diff
@@ -125,14 +136,7 @@ public IEnumerable<MagicSymbol> FindMagic(AssemblyInfo assm)
                         .Where(t =>
                         {
                             if (!t.IsClass && t.IsAbstract) { return false; }
-                            var matched = t.IsSubclassOf(typeof(MagicSymbol));
-
-                            // This logging statement is expensive, so we only run it when we need to for debugging.
-                            #if DEBUG
```
anpaz, line 131: I was running in Debug mode, so this could explain why it was eating so much of my resources.
However, this is very useful debug information, so maybe we should keep it?

anpaz (approved):

src/Kernel/Magic/Resolution/MagicResolver.cs
cgranade: Makes sense, added it back in!

xinyi-joffre (approved):

cgranade:
d42480b fixes a problem introduced by #470 where nondeterminism in service startup was causing some telemetry tests to randomly fail by making sure that services are awaited before beginning unit tests.

xinyi-joffre (approved):

PR #476, opened 2021-07-14
Begin porting custom shell handlers to use comms router

cgranade:
This PR parallels and depends on https://github.com/microsoft/jupyter-core/pull/78, using the new `CommsRouter` service to simplify custom messaging.

<s>In draft and will fail builds until https://github.com/microsoft/jupyter-core/pull/78 is included in a release and this PR is updated to use the new version.</s>


src/Tests/Startup.cs
```diff
@@ -63,6 +63,7 @@ public static void AddMocks(this IServiceCollection services)
             var shell = new MockShell();
             services.AddSingleton<IShellServer>(shell);
             services.AddSingleton<IShellRouter>(new MockShellRouter(shell));
```
anjbur, line 65: Is this line still necessary, with porting from shell to comms?

cgranade: It should be, yes. The shell router is still there and used to handle shell messages, including `comm_open` and `comm_msg`. This line makes sure that a mock for the shell router is made available during testing, allowing us to test functionality that depends on Microsoft.Jupyter.Core types without instantiating a real kernel.

src/Kernel/client/kernel.ts
```diff
@@ -147,52 +153,34 @@ class Kernel {
     }
 
     setupMeasurementHistogramDataListener() {
-        IPython.notebook.kernel.register_iopub_handler(
+        IPython.notebook.kernel.comm_manager.register_target<{state: DisplayableState}>(
```
anjbur, line 156: What about the other message handlers? Will those be moved to comms in a later PR?

anjbur (approved):
Looks good to me! Do these changes require any additional tests, beyond using comms in the tests that already existed?

PR #477, opened 2021-07-14
Enable ConfigurationSource to fallback to environment variables.

anpaz:


src/Jupyter/ConfigurationSource/IConfigurationSource.cs
```diff
@@ -51,17 +57,40 @@ public interface IConfigurationSource
         ///     <paramref name="optionName" /> if it exists, otherwise
         ///     the value of <paramref name="defaultValue" />.
         /// </returns>
-        public T GetOptionOrDefault<T>(string optionName, T defaultValue) =>
-            Configuration.TryGetValue(optionName, out var token)
-            ? token.ToObject<T>() ?? defaultValue
-            : defaultValue;
+        public T GetOptionOrDefault<T>(string optionName, T defaultValue, Func<string, T> parser)
```
cgranade, line 60: Strictly speaking, this is a breaking change as written. Would it make sense to make `parser` have nullable type, and to fall back to something like `System.Convert.ChangeType` if parser is missing?

src/Tests/ConfigurationSourceTests.cs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

cgranade (approved):

src/Jupyter/ConfigurationSource/IConfigurationSource.cs
anpaz: What I did is made all the convenience methods that picks a parser public, so it's almost fully backwards compatible now (at least for all basic datatypes). I rather make this than casting the object.

cgranade (approved):

PR #481, opened 2021-07-20
Add target ID and execution count to compile and magic events.

cgranade:
This PR adds new properties for target IDs and execution counts for compilation and magic command invocation events.


anjbur (approved):
The changes look good to me!

PR #483, opened 2021-07-21
Fix #478.

### LINKED ISSUE #478: Confusing error messages for mistyped magic commands
I accidentally typed `%simulator Rng` instead of `%simulate Rng`, which resulted in an error message
"C:\snippet_.qs(1,1): error QS3001: Syntax does not match any known patterns."

I think this error message is confusing. It looks like the cell when the magic command is not recognized is treated like Q# code even if the cell starts with `%`. I think it might be less confusing to assume that all cells that start with `%` were meant to be magic commands and have a dedicated error message listing all available magic commands when no such command is found.

cgranade:
This PR fixes #478 by detecting when compilation errors likely came from an invalid magic command (single-line cell starting with `%` that doesn't resolve to a magic command), and providing additional feedback as available.

### DIFF
--- a/src/Kernel/IQSharpEngine.cs
+++ b/src/Kernel/IQSharpEngine.cs
@@ -358,7 +358,7 @@ namespace Microsoft.Quantum.IQSharp.Kernel
                     if (input.TrimStart().StartsWith("%") && input.Split("\n").Length == 1)
                     {
                         var attemptedMagic = input.Split(" ", 2)[0];
-                        var msg = $"No such magic command {attemptedMagic}.";
+                        channel.Stderr($"No such magic command {attemptedMagic}.");
                         if (MagicResolver is MagicSymbolResolver iqsResolver)
                         {
                             var similarMagic = iqsResolver
@@ -369,9 +369,9 @@ namespace Microsoft.Quantum.IQSharp.Kernel
                                 .OrderBy(pair => pair.Item2)
                                 .Take(3)
                                 .Select(symbol => symbol.Name);
-                            msg += $"\nPossibly similar magic commands:\n\t{string.Join(", ", similarMagic)}";
+                            channel.Stderr($"Possibly similar magic commands:\n{string.Join("\n", similarMagic.Select(m => $"- {m}"))}");
                         }
-                        channel.Stderr(msg + "\nTo get a list of all available magic commands, run %lsmagic, or visit https://docs.microsoft.com/qsharp/api/iqsharp-magic/.");
+                        channel.Stderr("To get a list of all available magic commands, run %lsmagic, or visit https://docs.microsoft.com/qsharp/api/iqsharp-magic/.");
                     }
                     else
                     {
--- a/src/Kernel/IQSharpEngine.cs
+++ b/src/Kernel/IQSharpEngine.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using System;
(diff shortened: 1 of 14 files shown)


cgranade:
image

src/Kernel/IQSharpEngine.cs
```diff
@@ -342,7 +342,31 @@ public override async Task<ExecutionResult> ExecuteMundane(string input, IChanne
                 }
                 catch (CompilationErrorsException c)
                 {
-                    foreach (var m in c.Errors) channel.Stderr(m);
+                    // Check if the user likely tried to execute a magic
+                    // command and try to give a more helpful message in that
+                    // case.
+                    if (input.TrimStart().StartsWith("%") && input.Split("\n").Length == 1)
```
anpaz, line 348: since magics accept inputs on multiple lines, I'd say that we shouldn't restrict it to one-line inputs

src/Kernel/IQSharpEngine.cs
```diff
@@ -342,7 +342,31 @@ public override async Task<ExecutionResult> ExecuteMundane(string input, IChanne
...
+                        {
+                            var similarMagic = iqsResolver
+                                .FindAllMagicSymbols()
+                                .Select(symbol =>
+                                    (symbol.Name, symbol.Name.EditDistanceFrom(attemptedMagic))
+                                )
+                                .OrderBy(pair => pair.Item2)
+                                .Take(3)
+                                .Select(symbol => symbol.Name);
+                            msg += $"\nPossibly similar magic commands:\n\t{string.Join(", ", similarMagic)}";
```
anpaz, line 362: would it be better to output each message individually (i.e. call `channel.Stderr` on each line), as opposed to a single concatenated string?

src/Kernel/IQSharpEngine.cs
```diff
@@ -342,7 +342,31 @@ public override async Task<ExecutionResult> ExecuteMundane(string input, IChanne
...
+                    // Check if the user likely tried to execute a magic
+                    // command and try to give a more helpful message in that
+                    // case.
+                    if (input.TrimStart().StartsWith("%") && input.Split("\n").Length == 1)
+                    {
+                        var attemptedMagic = input.Split(" ", 2)[0];
+                        var msg = $"No such magic command {attemptedMagic}.";
+                        if (MagicResolver is MagicSymbolResolver iqsResolver)
+                        {
+                            var similarMagic = iqsResolver
```
anpaz, line 354: what happens if we don't find 3 similar magics, will the message be all weird?

src/Kernel/IQSharpEngine.cs
```diff
@@ -342,7 +342,31 @@ public override async Task<ExecutionResult> ExecuteMundane(string input, IChanne
...
+                                .FindAllMagicSymbols()
+                                .Select(symbol =>
+                                    (symbol.Name, symbol.Name.EditDistanceFrom(attemptedMagic))
+                                )
+                                .OrderBy(pair => pair.Item2)
+                                .Take(3)
+                                .Select(symbol => symbol.Name);
+                            msg += $"\nPossibly similar magic commands:\n\t{string.Join(", ", similarMagic)}";
+                        }
+                        channel.Stderr(msg + "\nTo get a list of all available magic commands, run %lsmagic, or visit https://docs.microsoft.com/qsharp/api/iqsharp-magic/.");
```
anpaz, line 364: I think it we output the magics we found as a list, they will actually show like:
* %magic1
* %magic2
* %magic3
which will probably look a little nicer.

anpaz (dismissed):
I would add a quick unittest to make sure this keeps working in the future and that it works as expected.

src/Kernel/IQSharpEngine.cs
cgranade: Agreed, but I did it this way to reduce the chance of a false positive (that is, an actual Q# compilation failure gets wrongly treated as a missing magic command). If possible, I'd like to keep it this way for now to be conservative with respect to false positives, and to revisit a more robust way of offering that feedback for multiline magics.

src/Kernel/IQSharpEngine.cs
cgranade: Good point, let me go make that change, then.

src/Kernel/IQSharpEngine.cs
cgranade: Good idea, let me go add that.

src/Kernel/IQSharpEngine.cs
cgranade: From testing, it looks like the worst case is that the suggestions are a bit off-topic. The actionable message about running `%lsmagic` or checking docs.microsoft.com should help in that case, I think?

cgranade:
Thanks for the review, @anpaz! I completely agree with adding unit tests for the specific new error message added with this PR. If possible, I'd like to make that change to this PR after https://github.com/microsoft/iqsharp/pull/485 is merged in, since that branch contains some nice extension methods that will make writing that test a bit easier.

anjbur (approved):

PR #485, opened 2021-07-21
Fix #484

### LINKED ISSUE #484: Submitting Q# operations with basic measurement feedback profile results in error
When compiling for targets supporting the [Basic Measurement Feedback profile](https://docs.microsoft.com/azure/quantum/concepts-targets-in-azure-quantum#quantum-processing-units-qpu-different-profiles) (e.g.: Honeywell) with IQ#, programs are incorrectly targeted for the No-Control Flow profile. As a result, when using IQ# either through Q# notebooks or Python host programs, some programs (e.g.: teleportation) are incorrectly reported as not being supported on affected targets.
 
image

This issue has affected IQ# usage (via Q# notebooks or Python host programs) of targets that support basic measurement feedback since public preview. Standalone Q# programs submitted with the `az` CLI tool are not affected by this issue.

## Steps to reproduce

- With IQ# 0.18.2106148911 installed, create a new Q# notebook.
- Execute a cell containing an operation using classical control flow. For example, teleportation:

```qsharp
open Microsoft.Quantum.Measurement;

operation PrepareBellPair(left : Qubit, right : Qubit) : Unit is Adj + Ctl {
    H(left);
    CNOT(left, right);
}

operation Teleport(msg : Qubit, target : Qubit) : Unit {
    use register = Qubit();

    PrepareBellPair(register, target);
    Adjoint PrepareBellPair(msg, register);

    if MResetZ(msg) == One      { Z(target); }
    if MResetZ(register) == One { X(target); }
}

operation RunTeleport() : Unit {
    use left = Qubit();
    use right = Qubit();

    H(left);
    Teleport(left, right);
    H(right);

    if M(right) == One {
        fail "Got wrong output from teleportation.";
    }
}
```

- Connect to an Azure Quantum workspace using the [`%azure.connect` command](https://docs.microsoft.com/qsharp/api/iqsharp-magic/azure.connect). Note that the workspace must have at least one provider that supports the Basic Measurement Feedback profile.
- Use the [`%azure.target` command](https://docs.microsoft.com/qsharp/api/iqsharp-magic/azure.target) to set the active target to a provider that supports the Basic Measurement Feedback profile. For example, to use the Honeywell API validator as a target: 
  ```
  %azure.target honeywell.hqs-lt-s1-apival
  ```
- Use either the `%azure.execute` or `%azure.submit` command to submit the operation using classical control flow to the target selected in the previous step. For example:
  ```
  %azure.execute RunTeleportation
  ```

cgranade:
This PR was collaboratively developed with @bettinaheim and fixes #484. This PR also improves unit test coverage to prevent regressions, slightly improves logging functionality to help identify causes of similar issues in the future, and addresses a small performance issue in magic command resolution discovered in the process of developing the fix for #484.

The main change needed to fix #484 is to regenerate C# code produced from library references when targeting for Azure Quantum service submission, allowing for Q# code in references to be transformed by compiler rewrite steps for that target.

### DIFF
--- a/src/AzureClient/EntryPoint/EntryPointGenerator.cs
+++ b/src/AzureClient/EntryPoint/EntryPointGenerator.cs
@@ -157,7 +157,20 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 throw new CompilationErrorsException(logger.Errors.ToArray());
             }
 
-            var entryPointOperationInfo = EntryPointAssemblyInfo.Operations.Single();
+            if (EntryPointAssemblyInfo.Operations.Count() <= 1)
+            {
+                Logger?.LogWarning("Entry point assembly contained zero or one operations; this may indicate that C# code is not being correctly regenerated. At least two operations (the entry point and the operation called from the entry point) are expected.");
+            }
+
+            var entryPointOperationInfo = EntryPointAssemblyInfo
+                .Operations
+                .Where(op => op.Header.QualifiedName.Namespace == "ENTRYPOINT" && op.Header.QualifiedName.Name == operationName)
+                .SingleOrDefault();
+
+            if (entryPointOperationInfo == null)
+            {
+                throw new Exception("Entry point assembly either contained multiple entry point operations, or contained zero.");
+            }
 
             // Construct the EntryPointInfo<,> object
             var parameterTypes = entryPointOperationInfo.RoslynParameters.Select(p => p.ParameterType).ToArray();
--- a/src/AzureClient/EntryPoint/EntryPointGenerator.cs
+++ b/src/AzureClient/EntryPoint/EntryPointGenerator.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 if (!workspaceAssemblies.Any() || logger.HasErrors)
                 {
                     Logger?.LogError($"Error compiling workspace.");
-                    throw new CompilationErrorsException(logger.Errors.ToArray());
+                    throw new CompilationErrorsException(logger);
                 }
 
                 WorkspaceAssemblies = workspaceAssemblies.ToArray();
@@ -135,7 +135,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 if (SnippetsAssemblyInfo == null || logger.HasErrors)
                 {
                     Logger?.LogError($"Error compiling snippets.");
-                    throw new CompilationErrorsException(logger.Errors.ToArray());
+                    throw new CompilationErrorsException(logger);
                 }
 
                 compilerMetadata = compilerMetadata.WithAssemblies(SnippetsAssemblyInfo);
@@ -154,7 +154,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             if (EntryPointAssemblyInfo == null || logger.HasErrors)
             {
                 Logger?.LogError($"Error compiling entry point for operation {operationName}.");
-                throw new CompilationErrorsException(logger.Errors.ToArray());
+                throw new CompilationErrorsException(logger);
             }
 
             if (EntryPointAssemblyInfo.Operations.Count() <= 1)
--- a/src/AzureClient/EntryPoint/EntryPointGenerator.cs
+++ b/src/AzureClient/EntryPoint/EntryPointGenerator.cs
@@ -162,14 +162,23 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 Logger?.LogWarning("Entry point assembly contained zero or one operations; this may indicate that C# code is not being correctly regenerated. At least two operations (the entry point and the operation called from the entry point) are expected.");
             }
 
-            var entryPointOperationInfo = EntryPointAssemblyInfo
+            var entryPointOperations = EntryPointAssemblyInfo
                 .Operations
-                .Where(op => op.Header.QualifiedName.Namespace == "ENTRYPOINT" && op.Header.QualifiedName.Name == operationName)
+                .Where(op => op.Header.Attributes.Any(
+                    attr =>
+                    {
+                        var qName = attr.TypeId.ValueOr(null);
+                        return qName != null &&
+                               qName.Name == "EntryPoint" &&
+                               qName.Namespace == "Microsoft.Quantum.Core";
+                    }
+                ));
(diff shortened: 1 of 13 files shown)


src/Tests/IQsharpEngineTests.cs
```diff
@@ -938,7 +869,65 @@ public async Task TestTraceMagic()
...
+                    .ExecutesSuccessfully()
+                .Input("%azure.target ionq.mock")
+                    .ExecutesSuccessfully()
+                .Input("%azure.submit RunTeleport")
+                    .ExecutesWithError(errors => errors.Any(error =>
+                        // Use StartsWith to avoid mismatching on newlines at
+                        // the end.
+                        error.StartsWith(
+                            "The Q# operation RunTeleport could not be " +
+                            "compiled as an entry point for job execution."
```
anpaz, line 925: I have to say the message here is kind of confusing.

src/Tests/TestExtensions.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            internal string Code;
+            internal int? CursorPos;
+        }
+
+        internal class UsingEngineAssert
+        {
+            internal IQSharpEngine Engine;
+        }
+        internal static async Task<T> WithMockAzure<T>(this Task<T> engine)
```
anpaz, line 32: nitpick: line space.

src/Kernel/Magic/Resolution/MagicResolver.cs
```diff
@@ -142,7 +143,7 @@ public IEnumerable<MagicSymbol> FindMagic(AssemblyInfo assm)
                         .GetTypes()
                         .Where(t =>
                         {
-                            if (!t.IsClass && t.IsAbstract) { return false; }
+                            if (!t.IsClass || t.IsAbstract) { return false; }
```
anpaz, line 146: !

cgranade:
The e2e build failure above is the same as the one reported and fixed in https://github.com/microsoft/Quantum/pull/528. Will push an empty commit to invalidate previous e2e builds and bring in the fix from https://github.com/microsoft/Quantum/pull/528.

src/Core/Compiler/CompilerService.cs
```diff
@@ -247,31 +255,31 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
                     metadata.RoslynMetadatas,
                     options);
 
-                var fromSources = qsCompilation.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.EndsWith(".qs")));
+                var fromSources = regenerateAll
+                                  ? qsCompilation.Namespaces
+                                  : qsCompilation.Namespaces.Select(ns => FilterBySourceFile.Apply(ns, s => s.EndsWith(".qs")));
 
-                // Only create the serialization if we are compiling for an execution target:
+                // Only create the serialization if we aren't compiling for an execution target:
```
anpaz, line 262: Looks like this comment does not even apply now, does it? Looks like we're always doing serialization.

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -154,10 +154,32 @@ internal class EntryPointGenerator : IEntryPointGenerator
...
             {
                 Logger?.LogError($"Error compiling entry point for operation {operationName}.");
-                throw new CompilationErrorsException(logger.Errors.ToArray());
+                throw new CompilationErrorsException(logger);
             }
 
-            var entryPointOperationInfo = EntryPointAssemblyInfo.Operations.Single();
+            if (EntryPointAssemblyInfo.Operations.Count() <= 1)
+            {
+                Logger?.LogWarning("Entry point assembly contained zero or one operations; this may indicate that C# code is not being correctly regenerated. At least two operations (the entry point and the operation called from the entry point) are expected.");
```
anpaz, line 162: I think this message is confusing and doesn't really tell the user what to do to fix the problem.

anpaz (approved):
I would just be on the lookout about making sure performance doesn't degrade too much now that we're always doing binary serialization of the AST. Other than that logtm.

src/AzureClient/EntryPoint/EntryPointGenerator.cs
cgranade: Fair enough! In this particular case, the message is only written to logs and is not user-visible from within notebooks or host programs, such that my intent was for the message to be actionable by developers or contributors rather than by users. That said, can look at rewording some here.

src/Core/Compiler/CompilerService.cs
cgranade: Good point, will remove.

src/Kernel/Magic/Resolution/MagicResolver.cs
cgranade: Indeed, this was a problem! As it turns out, the net effect of this typo was to cause spurious errors to appear in debug logs as well as a slight degradation in the performance of magic command discovery.

src/Tests/IQsharpEngineTests.cs
cgranade: Fair enough; this was the message as it originally existed before this PR, such that this unit test only makes sure that the message is correctly omitted when submitting to targets that don't support the basic measurement feedback profile. As above, happy to consider changing the message, though.

src/Tests/TestExtensions.cs
cgranade: Ah, good catch, thank you.
```suggestion

        internal static async Task<T> WithMockAzure<T>(this Task<T> engine)
```

src/Tests/IQsharpEngineTests.cs
cgranade: Source for the above error message: https://github.com/microsoft/iqsharp/blob/b789dcdeee32a578793f6be1a6db9f89c0c54161/src/AzureClient/AzureClient.cs#L318. If possible, for time constraints I'd prefer to defer an improvement to that message for now. Thank you!

src/AzureClient/EntryPoint/EntryPointGenerator.cs
cgranade: Fixed to clearly indicate that the log message is an indicator of an internal bug and not likely to be user-actionable beyond filing an issue.

anpaz (approved):

PR #493, opened 2021-08-07
Fix completion in Python.

cgranade:
While writing some samples using Python, I noticed an odd error internal to the IPython kernel when trying to use code completion from VS Code. This wound up giving me a clue to why code completion was not working for Q# namespaces imported into Python, thankfully!


anjbur (approved):
Looks good to me!

PR #497, opened 2021-08-13
Fix regression in error formatting.

cgranade:
This PR fixes a small regression in how error messages are formatted, and allows for consumers of the Q# compilation service to get access to underlying compilation errors rather than relying on formatting.


anjbur (approved):
LGTM!

PR #498, opened 2021-08-17
Add active subscription, workspace to event metadata

cgranade:
This PR adds information about active subscriptions and workspaces to metadata for events, and refactors property handling to make ensure that metadata is added to events more uniformly.


src/Tool/Program.cs
```diff
@@ -74,6 +74,7 @@ public static int Main(string[] args)
                         {
                             ["TELEMETRY_OPT_OUT"] = nameof(TelemetryOptOut),
                             ["USER_AGENT"] = "UserAgent",
+                            ["USER_AGENT_EXTRA"] = "UserAgentExtra",
```
anjbur, line 77: What is this change for? I don't see where it's used elsewhere in the PR

anjbur (approved):
Left a few questions, but otherwise looks good to me!

xinyi-joffre (approved):

src/Tests/HttpServerIntegrationTests.cs
```diff
@@ -74,7 +74,7 @@ public async Task CompileAndSimulateViaApi()
                 // TODO: Use constant strings for these http requests. Port and paths.
 
                 // Compile a snippet with the server and get back a list of now executable operations.
-                var compileResult = await "http://localhost:8888/api"
+                var compileResult = await "http://localhost:8008/api"
```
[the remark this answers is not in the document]
cgranade: Ah, this was because every bloody time I ran tests locally they failed due to conflicting with a Jupyter Notebook session I had open in the background. Making the tests uses a port that's less likely to conflict with Jupyter improves DX significantly.

src/Tool/Program.cs
cgranade: Apologies, forgot to note that. @xinyi-joffre had noted that we allowed overriding other properties via environment variables, but that `UserAgentExtra` was inadvertently omitted.

PR #499, opened 2021-08-18
Eagerly load references in the background.

cgranade:
This PR filters references passed to the Q# compiler to only those not known to be classical, and eagerly loads references in the background to make better use of time between when users execute each cell. This should reduce the time needed to execute the first cell, especially in cases where users don't execute the cell immediately after starting the kernel (e.g.: read the preamble first).


anjbur (commented):
Could we add unit tests for the new functionality? Looks good to me otherwise.

cgranade:
> Could we add unit tests for the new functionality? Looks good to me otherwise.

Thanks for taking a look! For unit tests, are you meaning the filtering on assemblies that get passed to the Q# compiler? Let me go look at that. In the meantime, the switch from `Lazy` to `Task` should be covered by existing tests, since there's no new functionality introduced by that change.

anjbur:
@cgranade Sorry for not being more specific! Yes, I meant on the assembly filtering. Agreed that current tests should cover the changes in `Snippet.cs`

cgranade:
No worries at all, just wanted to make sure I understood! Let me go add that, then.

anjbur (approved):

PR #501, opened 2021-08-19
Fix #500: Support Python 3.8 and 3.9.

### LINKED ISSUE #500: conda package requires Python 3.7, but default in new Anaconda installs is 3.8.
**Describe the bug**
The `qsharp` and `iqsharp` conda packages currently require Python 3.7.x, but the current default for new Anaconda installs is 3.8.8, such that attempting to run `conda install -c quantum-engineering qsharp` on a fresh install results in a package conflict.

cgranade:
This PR adds Python 3.8 and 3.9 to the pack script for conda packages, fixing #500 at the cost of more disk space on our anaconda.org channel.


anjbur:
Are our packages compatible with 3.8 and 3.9? I thought we required 3.7 because of compatibility issues (although I can't find that now in the docs, so I'm not sure where that assumption came from).

cgranade:
I don't recall a 3.8 incompatability, but it does look like `conda` found a conflict with 3.9. I'll step back to just 3.7 and 3.8 for now, then, since 3.8.8 is the current default provided by `conda` on a fresh install.

build/pack-conda.ps1
```diff
@@ -19,7 +19,7 @@ $PSVersionTable | Format-Table | Out-String | Write-Host;
 function Pack-CondaRecipe() {
     param(
         [string] $Path,
-        [string[]] $PythonVersions = @("3.7")
+        [string[]] $PythonVersions = @("3.7", "3.8")
```
vxfield, line 22: Let's see if we can include 3.9 as it's the default for miniconda.
```suggestion
        [string[]] $PythonVersions = @("3.7", "3.8", "3.9")
```

cgranade: Agreed, I realize I forgot to push my commit with this change. Thanks for the reminder!

vxfield (dismissed):
LGTM

cgranade:
@vxfield: Thanks for the review! Looks like had to relax one more version dependency in test environments, so would appreciate if you could take another look. Thanks!

vxfield (approved):

PR #507, opened 2021-08-25
Add count parameter to %azure.jobs list

anpaz:
A workspace could have thousands of jobs, and currently the `%azure.jobs` returns all of them with no way to limit how many jobs it should return.
This introduces a new parameter to limit the number of jobs that will be returned. It defaults to 30 (like the Portal).
If reaches the count of jobs, a message is printed to the user.

### DIFF
--- a/src/AzureClient/Magic/JobsMagic.cs
+++ b/src/AzureClient/Magic/JobsMagic.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                         - A string to filter the list of jobs. Jobs which have an ID, name, or target
                         containing the provided filter parameter will be displayed. If not specified,
                         all recent jobs are displayed.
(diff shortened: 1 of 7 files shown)


src/AzureClient/Magic/JobsMagic.cs
```diff
@@ -48,7 +49,8 @@ public JobsMagic(IAzureClient azureClient, ILogger<JobsMagic> logger)
                         - A string to filter the list of jobs. Jobs which have an ID, name, or target
                         containing the provided filter parameter will be displayed. If not specified,
                         all recent jobs are displayed.
```
cgranade, line 51: If I understand your change below correctly, this line is out of date?
```suggestion
                        the 30 most recent jobs are displayed.
```
It may also be worth using a constant for the default value, similar to the strategy you used with ParameterNameCount.

cgranade (approved):

PR #514, opened 2021-09-08
Revert "Eagerly load references in the background."

cgranade:
This PR reverts #499, as that PR was observed to increase startup performance at the cost of a large memory overhead. We will continue investigate ways of improving startup performance via background loading of packages going forward.


kuzminrobin (approved):

cgranade:
Build failures should be resolved by https://github.com/microsoft/iqsharp/pull/516.

PR #526, opened 2021-09-23
Enable temporary support for IQuantumMachineSubmissionContext.InputParams in IQSharp

ricardo-espinoza:
As part of pull request [qsharp-runtime:829](https://github.com/microsoft/qsharp-runtime/pull/829), we're adding a property in the interface `IQuantumMachineSubmissionContext` called `InputParams` that will be used to pass target specific parameters to Azure Quantum.

This allowing IQSharp to build with this change for now, however, support for this functionality will be added in the near future.

### DIFF
--- a/src/AzureClient/AzureSubmissionContext.cs
+++ b/src/AzureClient/AzureSubmissionContext.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Quantum.IQSharp.Jupyter;
 using Microsoft.Quantum.Runtime;
+using System.Collections.Immutable;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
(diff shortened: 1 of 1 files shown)


src/AzureClient/AzureSubmissionContext.cs
```diff
@@ -37,6 +38,13 @@ public sealed class AzureSubmissionContext : IQuantumMachineSubmissionContext
         /// </summary>
         public string OperationName { get; set; } = string.Empty;
 
+        // NOTE: The following member is inherited from IQuantumMachineSubmissionContext, but it's not used.
+        //       This is supporting the change in https://github.com/microsoft/qsharp-runtime/pull/829
+        //       Notice that this is different from the existing 'InputParameters' member below.
+        //       Upcoming work will enable this functionality and align those similar named items.
+        /// <inheritdoc/>
+        public ImmutableDictionary<string, string> InputParams { get; set; } = new Dictionary<string, string>();
```
cgranade, line 46: It feels a little odd to declare this property as an immutable dictionary, then populate it with a mutable dictionary? I'm guessing that relies on an implicit operator, but it may be better to explicitly refer to the empty immutable dictionary:

```suggestion
        public ImmutableDictionary<string, string> InputParams { get; set; } = Dictionary<string, string>.Empty;
```

cgranade (approved):
Approved with a minor suggestion.

PR #528, opened 2021-09-25
Begin adding support for device capabilities.

cgranade:
This PR adds support for logging device capabilities (# of CPUs, total RAM rounded to units of GiB).


src/Core/PerformanceMonitor.cs
```diff
@@ -13,7 +13,11 @@ namespace Microsoft.Quantum.IQSharp
     public class DeviceCapabilitiesEvent : Event<DeviceCapabilitiesArgs>
     {}
 
+    #if NET5_0 || NET6_0 || NET5_0_OR_GREATER
     public record DeviceCapabilitiesArgs
```
vxfield, line 17: I wonder if we should postpone moving to records until we require .NET 5.0 or greater for the whole project.
Otherwise we will have to spread the code with compiler conditionals like you suggested and not sure it's worth the effort...

cgranade: Agreed, only did so in this case because we're as close to November's release of .NET 6, such that we should be able to remove this soon.

src/Tool/Telemetry.cs
```diff
@@ -228,6 +235,11 @@ public void SetSharedContextIfChanged(IMetadataController metadataController, st
 
     public static class TelemetryExtensions
     {
+        // Generate a new UUID to use as a session ID.
+        // By making a static property of a non-generic static class, we
+        // ensure that there's one session ID per process.
+        internal static readonly Guid SessionId = Guid.NewGuid();
```
xinyi-joffre, line 241: I wonder if we want to rename this to KernelInstanceId or something like that? I am just finishing up a privacy review, and they typically categorize SessionIds as EUPI because it's usually referring to user's login session. Since we are not persisting it according to login sessions, it may be nice to avoid confusion or raised flags around storing SessionIds, and go with term that'll cause less alarm for privacy reviewers.

cgranade: Good idea, let me go on and rename that, then.

PR #529, opened 2021-09-26
Remove a redundant .orig file

jond01:
Reverts https://github.com/microsoft/iqsharp/pull/37/files#diff-a122204d440a992940dd4c69f2497ac1c678a1394959c5d873d14567e82e654e


jond01:
Since it is a bit confusing and not necessary, it's better to remove it

PR #538, opened 2021-11-04
Install .NET6 SDK as well in iqsharp-base Docker image

ricardo-espinoza:
Installing the SDK for .NET 6.0 as part of the iqsharp-docker image in preparation for changes tracked by:
https://github.com/microsoft/qsharp-compiler/issues/1224

### DIFF
--- a/images/iqsharp-base/Dockerfile
+++ b/images/iqsharp-base/Dockerfile
@@ -50,8 +50,7 @@ RUN wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor 
     chown root:root /etc/apt/trusted.gpg.d/microsoft.asc.gpg && \
     chown root:root /etc/apt/sources.list.d/microsoft-prod.list && \
     apt-get -y update && \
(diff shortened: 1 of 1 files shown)


images/iqsharp-base/Dockerfile
```diff
@@ -50,6 +51,7 @@ RUN wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor
     chown root:root /etc/apt/sources.list.d/microsoft-prod.list && \
     apt-get -y update && \
     apt-get -y install dotnet-sdk-3.1 && \
+    apt-get -y install dotnet-sdk-6.0 && \
```
cgranade, line 54: Very minor, but it may be faster to build this image with these grouped in

cgranade (approved):

images/iqsharp-base/Dockerfile
ricardo-espinoza: Makes sense. :)

PR #540, opened 2021-11-05
Make InputParams (aka jobParams) functional

warren-jones:
New code in this branch enables passing provider-specific custom job parameters to an execution target from a Jupyter notebook or a Python host program.  The user may now specify "jobParams=" followed by one or more JSON {"key":"value",...} pairs when submitting a job.  This adds functionality to the work started in [PR 526](https://github.com/microsoft/iqsharp/pull/526)


src/AzureClient/Magic/ExecuteMagic.cs
```diff
@@ -52,6 +52,8 @@ public ExecuteMagic(IAzureClient azureClient, ILogger<ExecuteMagic> logger)
 
                         - `{AzureSubmissionContext.ParameterNameJobName}=<string>`: Friendly name to identify this job. If not specified,
                         the Q# operation or function name will be used as the job name.
+                        - `{AzureSubmissionContext.ParameterNameJobParams}=< ImmutableDictionary<string, string>>`: Provider-specific job parameters
```
cgranade, line 55: The type `ImmutableDictionary<string, string>` has no meaning in this context; readers will be working either in a Q# notebook or in a Q# + Python workflow, neither of which has an exact mapping for that C# type. In general, the types listed in this documentation string are really intended to be informal summaries for users as to what kind of input users can expect to pass.

src/AzureClient/Magic/SubmitMagic.cs
```diff
@@ -51,6 +51,8 @@ public SubmitMagic(IAzureClient azureClient, ILogger<SubmitMagic> logger)
 
                         - `{AzureSubmissionContext.ParameterNameJobName}=<string>`: Friendly name to identify this job. If not specified,
                         the Q# operation or function name will be used as the job name.
+                        - `{AzureSubmissionContext.ParameterNameJobParams}=< ImmutableDictionary<string, string>>`: Provider-specific job parameters
```
cgranade, line 54: Same here.

cgranade (commented):
Two initial thoughts:

- I don't see any change to unit tests, would you please add coverage for the new parameter?
- How will this be accessible from Q# + Python workflows? What changes to `qsharp/clients/iqsharp.py` need to be made for Python users?

src/AzureClient/Magic/ExecuteMagic.cs
warren-jones: Looking at unit tests now -- thought I'd get this initial code out there for comments while I worked on them.  I will address the Q# + Python workflow requirements next.

Regarding the references to ImmutableDictionary<string, string> in the documentation:  I was following the pattern used for the other parameters.  Should a simplified description of ImmutableDictionary<string, string> be substituted, or should that data type not be mentioned at all?

src/AzureClient/Magic/SubmitMagic.cs
warren-jones: Please see reply to the ExecuteMagic.cs comment.

src/AzureClient/Magic/ExecuteMagic.cs
cgranade: > Looking at unit tests now -- thought I'd get this initial code out there for comments while I worked on them.

Makes sense, but in that case I'd suggest converting this to a draft PR so that it doesn't accidentally get merged before unit tests are added and the Q# + Python interop interface is modified for this new feature.

> Regarding the references to ImmutableDictionary<string, string> in the documentation: I was following the pattern used for the other parameters. Should a simplified description of ImmutableDictionary<string, string> be substituted, or should that data type not be mentioned at all?

The other entries use informal descriptions like "string" and "integer," since there's not any formal notion of types when calling magic commands from within a Q# notebook. In that sense, magic commands work a lot like command-line options (`--foo=bar`), such that simplified descriptions work well together with the more explicit examples later in the command documentation.

warren-jones: - Feature is supported in the Q# + Python interop interface
- Unit tests have been added (C# and Python)
- Documentation has been simplified and clarified

ricardo-espinoza (approved):
Looks good to me, but make sure you get approval from @anjbur before merging.

src/Tests/AzureClientMagicTests.cs
```diff
@@ -39,6 +40,12 @@ public class AzureClientMagicTests
         private readonly string operationName = "TEST_OPERATION_NAME";
         private readonly string targetId = "TEST_TARGET_ID";
 
+        private readonly string jobParams = "{\"TEST_KEY1\":\"TEST_VAL1\",\"TEST_PATH\":\"C:\\ABC\\XYZ\"}";
```
cgranade, line 43: This may be easier to read in terms of C#'s `@""` strings.

src/Tests/AzureClientMagicTests.cs
```diff
@@ -232,6 +239,14 @@ public void TestSubmitMagic()
             submitMagic.Test($"{operationName}");
             Assert.AreEqual(AzureClientAction.SubmitJob, azureClient.LastAction);
             Assert.IsTrue(azureClient.SubmittedJobs.Contains(operationName));
+
+            // jobParams argument
+            Assert.IsTrue(azureClient.JobParams.IsEmpty);
+            submitMagic.Test($"{operationName} jobParams={jobParams}");
+            Assert.IsTrue(azureClient.JobParams.TryGetValue(jobParamsKey1, out string value1));
+            Assert.AreEqual(value1, jobParamsVal1);
```
cgranade, line 247: This pattern happens enough that it may be worth adding a new `HasKey` extension method to the `Assert.That` object. Roughly,

```csharp
void HasKey<TKey, TValue>(this Assert assert, IDictionary<TKey, TValue> dict, TKey key, TValue expected)
{
     Assert.IsTrue(dict.TryGetValue(key, out TValue actual));
     Assert.AreEqual(actual, expected);
}
```

src/Python/qsharp-core/qsharp/tests/test_azure.py
```diff
@@ -138,6 +138,10 @@ def _test_workspace_with_providers_after_connection():
     assert isinstance(retrieved_job, AzureJob)
     assert job.id == retrieved_job.id
 
+    # Submit a job with the "jobParams" parameter
+    job2 = qsharp.azure.submit(op, jobParams={"key1": "value1", "key2": "value2"})
```
cgranade, line 142: Thanks for adding these tests; looks like the Python side picked up the new parameter correctly, then! Do we need to modify the docstrings accordingly?

src/AzureClient/AzureSubmissionContext.cs
```diff
@@ -38,11 +40,19 @@ public sealed class AzureSubmissionContext : IQuantumMachineSubmissionContext
...
-        // NOTE: The following member is inherited from IQuantumMachineSubmissionContext, but it's not used.
-        //       This is supporting the change in https://github.com/microsoft/qsharp-runtime/pull/829
-        //       Notice that this is different from the existing 'InputParameters' member below.
-        //       Upcoming work will enable this functionality and align those similar named items.
-        /// <inheritdoc/>
+        /// <summary>
+        ///     Provider-specific job parameters to be passed to the execution target, expressed as one or more JSON {"key":"value",...} pairs.
+        /// </summary>
+        /// <remarks>
+        ///     These parameters apply to %azure.execute. and %azure.submit. The JSON may not contain separating spaces when used in a Jupyter notebook.
```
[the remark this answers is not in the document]
warren-jones: Tags added as suggested.  Also applied these tags in lines 67 and 78 for consistency.

src/Jupyter/Extensions.cs
```diff
@@ -251,6 +252,29 @@ public static T DecodeParameter<T>(this Dictionary<string, string> parameters, s
...
                 return defaultValue!;
             }
+
+            // If this is a JSON-formatted parameter that's being deserialized into a dictionary, remove the extra quotes and backslashes.   
+            if (type == typeof(ImmutableDictionary<string, string>) && parameterValue.Length > 1)
+            {
+                // Jupyter wraps JSON in double quotes. Make sure that's what we have...
+                if (parameterValue[0] == '"' && parameterValue[parameterValue.Length - 1] == '"')
+                {
+                    StringBuilder parameterValueBuilder = new StringBuilder(parameterValue.Length);
```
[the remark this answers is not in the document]
warren-jones: This line was eliminated by the following code re-write.

src/Jupyter/Extensions.cs
```diff
@@ -251,6 +252,29 @@ public static T DecodeParameter<T>(this Dictionary<string, string> parameters, s
...
+                    for (int i = 1; i < parameterValue.Length - 1; i++)     // Ignore (remove) the leading and trailing quotes.
```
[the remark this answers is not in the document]
warren-jones: Eliminated the loop.  Simplified the code.

src/Tests/AzureClientMagicTests.cs
warren-jones: Good point, the `@` notation does look cleaner.  Also changed line 47 for consistency.

src/Tests/AzureClientMagicTests.cs
warren-jones: In my humble opinion, that extra layer of abstraction would make this code a little harder to read.  Granted, there is a bit of repetition, but not enough to significantly complicate maintenance.

src/Python/qsharp-core/qsharp/tests/test_azure.py
warren-jones: The docstrings in [azure.py](https://github.com/microsoft/iqsharp/blob/main/src/Python/qsharp-core/qsharp/azure.py) refer to the magic docs, so we're covered.   Added jobParams to examples in the magic docs.

src/AzureClient/Magic/ExecuteMagic.cs
```diff
@@ -93,9 +95,9 @@ public ExecuteMagic(IAzureClient azureClient, ILogger<ExecuteMagic> logger)
                         @"
                             Run a Q# operation defined as `operation MyOperation(a : Int, b : Int) : Result`
                             on the active target in the current Azure Quantum workspace,
-                            specifying a custom job name, number of shots, timeout, and polling interval:
+                            specifying a custom job name, number of shots, timeout, polling interval, and provider-specific job parameters:
                             ```
-                            In []: %azure.submit MyOperation a=5 b=10 jobName=""My job"" shots=100 timeout=60 poll=10
+                            In []: %azure.submit MyOperation a=5 b=10 jobName=""My job"" shots=100 timeout=60 poll=10 jobParams={""Key1"":""Val1"",""Key2"":""Val2""}
```
guenp, line 100: Should this be `--a=5 --b=10`

anjbur: The way inputs are currently parsed, adding `--` to the front would prevent the parameters from being recognized. It's a change we could theoretically make to the input parsing, but it'd be out of scope to this PR!

src/Tests/AzureClientMagicTests.cs
cgranade: Fair enough; I've found such abstractions to be highly useful in the code base so far, in part because it reduces friction to writing new tests that would otherwise introduce more duplication. This abstraction also reduces the probability of forgetting to check that the key exists when writing new tests.

That said, it is a minor comment, and something easily handled in code maintenance following this PR regardless.

src/AzureClient/Magic/ExecuteMagic.cs
cgranade: Ah, that's a good point; I hadn't noticed that the way its written, the new parameter strictly prevents adding `jobParams` as a Q# parameter name. That's probably minor, but we maybe should handle that better going forward. Since doing so would be a nontrivial breaking change, let's make sure to create an issue now and defer to that issue.

warren-jones: Resolving this conversation: Out of scope. The issue will be deferred.

src/Tests/AzureClientMagicTests.cs
ricardo-espinoza: Thanks a lot @cgranade ; for the purposes of this change let's leave this as is. I'll mark the comment as resolved. Going forward, it it makes sense to abstract this as part of feature work, let's do it. :)

anjbur (approved):
Everything looks good to me!

PR #541, opened 2021-11-14
Add default timeout to `qsharp._execute`

anpaz:
Currently, by default there is currently no timeout for call to the IQsharp kernel from the `qsharp` package. In general this has and should not been a problem as most calls will return immediately. However, some of the tests run during conda validation are causing some deadlock and the call never returns.
Changing the default timeout to be 2 min for most operations. Leaving `None` only for `simulate` and `estimate` as those can be long-lived operations.

### DIFF
--- a/src/Python/qsharp-core/qsharp/clients/iqsharp.py
+++ b/src/Python/qsharp-core/qsharp/clients/iqsharp.py
@@ -168,14 +168,14 @@ class IQSharpClient(object):
         return self._execute("%project", raise_on_stderr=False)
 
     def simulate(self, op, **kwargs) -> Any:
-        kwargs.setdefault('timeout', 180)
+        kwargs.setdefault('timeout', None)
         return self._execute_callable_magic('simulate', op, **kwargs)
 
     def toffoli_simulate(self, op, **kwargs) -> Any:
         return self._execute_callable_magic('toffoli', op, **kwargs)
 
     def estimate(self, op, **kwargs) -> Dict[str, int]:
-        kwargs.setdefault('timeout', 180)
+        kwargs.setdefault('timeout', None)
         raw_counts = self._execute_callable_magic('estimate', op, **kwargs)
         # Note that raw_counts will have the form:
         # [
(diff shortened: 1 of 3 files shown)


src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -168,12 +168,14 @@ def get_projects(self) -> List[str]:
         return self._execute("%project", raise_on_stderr=False)
 
     def simulate(self, op, **kwargs) -> Any:
+        kwargs.setdefault('timeout', None)
```
cgranade, line 171: Would it make sense to override for simulate_noise well?

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -168,12 +168,14 @@ def get_projects(self) -> List[str]:
...
         return self._execute_callable_magic('simulate', op, **kwargs)
```
cgranade, line 172: I think the kwargs here get inserted into the command being executed (e.g.: `%simulate RunMain timeout=120`) rather than a parameter to `_execute` itself. We may need a new `_timeout_` parameter on `_execute_callable_magic` and `_execute`, similar to the existing `_quiet_` parameter.

cgranade:
I think this looks good, thanks for adding this! Just had the one question about how to handle turning off timeouts in simulate and friends. Thanks!

cgranade (commented):
Looks good, thanks for adding this! Just had the one question about how to handle turning out timeouts in `simulate` and friends. Thanks!

src/Python/qsharp-core/qsharp/clients/iqsharp.py
```diff
@@ -48,6 +48,8 @@ def display_raw(content):
 import logging
 logger = logging.getLogger(__name__)
 
+DEFAULT_TIMEOUT=120
```
cgranade, line 51: [nit]
```suggestion
DEFAULT_TIMEOUT = 120
```

cgranade (approved):

PR #545, opened 2021-11-16
Fix #543.

### LINKED ISSUE #543: The test `TestHintOnFailedMagic` fails in VS Test Explorer
In the latest `main` a couple of tests crash in VS Test Explorer (and likely in command line upon "dotnet test"), which results in these 2 tests and all the subsequent tests (totally 65) to stay "Not Run". The CI build does not notice the crash and the not run tests. Bettina says that such crashes typically happen upon `Debug.Assert()` failure (that is why she recommends to avoid using `Debug.Assert()` in C# code).
If we skip the 2 crashing tests and execute the subsequent tests, then we see that the test `TestHintOnFailedMagic` fails.

I'm making a change that removes the crashing code (no need to fix those crashes), but the failing test `TestHintOnFailedMagic` still fails.

**To Reproduce**
```
git clone https://github.com/microsoft/iqsharp.git

cd iqsharp

# (See details in https://github.com/microsoft/iqsharp#building-iq-from-source)
pushd src\Kernel
npm install
popd

# Open `iqsharp.sln` in VS
# <F7>   (Build Solution. Takes a few seconds)
# <Ctrl+e>, <t> or menu Test/"Test Explorer"    (Open the Test Explorer)
# Feel free to expand the tree of the tests - item "Tests.IQsharp"
# <Ctrl+r>, <v>  or Click top left button in Test Explorer toolbar (Run All Tests. Takes couple of minutes)
# Observe that the test "﻿Tests.IQsharp"/"﻿Tests.IQSharp"/﻿IQSharpEngineTests/OpenAliasedNamespaces stays "Not Run", 
  and all the tests after it either (totally 65).
# In the Output window (bottom of the screen) choose in the dropdown menu the item "Tests" and see:
    The active test run was aborted. Reason: Test host process crashed : Process terminated. Assertion Failed
    No Jupyter comms router was provided when this operation was registered. This is an internal error, and should never occur.
# Right-click the test OpenNamespaces (immediately after the crashed OpenAliasedNamespaces) and choose "Run". 
  Observe that it crashes too in the same manner.
# Select all subsequent tests, right-click, and choose "Run". They all succeed except TestHintOnFailedMagic.
```

**Expected behavior**
All tests succeed

**System information**
 - OS: Win

cgranade:


anjbur (approved):

src/Tests/IQsharpEngineTests.cs
```diff
@@ -713,6 +713,21 @@ await Assert.That
                 .UsingEngine()
                 .Input("%lsmagi")
                 .ExecutesWithError(containing:
+                    // NB: Since %attach is only present in local development
+                    //     mode, hints may in general change between debug
+                    //     and release configuration.
+                    #if DEBUG
```
kuzminrobin, line 719: Nit:
I don't know this par of C# yet, but if it is the same as a C/C++ preprocessor then consider 
`#ifdef DEBUG` or `#if defined(DEBUG)` (`ifdef` instead of `if`)
or a standard `#ifndef NDEBUG` or `#if !defined(NDEBUG)` (`NDEBUG` is a standard identifier in C/C++, as opposed to `DEBUG`).

kuzminrobin (approved):

src/Tests/IQsharpEngineTests.cs
cgranade: The syntax in C# is similar to, but not exactly the same as, that used by the C/C++ preprocessor. In particular, `#if` in C# is similar to `#ifdef` in C and C++: https://docs.microsoft.com/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

PR #573, opened 2022-01-20
Add support to print Job Cost Estimate Reporting upon job completion

vxfield:
When executing `%azure.jobs`, it nows print an additional "Cost Estimate" column with the cost estimated by the provider upon job completion.

> Note 1: The final cost on your bill might be slightly different due to added taxes and currency conversion rates. Click on the job to see how the cost was calculated.

> Note 2: Not all providers support this feature yet.

image


src/AzureClient/Mocks/MockCloudJob.cs
```diff
@@ -7,9 +7,89 @@
...
+        public string CurrencyCode
+        {
+            get => this.InternalCostEstimate.CurrencyCode;
+            set => this.InternalCostEstimate.SetReadOnlyProperty<CostEstimate>(nameof(CurrencyCode), value);
+        }
+
+        public IReadOnlyList<UsageEvent> Events
+        {
+            get => this.InternalCostEstimate.Events;
+        }
```
cgranade, line 39: [code style nit] 
```suggestion
        public IReadOnlyList<UsageEvent> Events => this.InternalCostEstimate.Events;
```

src/AzureClient/Mocks/MockCloudJob.cs
```diff
@@ -7,9 +7,89 @@
 using Azure.Quantum.Jobs.Models;
 using System;
 using System.IO;
+using System.Collections.Generic;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
+    public class MockCostEstimate
```
cgranade, line 23: It definitely makes me a bit uneasy how much of this uses reflection to bypass visibility modifiers. I understand how that's forced by the fact that the autogenerated classes in Azure.Quantum.Jobs are immutable and only have `internal` constructors, so I don't even have a better alternative to suggest, but it may be good to add some comments indicating why this is needed. Hopefully, we can file an upstream feature request to the code generator used as well?

src/AzureClient/Mocks/MockCloudJob.cs
```diff
@@ -7,9 +7,89 @@
...
+    {
+        public MockCostEstimate(string currencyCode, IReadOnlyList<UsageEvent> events, float? estimatedTotal)
+        {
+            this.InternalCostEstimate = MockHelper.CreateWithNonPublicConstructor<CostEstimate>(currencyCode, events, estimatedTotal);
+        }
+
+        public string CurrencyCode
+        {
+            get => this.InternalCostEstimate.CurrencyCode;
+            set => this.InternalCostEstimate.SetReadOnlyProperty<CostEstimate>(nameof(CurrencyCode), value);
```
cgranade, line 33: Minor code style nit, but can `<CostEstimate>` elided here? It looks like that's the receiver type for an extension method, which normally can be inferred as a type parameter.

src/AzureClient/Mocks/MockCloudJob.cs
```diff
@@ -7,9 +7,89 @@
...
+
+        private MockCostEstimate? costEstimate;
+        public new MockCostEstimate? CostEstimate
+        {
+            get => this.costEstimate;
+            set
+            {
+                this.costEstimate = value;
+                this.SetReadOnlyProperty<JobDetails>(nameof(CostEstimate), value?.InternalCostEstimate);
+            }
```
cgranade, line 76: For eventual .NET 6, we may want to consider using the new `field` keyword in C# 10 to simplify this some.

```csharp
        public new MockCostEstimate? CostEstimate
        {
            get;
            set
            {
                field = value;
                this.SetReadOnlyProperty<JobDetails>(nameof(CostEstimate), value?.InternalCostEstimate);
            }
        }
```

src/AzureClient/Mocks/MockHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+            var field = typeof(T).GetField($"<{propertyName}>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic)
+                        ?? typeof(T).GetField($"_{propertyName}", BindingFlags.Instance | BindingFlags.NonPublic)
+                        ?? typeof(T).GetField($"_{propertyName.FirstCharToLowerCase()}", BindingFlags.Instance | BindingFlags.NonPublic)
+                        ?? typeof(T).GetField($"{propertyName.FirstCharToLowerCase()}", BindingFlags.Instance | BindingFlags.NonPublic);
+            if (field != null)
+            {
+                field.SetValue(obj, value);
+                return;
+            }
+            var fields = typeof(T).GetFields(BindingFlags.Instance | BindingFlags.NonPublic);
```
cgranade, line 54: Is this variable used?

src/AzureClient/Mocks/MockHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            throw new NotSupportedException($"No setter or internal field found for property {propertyName} in {typeof(T).FullName}.");
+        }
+
+        public static T CreateWithNonPublicConstructor<T>(params object?[] parameters) =>
+            typeof(T).GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance)
+                     .Where((c) => c.GetParameters().Length == parameters.Length)
+                     .Select((c) => (T)(c.Invoke(parameters)))
+                     .FirstOrDefault() ?? throw new NotSupportedException($"No constructor found for type {typeof(T).FullName} accepting {parameters.Length} parameters.");
```
cgranade, line 63: This may have unintended consequences if two or more overloads in a method group have the same number of parameters; perhaps it may be reasonable to use `SingleOrDefault` instead so as to raise an exception in that case?

src/AzureClient/Mocks/MockHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
```
cgranade, line 65: [nit]
```suggestion
}

```

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -51,9 +52,63 @@ internal static class CloudJobExtensions
...
                     ("End Execution Time", cloudJob => cloudJob.Details.EndExecutionTime?.ToString() ?? string.Empty),
+                    ("Cost Estimate", cloudJob => cloudJob.GetCostEstimateText()),
                 },
                 Rows = jobsList.OrderByDescending(job => job.Details.CreationTime).ToList(),
             };
+
+        internal static string GetCostEstimateText(this CloudJob cloudJob) =>
+            cloudJob?.Details?.CostEstimate == null ? String.Empty :
+                                                      CurrencyHelper.FormatValue(cloudJob.Details.CostEstimate?.CurrencyCode,
+                                                                                 cloudJob.Details.CostEstimate?.EstimatedTotal);
```
cgranade, line 63: [nit]
```suggestion
            cloudJob?.Details?.CostEstimate == null
            ? String.Empty
            : CurrencyHelper.FormatValue(cloudJob.Details.CostEstimate?.CurrencyCode,
                                         cloudJob.Details.CostEstimate?.EstimatedTotal);
```

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -51,9 +52,63 @@ internal static class CloudJobExtensions
...
+
+        public static CultureInfo? GetCultureInfoForCurrencyCode(string currencyCode)
+        {
+            if (currencyCodeToCultureInfo.TryGetValue(currencyCode, out var cultureInfo))
+            {
+                return cultureInfo;
+            }
+
+            return null;
+        }
```
cgranade, line 91:
```suggestion
        public static CultureInfo? GetCultureInfoForCurrencyCode(string currencyCode) =>
            currencyCodeToCultureInfo.TryGetValue(currencyCode, out var cultureInfo)
            ? cultureInfo
            : (CultureInfo?) null;
```

src/AzureClient/Visualization/CloudJobEncoders.cs
```diff
@@ -51,9 +52,63 @@ internal static class CloudJobExtensions
...
+
+        public static string FormatValue(string? currencyCode, float? value)
```
cgranade, line 93: Depending on the scale of the particular currency, `float` can be numerically unstable. For currency values, `Decimal` may be preferable, especially if making equality comparisons.

src/Tests/AzureClientTests.cs
```diff
@@ -381,5 +382,77 @@ public void TestConnectedEvent()
                 Assert.AreEqual(true, lastArgs.UseCustomStorage);
             }
         }
+
+        [TestMethod]
+        public void TestCloudJobExtensions()
+        {
+            const string jobId = "myjobid";
+            const string jobName = "myjobname";
+            JobStatus jobStatus = JobStatus.Succeeded;
```
cgranade, line 400:
```suggestion
            var jobStatus = JobStatus.Succeeded;
```

src/Tests/AzureClientTests.cs
```diff
@@ -381,5 +382,77 @@ public void TestConnectedEvent()
...
+            const string jobProviderId = "microsoft";
+            const string jobTarget = "microsoft.paralleltempering-parameterfree.cpu";
+            DateTimeOffset jobCreationTime =  new DateTimeOffset(2021, 08, 12, 01, 02, 03, TimeSpan.Zero);
+            DateTimeOffset jobBeginExecutionTime =  new DateTimeOffset(2021, 08, 12, 02, 02, 03, TimeSpan.Zero);
+            DateTimeOffset jobEndExecutionTime =  new DateTimeOffset(2021, 08, 12, 03, 02, 03, TimeSpan.Zero);
```
cgranade, line 405:
```suggestion
            var jobCreationTime =  new DateTimeOffset(2021, 08, 12, 01, 02, 03, TimeSpan.Zero);
            var jobBeginExecutionTime =  new DateTimeOffset(2021, 08, 12, 02, 02, 03, TimeSpan.Zero);
            var jobEndExecutionTime =  new DateTimeOffset(2021, 08, 12, 03, 02, 03, TimeSpan.Zero);
```

src/Tests/AzureClientTests.cs
```diff
@@ -381,5 +382,77 @@ public void TestConnectedEvent()
...
+            var costEstimate = new MockCostEstimate("USD", new List<UsageEvent>(), 123.45f);
+            string costEstimateString = 123.45f.ToString("C", CurrencyHelper.GetCultureInfoForCurrencyCode("USD"));
```
cgranade, line 407:
```suggestion
            var costEstimateString = 123.45f.ToString("C", CurrencyHelper.GetCultureInfoForCurrencyCode("USD"));
```

src/Tests/AzureClientTests.cs
```diff
@@ -381,5 +382,77 @@ public void TestConnectedEvent()
...
+                ("End Execution Time", jobEndExecutionTime.ToString()),
+                ("Cost Estimate", costEstimateString),
+            };
+            Assert.AreEqual(cloudJobs.Count, table.Rows.Count);
+            Assert.AreEqual(expectedValues.Count, table.Columns.Count);
+            for (int columnIndex = 0; columnIndex < expectedValues.Count; columnIndex++)
+            {
+                Assert.AreEqual(expectedValues[columnIndex].Item1, table.Columns[columnIndex].Item1);
+                Assert.AreEqual(expectedValues[columnIndex].Item2, table.Columns[columnIndex].Item2(cloudJob));
+            }
```
cgranade, line 464:
```suggestion
            foreach ((var expected, var actual) in Enumerable.Zip(expectedValues, table.Column))
            {
                Assert.AreEqual(expected.Item1, actual.Item1);
                Assert.AreEqual(expected.Item2, actual.Item2(cloudJob));
            }
```

cgranade (approved):

src/AzureClient/Mocks/MockCloudJob.cs
vxfield: In this case it cannot be inferred because it would pick this same implementation of the setter and create a infinite recursion. We want to pick the implementation of the base class.

src/AzureClient/Visualization/CloudJobEncoders.cs
vxfield: The data model is a float, so I'll keep as such for now.

src/AzureClient/Mocks/MockHelper.cs
vxfield: No. Left over from debugging. Will remove it.

src/AzureClient/Mocks/MockHelper.cs
vxfield: Will make the change.

src/AzureClient/Mocks/MockCloudJob.cs
vxfield: Since the unit test project is still targeting `netcoreapp3.1`, I'll leave it like this for now.

PR #589, opened 2022-02-15
Update to 0.22.192090-beta.

cgranade:
It appears as though some of the packages used during builds have aged out of the alpha feed, breaking the per-repo build (e.g.: #588). This PR updates package versions to 0.22.192090-beta in order to restore the build.

### DIFF
--- a/images/iqsharp-base/Dockerfile
+++ b/images/iqsharp-base/Dockerfile
@@ -1,4 +1,4 @@
-0.22.192090-beta0.22.192090-beta# The same image used by mybinder.org
+# The same image used by mybinder.org
 FROM python:3.7-slim-buster
 
 # install the notebook package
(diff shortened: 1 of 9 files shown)


images/iqsharp-base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# The same image used by mybinder.org
+0.22.192090-beta0.22.192090-beta# The same image used by mybinder.org
```
swernli, line 1: This seems like a mistaken replacement... I've seen this happen if the global find-replace is used in the GitHub web editor. It adds the replacement text to the top of the file instead of the right location in the middle of the file. This appears to have happened to a few of the files in here. Unfortunately that means this change will have to be done locally with global find-replace or manually in the web editor.

swernli (changes requested):

images/iqsharp-base/Dockerfile
cgranade: Absolutely is, thank you for catching that! Let me go fix that, then.

swernli (approved):

PR #592, opened 2022-02-18
Quantinuum rebrand

cgranade:
This draft PR begins changes needed to update Quantinuum targets to reflect their recent rebranding.


src/Tests/AzureClientTests.cs
```diff
@@ -150,7 +156,7 @@ public void TestManualTargets()
             // set up the mock workspace
             var azureWorkspace = azureClient.ActiveWorkspace as MockAzureWorkspace;
             Assert.IsNotNull(azureWorkspace);
-            azureWorkspace?.AddProviders("ionq", "honeywell", "unrecognized");
+            azureWorkspace?.AddProviders("ionq", "quantinuum", "unrecognized");
```
vxfield, line 159: should we keep `honeywell` scenario here to make sure it continues to work?

src/Tests/IQsharpEngineTests.cs
```diff
@@ -931,7 +931,7 @@ public async Task CompileAndSubmitWithClassicalControl()
                 .Input(source)
                     .ExecutesSuccessfully()
                 .WithMockAzure()
-                .Input("%azure.target honeywell.mock")
+                .Input("%azure.target quantinuum.mock")
```
vxfield, line 934: should we keep `honeywell` scenario here to make sure it continues to work?

src/Tests/TestExtensions.cs
```diff
@@ -41,7 +41,7 @@ await engine
             var client = await (await engine).Engine.GetEngineService<IAzureClient>();
             if (client is AzureClient azureClient && azureClient.ActiveWorkspace is MockAzureWorkspace workspace)
             {
-                workspace.AddProviders("ionq", "honeywell");
+                workspace.AddProviders("ionq", "quantinuum");
```
vxfield, line 44: should we keep `honeywell` scenario here to make sure it continues to work?

vxfield (commented):
Looking good so far.

Is there any place in IQ# in which input/output-format values are used?
To recap, the Quantinuum input/output-format values will be the same as Honeywell.

src/Tests/AzureClientTests.cs
cgranade: I believe that's covered in other tests, while this one makes sure that IonQ works when Quantinuum neé Honeywell is also added. That said, happy to add both providers here to be sure to cover that case when using IonQ as well.

src/Tests/IQsharpEngineTests.cs
cgranade: Fixed.

src/Tests/TestExtensions.cs
cgranade: Fixed.

cgranade:
Thanks for all your help, @cesarzc, and for your review @vxfield! The PR now works locally, so moving this out of draft and marking as ready to review.

image

cesarzc (approved):
LGTM!

PR #597, opened 2022-02-23
Extend testing CI to re-run conda tests in self-contained environment

IsraelMiles:
Testing CI runs conda test on the build agent host. This can make it difficult to find and diagnose issues caused by missing dependencies. For example the build agent host can inadvertently provide dependencies that should be provided by conda packages themselves. In addition to all existing tests, conda test should be re-executed from within a brand-new Docker container, in order to emulate a minimal environment as may be installed by users.


build/e2e.yml
```diff
@@ -24,7 +24,7 @@ resources:
     type: github
     endpoint: github
     name: microsoft/qdk
-    ref: refs/heads/main
+    ref: refs/heads/imiles/task-36357
```
cgranade, line 27: This looks like it should be changed back to `main`?

build/ci.yml
```diff
@@ -48,26 +49,41 @@ jobs:
...
   - template: steps-conda.yml
   condition: ne(variables['Enable.Conda'], 'false')
 
-- job: "pack_conda_win"
+- job: "pack_conda_windows"
   dependsOn: "pack_selfcontained"
   pool:
     vmImage: 'windows-2019'
+  container:
+    image: qdkimages.azurecr.io/internal/quantum/windows-test:latest
```
cgranade, line 82: We may want a slightly more descriptive name than `windows-test` here. It'd also be good to pick this up from the internal MCR (`msint.azurecr.io`) and not from our ACR instance that feeds into it (`qdkimages.azurecr.io`). To do so, we'll likely need a matching PR against the microsoft/mcr repo.

build/steps-conda.yml
```diff
@@ -12,29 +11,78 @@ steps:
...
+    pwsh ./pack-conda.ps1
+  displayName: "Packing IQ# packages"
+  workingDirectory: '$(System.DefaultWorkingDirectory)/build'
+  condition: and(succeeded(), ne(variables['Skip.Tests'], 'true'), or(eq(variables.OS, 'linux'), eq(variables.OS, 'mac')))
+
+- pwsh: |
+    (& "$(CondaPath)/conda.exe" "shell.powershell" "hook") | Out-String | Invoke-Expression
+    conda activate conda_env
+    conda info
+    pwsh -NoProfile -Command ./pack-conda.ps1
```
cgranade, line 61: A comment here as to why we're calling pwsh from pwsh may be helpful.

build/steps-conda.yml
```diff
@@ -12,29 +11,78 @@ steps:
...
 ##
-# Pack conda packages
+# Pack conda packages for each OS
 ##
-- pwsh: .\pack-conda.ps1
+- script: |
+    eval "$(CondaShellHook)"
+    conda activate conda_env
+    conda info
+    pwsh ./pack-conda.ps1
```
cgranade, line 52:
```suggestion
    pwsh -NoProfile ./pack-conda.ps1
```

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG CONDA_VER=latest
```
cgranade, line 1: A readme or similar in this directory could help explain what each of these three new Dockerfiles is used for.

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM ubuntu:latest
+
+RUN apt-get update && apt-get install -yq curl wget jq
+
+ARG CONDA_VER
+ARG OS_TYPE
+# Install miniconda to /miniconda
+RUN curl -LO "http://repo.continuum.io/miniconda/Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh"
+RUN bash Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh -p /miniconda -b
+RUN rm Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh
```
cgranade, line 15: Combining these three steps in a single layer can significantly reduce layer sizes (and hence the size of the final image) by preventing temporary files that are removed in later steps from being committed as layers. That can help speed up pipelines that depend on this image by reducing the amount that those builds need to download.
```suggestion
RUN curl -LO "http://repo.continuum.io/miniconda/Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh" && \
    bash Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh -p /miniconda -b && \
    rm Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh
```

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ENV PATH=/miniconda/bin:${PATH}
```
cgranade, line 16: Should this be on `PATH` since you're using the shell hooks later to enable conda? As is, this could cause binaries from the base environment to leak through to other environments later.

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG CONDA_VER=latest
+ARG UBUNTU_VER=20.04
+ARG OS_TYPE=x86_64
+ARG PY_VER=3.7
+
+FROM ubuntu:latest
```
cgranade, line 6: You may want to reuse `UBUNTU_VER` here.

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG CONDA_VER
+ARG OS_TYPE
+# Install miniconda to /miniconda
+RUN curl -LO "http://repo.continuum.io/miniconda/Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh"
+RUN bash Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh -p /miniconda -b
+RUN rm Miniconda3-${CONDA_VER}-Linux-${OS_TYPE}.sh
+ENV PATH=/miniconda/bin:${PATH}
+RUN conda update -y conda
+
+ARG PY_VER
```
cgranade, line 19: A default here may be useful.

images/test-environments/linux/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN wget -q "https://packages.microsoft.com/config/ubuntu/${UBUNTU_VER}/packages-microsoft-prod.deb"
+# Register the Microsoft repository GPG keys
+RUN dpkg -i packages-microsoft-prod.deb
+
+# Update the list of packages after we added packages.microsoft.com
+RUN apt-get update
+
+# Install powershell and git
+RUN apt-get install -y powershell
+RUN apt-get install -y git
```
cgranade, line 34: Generally, when installing from dpkg and apt, it's really critical to do so in a single layer that also removes the apt-get cache. Otherwise, you can wind up committing several GB to the image by accident. There should be an example in the iqsharp-base Dockerfile.

cgranade (changes requested):
Looking really good! I think there's a few more things that would be nice to handle before merging to main, but this is really coming along — thanks for tagging me in!

build/e2e.yml
IsraelMiles: I was referencing my separate branch of the qdk which is also using containers for the e2e builds. Once this PR looks good I will change this back to main.

images/test-environments/linux/Dockerfile
IsraelMiles: Great catch! I'll remove this line then.

images/test-environments/linux/Dockerfile
IsraelMiles: Great tip, I've updated accordingly!

build/ci.yml
IsraelMiles: I've updated the images to `linux/windows-selfcontained`. I also had the MCR PR approved below:
https://github.com/microsoft/mcr/pull/1871

Lastly I've updated the `push-selfcontained.yml` workflow to run for both linux and windows images.

.github/workflows/push-selfcontained.yml
```diff
@@ -0,0 +1,90 @@
+name: Push self-contained images on Dockerfile change
+
+on:
+  push:
+    branches:
+    - main
```
cgranade, line 6:
```suggestion
    - main
    paths:
    - 'images/test-environments/**/Dockerfile`
```

I think you can filter this to run only on those pushes that update the self-contained images: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows?ref=hackernoon.com&msclkid=e30890edcb2411ecb18b3cc68805c075#running-your-workflow-only-when-a-push-affects-specific-files

.github/workflows/push-selfcontained.yml
```diff
@@ -0,0 +1,90 @@
...
+    - uses: actions/checkout@v1
+    - name: Azure Login
+      uses: Azure/login@v1
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+    - name: Docker Login
+      run: az acr login -n qdkimages
+    - name: Build and push Docker image
+      run: |
+        $Now = [DateTime]::Now;
```
cgranade, line 22: It looks like `$Now` is no longer used?
```suggestion
```

.github/workflows/push-selfcontained.yml
```diff
@@ -0,0 +1,90 @@
...
+        docker tag $WindowsLocalTag "${WindowsRemoteRepo}:${ImageTag}-windows"
+
+        docker push "${LinuxRemoteRepo}:${ImageTag}-linux"
+        docker push "${WindowsRemoteRepo}:${ImageTag}-windows"
+
+        docker tag $LinuxLocalTag "${LinuxRemoteRepo}:latest"
+        docker tag $WindowsLocalTag "${WindowsRemoteRepo}:latest"
+
+        docker push "${LinuxRemoteRepo}:latest"
+        docker push "${WindowsRemoteRepo}:latest"
```
cgranade, line 47: Nit, but it may be helpful to abstract this into a function to reduce the likelihood of copy-paste errors when this workflow is next modified:
```suggestion
        function Update-Image() {
            param(
                [string] $Path
                [string] $RepoName,
                [string] $LocalTag,
            );
            $fullRepoName = "public/quantum/$RepoName";
            $remoteRepo = "${{ secrets.ACR_REGISTRY }}/${RepoName}";

            docker build $Path --tag $LocalTag;

            docker tag $LocalTag "${remoteRepo}:${LocalTag}";
            docker push "${remoteRepo}:${LocalTag}";

            docker tag $LocalTag "${remoteRepo}:latest";
            docker push "${remoteRepo}:latest";
        }
        Update-Image `
            -RepoName "linux-selfcontained" `
            -LocalTag "${ImageTag}-linux" `
            -Path images/test-environments/linux;
            
        Update-Image `
            -RepoName "windows-selfcontained" `
            -LocalTag "${ImageTag}-windows" `
            -Path images/test-environments/windows;
```

build/ci.yml
```diff
@@ -48,26 +49,42 @@ jobs:
...
   condition: ne(variables['Enable.Conda'], 'false')
 
-- job: "pack_conda_macos"
+- job: "pack_conda_linux"
   dependsOn: "pack_selfcontained"
   pool:
-    vmImage: 'macOS-latest'
+    vmImage: 'ubuntu-latest'
+  container:
+    image: qdkimages.azurecr.io/public/quantum/linux-selfcontained:latest
```
cgranade, line 68: For reliability, it would be good to move this over to `mcr.microsoft.com` (since `public`).

cgranade (approved):

.github/workflows/push-selfcontained.yml
IsraelMiles: Great tip, thank you for refactoring that!

PR #601, opened 2022-03-01
Initial Support for microsoft.simulator Targets on Jupyter Notebooks

ScottCarda-MS:
Allow the `%azure.target` to be set to `microsoft.simulator.fullstate` in notebooks. This allows QIR generated from notebooks to be submitted to the full state simulator on azure quantum.


src/AzureClient/Visualization/HistogramEncoders.cs
```diff
@@ -20,9 +20,35 @@ internal class Histogram : Dictionary<string, double>
...
+        internal static Histogram ToHistogram(this Stream stream, ILogger? logger = null, bool isSimulatorOutput = false)
         {
             var output = new StreamReader(stream).ReadToEnd();
+            if (isSimulatorOutput)
+            {
+                // This routine seems to be using what I think may be an older format
+                // (the az quantum cli had handling for both):  {"Histogram":["0",0.5,"1",0.5]}
+                output = "{ \"Histogram\" : [ \"" + output.Trim() + "\", 1.0 ] }";
+
+                // TODO: Make this more general. The corresponding implementation from the az quantum cli is:
```
ScottCarda-MS, line 32: This code is still a work progress and will be addressed in an upcoming PR that will update the output format of jobs submitted to the full state simulator. Until then this comment out section should remain.

src/Python/qsharp-core/qsharp/loader.py
```diff
@@ -114,6 +127,22 @@ def trace(self, **kwargs) -> Any:
         """
         return qsharp.client.trace(self, **kwargs)
 
+    def as_qir(self) -> pqp.QirModule:
```
cgranade, line 130: This was a fairly rough stab on my part for what the final API should look like; before merging to `main`, I think it might make sense to use a `format` keyword argument that allows selecting between `*.ll`, bitcode, and a PyQIR module return values.

src/Tests/AzureClientEntryPointTests.cs
```diff
@@ -81,6 +82,94 @@ public async Task FromSnippet()
...
+        [TestMethod]
+        public async Task ValidParameterTypes()
+        {
+            var entryPointGenerator = Init("Workspace", new string[] { SNIPPETS.ValidParameterTypes });
+            var entryPoint = entryPointGenerator.Generate("ValidParameterTypes", null);
+
+            Assert.IsNotNull(entryPoint);
+
+            var validArguments = new List<(Argument, string)>()
+            {
```
cgranade, line 107: You may be able to use dictionary initializer syntax here to make this test case a bit simpler; `validArguments.Select(x => X.Item1)` below can then be replaced by `validArguments.Keys`.

src/Tests/SNIPPETS.cs
```diff
@@ -365,5 +361,49 @@ operation DependsOnChemistryWorkspace() : ((JordanWignerEncodingData, Int, Doubl
     }
 ";
 
+        public static string ValidParameterTypes =
+@"
+    operation ValidParameterTypes(
```
cgranade, line 366: Please make sure that these snippets follow Q# style guide as best as reasonable. In particular, please use verbs for operation names. (The above case should do so as well, to be fair.)

cgranade (commented):
Looks like it's really coming along, thanks for tagging me in to review! Left some initial comments, and happy to expand or collaborate as it's helpful. Thanks!

src/AzureClient/Visualization/HistogramEncoders.cs
ScottCarda-MS: I've added a link to https://github.com/microsoft/iqsharp/issues/607.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -73,12 +81,78 @@ public Task<IQuantumMachineJob> SubmitAsync(IQuantumMachine machine, AzureSubmis
...
+        {
+            var argumentList = new List<Argument>();
+            foreach (var parameter in OperationInfo.RoslynParameters)
+            {
+                if (!submissionContext.InputParameters.ContainsKey(parameter.Name))
+                {
+                    throw new ArgumentException($"Required parameter {parameter.Name} was not specified.");
+                }
+
+                string rawParameterValue = submissionContext.InputParameters[parameter.Name];
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/AzureClient/EntryPoint/IEntryPointGenerator.cs
```diff
@@ -40,8 +40,9 @@ public interface IEntryPointGenerator
         /// <param name="operationName">The name of the operation to wrap in an entry point.</param>
         /// <param name="executionTarget">The intended execution target for the compiled entry point.</param>
         /// <param name="runtimeCapabilities">The runtime capabilities of the intended execution target.</param>
+        /// <param name="forceQir">When true, uses QIR to generate the entry point.</param>
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/AzureClient/Mocks/MockQIRSubmitter.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        private bool IsArgumentValueEqual(ArgumentValue fst, ArgumentValue snd)
+        {
+            if (fst.Type != snd.Type)
+            {
+                return false;
+            }
+
+            if (fst is ArgumentValue.Bool fstBool && snd is ArgumentValue.Bool sndBool)
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/AzureClient/Mocks/MockQIRSubmitter.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        private bool IsEqualToExpected(IReadOnlyList<Argument> arguments)
+        {
+            if (this.ExpectedArguments.Count != arguments.Count)
+            {
+                return false;
+            }
+
+            for (int i = 0; i < this.ExpectedArguments.Count; i++)
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: done

src/Tests/SNIPPETS.cs
ScottCarda-MS: done

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -24,6 +26,9 @@ internal class EntryPoint : IEntryPoint
         private OperationInfo OperationInfo { get; }
         private ILogger? Logger { get; }
 
+        /// <inheritdoc/>
+        public Stream QirStream { get; }
```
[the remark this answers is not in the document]
ScottCarda-MS: It has been made nullable.

src/Kernel/Magic/QirMagic.cs
```diff
@@ -0,0 +1,106 @@
...
+                return AzureClientError.InvalidEntryPoint.ToExecutionResult();
+            }
+
+            if (entryPoint == null)
+            {
+                return "Internal error: generated entry point was null, but no compilation errors were returned."
+                    .ToExecutionResult(ExecuteStatus.Error);
+            }
+
+            if (entryPoint.QirStream == null)
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -34,18 +39,21 @@ internal class EntryPoint : IEntryPoint
         /// <param name="outputType">Specifies the output parameter type for the
         /// <see cref="EntryPointInfo{I,O}"/> object provided as the <c>entryPointInfo</c> argument.</param>
         /// <param name="operationInfo">Information about the Q# operation to be used as the entry point.</param>
+        /// <param name="qirStream">
+        ///     Stream from which QIR bitcode for the entry point can be read.
+        /// </param>
         /// <param name="logger">Logger used to report internal diagnostics.</param>
-        public EntryPoint(object entryPointInfo, Type inputType, Type outputType, OperationInfo operationInfo, ILogger? logger)
+        public EntryPoint(object entryPointInfo, Type inputType, Type outputType, OperationInfo operationInfo, Stream qirStream, ILogger? logger)
```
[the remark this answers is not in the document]
ScottCarda-MS: It has been made nullable here.

src/Core/AssemblyInfo.cs
```diff
@@ -25,19 +26,36 @@ public class AssemblyInfo : IEquatable<AssemblyInfo>
...
         /// </summary>
-        public AssemblyInfo(Assembly assembly, string location, QsNamespace[] syntaxTree)
+        /// <remarks>
+        ///     If <paramref name="qirBitcode"/> is not <c>null</c>, its
+        ///     contents will be read by this constructor, and its position
+        ///     will not be reset. That is, the position of <paramref name="qirBitcode" />
+        ///     after this call will be that set by the <see cref="System.IO.Stream.CopyTo(Stream)"/>
+        ///     method (generally, the end of the QIR bitcode stream).
+        /// </remarks>
+        public AssemblyInfo(Assembly assembly, string location, QsNamespace[] syntaxTree, Stream qirBitcode)
```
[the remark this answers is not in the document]
ScottCarda-MS: This file doesn't use the `#nullable enable` directive. I'm not sure it is worth it to add the directive either as most everything in this file is expected to be nullable.

src/AzureClient/AzureExecutionTarget.cs
```diff
@@ -64,7 +68,7 @@ protected AzureExecutionTarget(string? targetId)
         ///     Valid target IDs are structured as "provider.target".
         ///     For example, "ionq.simulator" or "honeywell.qpu".
         /// </remarks>
-        protected static AzureProvider? GetProvider(string? targetId)
+        protected internal static AzureProvider? GetProvider(string? targetId)
```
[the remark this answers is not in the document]
ScottCarda-MS: Having this method as `protected internal` instead of `protected` is what allows [this line](https://github.com/microsoft/iqsharp/blob/78ecc32e904db972078416b0589e31cddd4467d4/src/AzureClient/AzureClient.cs#L47) to reference the function.

src/AzureClient/AzureClient.cs
```diff
@@ -39,7 +41,10 @@ public class AzureClient : IAzureClient
         private AzureExecutionTarget? ActiveTarget { get; set; }
         private string MostRecentJobId { get; set; } = string.Empty;
         private IEnumerable<ProviderStatusInfo>? AvailableProviders { get; set; }
-        private IEnumerable<TargetStatusInfo>? AvailableTargets => AvailableProviders?.SelectMany(provider => provider.Targets);
+        private IEnumerable<TargetStatusInfo>? AvailableTargets => 
+            AvailableProviders?
+            .SelectMany(provider => provider.Targets)
+            .Where(t => t.TargetId != null && (t.TargetId.StartsWith(MicrosoftSimulator) || AzureExecutionTarget.GetProvider(t.TargetId) != AzureProvider.Microsoft));
```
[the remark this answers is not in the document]
ScottCarda-MS: It didn't raise any warnings regarding this, but it does seem strange. The `?` for the `SelectMany` is on the previous line, but there isn't one for the `Where`. I've moved the `?` from the first line to be in front of the `SelectMany` and added one in front of the `Where`.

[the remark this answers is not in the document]
ScottCarda-MS: I wonder if Linq is smart enough to be able to handle null references? It doesn't look like it ever complains about Linq methods on nullable objects.

src/AzureClient/AzureClient.cs
```diff
@@ -543,7 +567,10 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string j
                 //                 cancellation token support.
                 var request = WebRequest.Create(job.OutputDataUri);
                 using var responseStream = (await request.GetResponseAsync()).GetResponseStream();
-                return responseStream.ToHistogram(Logger).ToExecutionResult();
+                return responseStream.ToHistogram(
```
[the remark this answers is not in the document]
ScottCarda-MS: Yes, @bettinaheim said that she wants a histogram with a single record for the return value of the entry point as the output of a full-state simulator job.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -73,12 +81,78 @@ public Task<IQuantumMachineJob> SubmitAsync(IQuantumMachine machine, AzureSubmis
...
+            {
+                return new ArgumentValue.String(parameterValue);
+            }
+            else if (parameterType == typeof(Pauli))
+            {
+                return new ArgumentValue.Pauli(Newtonsoft.Json.JsonConvert.DeserializeObject<Pauli>(parameterValue));
+            }
+            else if (parameterType == typeof(Result))
+            {
+                return new ArgumentValue.Result(Newtonsoft.Json.JsonConvert.DeserializeObject<Result>(parameterValue));
```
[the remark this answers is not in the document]
ScottCarda-MS: I added the `!` operator to assert that the output of the deserialize must be non-null.

src/Python/qsharp-core/qsharp/loader.py
ScottCarda-MS: I'm fairly unfamiliar with the python code for these changes, as I inherited them from previous work. Would you be able to walk me through some of the changes you are thinking about for this sometime?

src/Python/qsharp-core/qsharp/loader.py
```diff
@@ -3,18 +3,31 @@
...
 from types import ModuleType, new_class
 import importlib
 from importlib.abc import MetaPathFinder, Loader
+import tempfile as tf
+
 import qsharp
 
 from typing import Iterable, List, Optional, Any, Dict, Tuple
 
+try:
```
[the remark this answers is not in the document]
ScottCarda-MS: Could you direct me on what needs to be done for this?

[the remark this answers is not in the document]
cgranade: That will depend slightly on what API PyQIR goes live with once https://github.com/qir-alliance/pyqir/issues/34 is resolved, but the long story short is that this was developed using 0.1a1, and will need to be updated to use the current version of PyQIR.

src/Python/qsharp-core/qsharp/loader.py
cgranade: Happy to schedule some time offline, but my thought was an API that's a little closer to that used in https://github.com/qutip/qutip-qip/pull/125.

src/AzureClient/AzureClient.cs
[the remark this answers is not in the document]
cgranade: I think that would be a pretty surprising and counter-intuitive user experience for Q# notebook users especially, in that running an operation locally and via the full-state simulator would then give very different visualizations as a result.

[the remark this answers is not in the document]
ScottCarda-MS: I couldn't agree more, but we've decided for now to go forward with it like this. Perhaps before merging into main we can revisit this topic.

src/AzureClient/AzureClient.cs
```diff
@@ -346,33 +341,62 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel? channel, C
...
+                    // We should never get here, since ActiveTarget should have already been validated at the time it was set.
+                    channel?.Stderr($"Unexpected error while preparing job for execution on target {ActiveTarget.TargetId}.");
+                    return AzureClientError.InvalidTarget.ToExecutionResult();
+                }
+
+                var job = await entryPoint.SubmitAsync(submitter, submissionContext);
+                channel?.Stdout($"Job successfully submitted.");
                 channel?.Stdout($"   Job name: {submissionContext.FriendlyName}");
                 channel?.Stdout($"   Job ID: {job.Id}");
                 MostRecentJobId = job.Id;
```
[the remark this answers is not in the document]
ScottCarda-MS: They don't share a base type. I'm not sure how to refactor the code so that there is less logic between the two of these cases.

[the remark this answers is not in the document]
cgranade: I was able to substantially reduce the duplication, and would be happy to commit that to your branch if that would be helpful.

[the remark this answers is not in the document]
ScottCarda-MS: Sure, please do!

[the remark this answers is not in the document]
ScottCarda-MS: I see, thanks! Those async tasks always confuse me.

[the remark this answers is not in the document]
cgranade: No worries, happy to help! One thing to note here is that exceptions are thrown at the point that a task is awaited — simply launching the task doesn't itself throw. That lets us consolidate error handling at the point where we `await`. More details at https://docs.microsoft.com/dotnet/csharp/programming-guide/concepts/async/#asynchronous-exceptions.

src/Python/qsharp-core/qsharp/loader.py
ScottCarda-MS: I believe we've now made the needed changes together.

src/AzureClient/AzureClient.cs
```diff
@@ -39,7 +41,10 @@ public class AzureClient : IAzureClient
         private AzureExecutionTarget? ActiveTarget { get; set; }
         private string MostRecentJobId { get; set; } = string.Empty;
         private IEnumerable<ProviderStatusInfo>? AvailableProviders { get; set; }
-        private IEnumerable<TargetStatusInfo>? AvailableTargets => AvailableProviders?.SelectMany(provider => provider.Targets);
+        private IEnumerable<TargetStatusInfo>? AvailableTargets => 
+            AvailableProviders
+            ?.SelectMany(provider => provider.Targets)
+            ?.Where(t => t.TargetId != null && (t.TargetId.StartsWith(MicrosoftSimulator) || AzureExecutionTarget.GetProvider(t.TargetId) != AzureProvider.Microsoft));
```
cesarzc, line 47: I don't `t.TargetId.StartsWith(MicrosoftSimulator)` part in this condition is needed. Additionally, I believe it would be preferable to just use `AzureExecutionTarget.GetProvider` since that method should be able to get the right provider from the simulator target.

src/AzureClient/AzureClient.cs
```diff
@@ -26,6 +26,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
     /// <inheritdoc/>
     public class AzureClient : IAzureClient
     {
+        private const string MicrosoftSimulator = "microsoft.simulator";
```
cesarzc, line 29: As a consequence of this comment: https://github.com/microsoft/iqsharp/pull/601/files#r825100550, this member should no longer be needed.

src/AzureClient/AzureClient.cs
```diff
@@ -346,11 +341,41 @@ public async Task<ExecutionResult> GetConnectionStatusAsync(IChannel? channel, C
...
+                {
+                    var machine = AzureFactory.CreateMachine(this.ActiveWorkspace, this.ActiveTarget.TargetId, this.StorageConnectionString);
+                    if (machine == null)
+                    {
+                        // We should never get here, since ActiveTarget should have already been validated at the time it was set.
+                        channel?.Stderr($"Unexpected error while preparing job for execution on target {ActiveTarget.TargetId}.");
+                        return AzureClientError.InvalidTarget.ToExecutionResult();
+                    }
+                    jobTask = entryPoint.SubmitAsync(machine, submissionContext);
+                }
```
cesarzc, line 374: Since branching here depends on whether a `IQirSubmitter` or a `IQuantumMachine` object is available for the particular target, I believe it would be better to organize this part as:
```csharp
if (SubmitterFactory.QirSubmitter(this.ActiveTarget.TargetId, this.ActiveWorkspace, this.StorageConnectionString) is IQirSubmitter submitter)
{
    // Continue with submitter logic.
}
else if (AzureFactory.CreateMachine(this.ActiveWorkspace, this.ActiveTarget.TargetId, this.StorageConnectionString) is IQuantumMachine machine)
{
    // Continue with machine logic.
}
```

src/AzureClient/AzureClient.cs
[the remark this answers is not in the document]
cesarzc: Since how to handle the output will be addressed in another PR and the conversation on how to do that is still ongoing, we could leave this part as is for now (since it is going into a feature branch) with the expectation that it will be addressed before any of these changes are merged to main.

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -104,5 +178,29 @@ public Task<IQuantumMachineJob> SubmitAsync(IQuantumMachine machine, AzureSubmis
...
+                .GetMethods()
+                .Single(method =>
+                    method.Name == "SubmitAsync"
+                    && method.GetParameters().Length == 4
+                    && method.GetParameters()[0].ParameterType == typeof(Stream)
+                    && method.GetParameters()[1].ParameterType == typeof(string)
+                    && method.GetParameters()[2].ParameterType == typeof(IReadOnlyList<Argument>)
+                    && method.GetParameters()[3].ParameterType == typeof(SubmissionOptions)
+                    );
+            var submitParameters = new object[] { QirStream!, $"ENTRYPOINT__{submissionContext.OperationName}", entryPointInput, options };
```
cesarzc, line 202: I think the `"ENTRYPOINT"` string literal is used in other places in this code base. It would be good to make it a `static readonly string` somewhere and reference it instead of using a string literal.

src/AzureClient/AzureClient.cs
ScottCarda-MS: The condition here is actually confusing to me, and maybe that stems from my not understanding the relationship between the target and the provider. The condition is saying the target either needs to be 'microsoft.simulator' or the target needs to be *not* under the Microsoft provider. Isn't that all the targets? I don't know what this line of code is meant to do.

src/AzureClient/AzureClient.cs
ScottCarda-MS: I'm not quite following your suggested change. Why would we want to call into the SubmitterFactory if we don't want to deal with QIR submitters? Shouldn't the deciding logic be external to those calls (SubmitterFactory.QirSubmitter and AzureFactory.CreateMachine)? If we are calling a factory function we should already be expecting that it will return a valid object, right?

src/AzureClient/EntryPoint/EntryPointGenerator.cs
```diff
@@ -79,8 +79,9 @@ internal class EntryPointGenerator : IEntryPointGenerator
             return null;
         }
 
+        /// <inheritdoc/>
         public IEntryPoint Generate(string operationName, string? executionTarget,
-            RuntimeCapability? runtimeCapability = null)
+            RuntimeCapability? runtimeCapability = null, bool forceQir = false)
```
cesarzc, line 84: nit: the last parameter could be named `generateQir` instead of `forceQir`.

src/AzureClient/EntryPoint/IEntryPoint.cs
```diff
@@ -26,5 +25,19 @@ public interface IEntryPoint
...
+        /// <param name="submitter">The <see cref="IQirSubmitter"/> object representing the job submission target.</param>
+        /// <param name="submissionContext">The <see cref="AzureSubmissionContext"/> object representing the submission context for the job.</param>
+        /// /// <param name="cancellationToken">Cancellation token used to interrupt this submission.</param>
+        /// <returns>The details of the submitted job.</returns>
+        public Task<IQuantumMachineJob> SubmitAsync(IQirSubmitter submitter, AzureSubmissionContext submissionContext, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// The stream from which QIR bitcode for the entry point can be read.
+        /// </summary>
+        public Stream? QirStream { get; }
```
cesarzc, line 41: I am curious why `QirStream` needs to be part of this interface? Since `SubmitAsync` is implemented in the `EntryPoint` class it can access `QirStream` there without the need of including it in this interface too.

src/AzureClient/Extensions.cs
```diff
@@ -89,7 +87,11 @@ public static void AddAzureClient(this IServiceCollection services)
             (job.Details.Name != null && job.Details.Name.Contains(filter, StringComparison.OrdinalIgnoreCase)) ||
             (job.Details.Target != null && job.Details.Target.Contains(filter, StringComparison.OrdinalIgnoreCase));
 
-        internal static void Log(this Exception ex, IChannel? channel, ILogger? logger, string msg = "")
+        /// <summary>
+        ///      Writes an exception to a logger and emits it to the stderr stream
+        ///      of a Jupyter display channel.
+        /// </summary>
+        public static void Log(this Exception ex, IChannel? channel, ILogger? logger, string msg = "")
```
cesarzc, line 94: I noticed this changed from `internal` to `public`. Is this method now being referenced outside this project? If so, wouldn't it be better to make this implementation part of another project such as `Core`?

src/AzureClient/Mocks/MockQIRSubmitter.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Quantum.IQSharp.AzureClient
+{
+    internal class MockQIRSubmitter : IQirSubmitter
+    {
+        public string Target => throw new NotImplementedException();
+
+        private IReadOnlyList<Argument> ExpectedArguments = new List<Argument>();
```
cesarzc, line 20: Could you please add a documentation comment on the `ExpectedArguments` member explaining its purpose?

src/AzureClient/EntryPoint/EntryPoint.cs
ScottCarda-MS: Knowing where to put that is the tricky part 😄.

src/AzureClient/Mocks/MockQIRSubmitter.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                return false;
+            }
+
+            return this.ExpectedArguments
+                .Zip(arguments, (fst, snd) => (fst, snd))
+                .All(tup => tup.fst.Name == tup.snd.Name && IsArgumentValueEqual(tup.fst.Value, tup.snd.Value));
+        }
+
+        public Task<IQuantumMachineJob> SubmitAsync(Stream qir, string entryPoint, IReadOnlyList<Argument> arguments, SubmissionOptions options)
```
cesarzc, line 58: Could you please add a documentation comment to this method explaining that its purpose is to validate that the arguments it receives are the expected ones?

src/Core/Compiler/CompilerService.cs
```diff
@@ -141,15 +141,16 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
                 LoadReferencesBasedOnGeneratedCsharp = false, // deserialization of resources in references is only needed if there is an execution target
                 IsExecutable = compileAsExecutable,
                 AssemblyConstants = new Dictionary<string, string> { [AssemblyConstants.ProcessorArchitecture] = executionTarget ?? string.Empty },
-                RuntimeCapability = runtimeCapability ?? RuntimeCapability.FullComputation
+                RuntimeCapability = runtimeCapability ?? RuntimeCapability.FullComputation,
+                PrepareQirGeneration = executionTarget?.StartsWith("microsoft.simulator") ?? false
```
cesarzc, line 145: This should be abstracted in a `TargetSupportsQir` method instead of having a hardcoded value. That way we can just update that method whenever a new QIR target becomes available.

src/AzureClient/EntryPoint/EntryPointGenerator.cs
ScottCarda-MS: updated to use `generateQir`

src/Core/Compiler/CompilerService.cs
```diff
@@ -260,25 +269,54 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source)
...
-                    logger.LogError("IQS005", "Failed to write compilation to binary stream.");
+                    var transformed = TrimSyntaxTree.Apply(qsCompilation, keepAllIntrinsics: false);
+                    transformed = InferTargetInstructions.ReplaceSelfAdjointSpecializations(transformed);
+                    transformed = InferTargetInstructions.LiftIntrinsicSpecializations(transformed);
+                    var allAttributesAdded = InferTargetInstructions.TryAddMissingTargetInstructionAttributes(transformed, out transformed);
+
+                    using var generator = new Generator(transformed);
+                    generator.Apply();
+
+                    // write generated QIR to disk
```
cesarzc, line 306:
```suggestion
                    // Write generated QIR to disk.
```

cesarzc: nit

src/Kernel/Magic/QirMagic.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            this.EntryPointGenerator = entryPointGenerator;
+            this.Logger = logger;
+        }
+
+        private ILogger? Logger { get; }
+
+        // TODO: EntryPointGenerator might should move out of the Azure Client
+        //       project.
+        public IEntryPointGenerator EntryPointGenerator { get; }
```
cesarzc, line 53: Is this TODO being tracked as an issue? It seems like a possibly large change.

src/Tests/AzureClientEntryPointTests.cs
```diff
@@ -81,6 +82,94 @@ public async Task FromSnippet()
...
+
+        [TestMethod]
+        public async Task ValidParameterTypes()
+        {
+            var entryPointGenerator = Init("Workspace", new string[] { SNIPPETS.ValidParameterTypes });
+            var entryPoint = entryPointGenerator.Generate("UseValidParameterTypes", null, forceQir: true);
+
+            Assert.IsNotNull(entryPoint);
+
+            var validArguments = new Dictionary<Argument, string>()
```
cesarzc, line 106: Can you please add test cases in which an entry-point operation receives multiple parameters (mixed-type parameters, same-type parameters)?

src/AzureClient/EntryPoint/IEntryPoint.cs
ScottCarda-MS: The QirStream is used here in QirMagic.cs: https://github.com/microsoft/iqsharp/blob/36f2ebffe60bf66261435c4dad99e96d47c833f2/src/Kernel/Magic/QirMagic.cs#L92

src/Tests/AzureClientEntryPointTests.cs
```diff
@@ -81,6 +82,94 @@ public async Task FromSnippet()
...
+                new MockQIRSubmitter(validArguments.Select(x => x.Key).ToList()),
+                new AzureSubmissionContext()
+                {
+                    InputParameters = validArguments.ToDictionary(x => x.Key.Name, x => x.Value)
+                });
+            Assert.IsNotNull(job);
+        }
+
+        [TestMethod]
+        public async Task InvalidParameterTypes()
```
cesarzc, line 128: Can you please add test cases for operations that have a mix of valid and invalid parameter types?

src/Tests/AzureClientTests.cs
```diff
@@ -180,7 +179,9 @@ public void TestAllTargets()
             // connect to mock workspace with all providers
             var targets = ExpectSuccess<IEnumerable<TargetStatusInfo>>(ConnectToWorkspaceAsync(azureClient, MockAzureWorkspace.NameWithMockProviders));
             // 2 targets per provider: mock and simulator.
-            Assert.AreEqual(2 * Enum.GetNames(typeof(AzureProvider)).Length, targets.Count());
+
+            // TODO: uncomment once there is a microsoft.mock target
+            //Assert.AreEqual(2 * Enum.GetNames(typeof(AzureProvider)).Length, targets.Count());
```
cesarzc, line 184: Will the addition of a microsoft.mock target be addressed as a separate PR? If so, this TODO should also be tracked somewhere so we don't forget.

src/AzureClient/Extensions.cs
ScottCarda-MS: It is being used here in QirMagic.cs: https://github.com/microsoft/iqsharp/blob/36f2ebffe60bf66261435c4dad99e96d47c833f2/src/Kernel/Magic/QirMagic.cs#L81

ScottCarda-MS: I'm not sure where in the Core would be a good place to put it.

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: I see the hardcoded value you are referring to here, but I don't understand your suggestion regarding a `TargetSupportsQir` method. What object would the method live on, and what would the method do?

src/AzureClient/AzureClient.cs
cesarzc: Thinking more carefully about it, it seems like `AvailableTargets` should not have any changes made to it at all. My guess is that this was added to workaround the `Microsoft` provider not being listed but I am not convinced it is still needed.

@cgranade could you please help us understand whether this change makes sense or if it is a workaround?

src/AzureClient/Mocks/MockQIRSubmitter.cs
ScottCarda-MS: done

src/AzureClient/Mocks/MockQIRSubmitter.cs
ScottCarda-MS: done

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: done

src/Tests/AzureClientTests.cs
ScottCarda-MS: When I raised the issue, it didn't seem to be a priority with anyone. I can make a git issue for it nevertheless. I currently have no plans to execute on this, and if I did I would need to have someone more experienced walk me through how to add `microsoft.mock`, as I wasn't able to figure it out on my own.

src/AzureClient/AzureClient.cs
cesarzc: In the way the logic is currently written, we decide whether we need an `IQirSubmitter` or an `IQuantumMachine` object based on the target ID. However, both the `SubmitterFactory.QirSubmitter` and the `AzureFactory.CreateMachine` factory methods already have logic inside them to handle this. If the target ID is not valid to create a submitter object or a machine object, they return null.

My suggestion is to leverage this logic. If a submitter object for the target can be created, then we can just use it. Otherwise, we try to create a machine object for the target and use it if it could be created.

Hopefully this clarifies my comment.

src/AzureClient/EntryPoint/EntryPoint.cs
cesarzc: It is tricky indeed!

src/AzureClient/Extensions.cs
cesarzc: In that case I think instead of making this public, the `Logger` object should be used directly instead of an extension method on the `Exception` class. Another option would be to create a `Common` project where these types of extension methods are implemented.

src/Kernel/Magic/QirMagic.cs
ScottCarda-MS: The EntryPointGenerator is supposed to be moved out of the Azure Client project? I'm not sure why that would be desirable. I will make a GitHub issue for it anyway.

ScottCarda-MS: Issue: #610

src/AzureClient/Extensions.cs
ScottCarda-MS: Okay I've removed the external reference to Log and just had it call the Logger object directly.

ScottCarda-MS: And the access has been set back to `internal` for Log.

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: Not sure what you mean. The test ValidParameterTypes tests a mix of all the valid parameter types, so I believe this is what you are suggesting, right?

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: The "UseEmbeddedUnitType" Snippet case should test that, I believe. It is part of the InvalidParameterTypes test. Was there something different you had in mind?

src/AzureClient/EntryPoint/EntryPoint.cs
ScottCarda-MS: I've added it as a public static string to CompilerService. However, now we have a public-facing string that is just "ENTRYPOINT". Is that the direction we want to go with this?

src/Core/Compiler/CompilerService.cs
cesarzc: The ideal place for this method to live would be here: [SubmitterFactory.cs](https://github.com/microsoft/qsharp-runtime/blob/main/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs). It is a cross-repo change so I suggest adding a ToDo comment here and create the corresponding issue.

src/Kernel/Magic/QirMagic.cs
cesarzc: Thanks Scott!

src/Tests/AzureClientEntryPointTests.cs
cesarzc: I see, I did not read the code correctly. Then, in addition to this mixed-types case, can you please add one individual test case per argument type? That way if there's an issue with a specific type it would be easy to know the problematic one.

src/Tests/AzureClientEntryPointTests.cs
cesarzc: It would be great to add similar cases but with the other two unsupported types (array and range). For these new cases it would also be great to position the unsupported type at the beginning and end of the list just to add a bit of coverage on the position of arguments too.

src/AzureClient/AzureClient.cs
cgranade: I think that the workaround is that `AzureExecutionTarget` assumes providers need their own packages, but that could possibly be mitigated by updating https://github.com/microsoft/iqsharp/blob/f5274e9a4b42160a1725633792521300889e6f1c/src/AzureClient/AzureExecutionTarget.cs#L24 instead?

src/AzureClient/AzureClient.cs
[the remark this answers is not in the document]
cgranade: Agreed; this should be resolved before the final merge to main, but let's defer that for now.

src/Kernel/Magic/QirMagic.cs
cgranade: > I'm not sure why that would be desirable.

For the reasoning there, the Azure client should be specific to what's needed to submit to Azure, but we use entry point generation in local QIR workflows as well.

src/Python/qsharp-core/qsharp/loader.py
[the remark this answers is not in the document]
cgranade: This whole block of code can likely be removed now that `as_qir` just returns a `bytes` value.

src/Python/qsharp-core/qsharp/loader.py
```diff
@@ -114,6 +127,23 @@ def trace(self, **kwargs) -> Any:
         """
         return qsharp.client.trace(self, **kwargs)
 
+    def as_qir(self) -> bytes:
+        """
+        Returns the QIR bitcode representation of the callable,
+        assuming the callable is an entry point.
+        """
+        if getattr(pqp, "failed_with", None) is not None:
+            raise pqp.failed_with
```
cgranade, line 136:
```suggestion
```

src/Tests/AzureClientEntryPointTests.cs
```diff
@@ -81,6 +82,94 @@ public async Task FromSnippet()
...
+        public async Task ValidParameterTypes()
+        {
+            var entryPointGenerator = Init("Workspace", new string[] { SNIPPETS.ValidParameterTypes });
+            var entryPoint = entryPointGenerator.Generate("UseValidParameterTypes", null, generateQir: true);
+
+            Assert.IsNotNull(entryPoint);
+
+            var validArguments = new Dictionary<Argument, string>()
+            {
+                { new Argument("myBool", new ArgumentValue.Bool(true)), "\"true\"" },
```
cgranade, line 108: [code style nit] Dictionary collection syntax is preferred when initializing a dictionary in particular.
```suggestion
                [new Argument("myBool", new ArgumentValue.Bool(true))] = "\"true\"",
```

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: done

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: done

src/Tests/AzureClientEntryPointTests.cs
ScottCarda-MS: done

src/Python/qsharp-core/qsharp/loader.py
ScottCarda-MS: done

src/Python/qsharp-core/qsharp/loader.py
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Core/AssemblyInfo.cs
```diff
@@ -55,6 +73,11 @@ public AssemblyInfo(Assembly assembly, string location, QsNamespace[] syntaxTree
         /// </summary>
         public QsNamespace[] SyntaxTree { get; }
 
+        /// <summary>
+        /// The stream from which QIR bitcode for the entry point can be read.
+        /// </summary>
+        public Stream QirBitcode { get; }
```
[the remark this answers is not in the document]
ScottCarda-MS: This file doesn't use the `#nullable enable` directive, and it doesn't make sense to add it given that most everything in this file is nullable.

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: The logic for determining if we should generate QIR should live in the SubmitterFactory as a static function, `TargetSupportsQir`. We should then be calling that static function here for PrepareQirGeneration instead of this logic.

src/AzureClient/EntryPoint/EntryPoint.cs
ScottCarda-MS: As per our discussion, I've made the CompilerService's EntrypointNamespaceName member private, and changed the other two locations to use the string literal with a comment that they must be the same as the CompilerService's namespace name.

src/AzureClient/AzureClient.cs
cesarzc: Ok, after going through this more in depth, I am more convinced that this line needs to remain as it was.

My analysis is the following. The problem with the `microsoft.simulator` target is that the `AzureExecutionTarget.GetProvider()` method would return null because of how it is implemented. However, the `microsoft.simulator` target is not really filtered from `AvaliableProviders` so it seems like the where is effectively a no-op.

Am I missing something?

ScottCarda-MS: I've had to undo @cgranade's suggestion as that was causing issues with the system finding and loading package Microsoft.Quantum.Providers.Core. As for AvailableTargets, the 'where' clause is actually filtering out quite a lot of targets that I'm seeing. There are a lot of targets whose provider is Microsoft that don't start with 'microsoft.simulator'. Evidently we only want targets with the Microsoft provider if they start with 'microsoft.simulator' plus any non-Microsoft targets.

ScottCarda-MS: Part of this may be caused by something strange I've noticed: the targets we want originate from the MicrosoftSimulator provider, but when those targets are fed into GetProvider, they return the Microsoft provider instead.

src/AzureClient/AzureClient.cs
cesarzc:
```suggestion
        /// <summary>Returns whether a target ID is meant for quantum execution since not all targets exposed by providers are meant for that. More specifically, the Microsoft provider exposes targets that are not meant for quantum execution and the only ones meant for that start with "microsoft.simulator"</summary>
        private static bool IsQuantumExecutionTarget(string targetId) =>
            AzureExecutionTarget.GetProvider(targetId) == AzureProvider.Microsoft ?
                targetId.StartsWith("microsoft.simulator") :
                true;
```

src/AzureClient/AzureClient.cs
```diff
@@ -39,7 +43,12 @@ public class AzureClient : IAzureClient
         private AzureExecutionTarget? ActiveTarget { get; set; }
         private string MostRecentJobId { get; set; } = string.Empty;
         private IEnumerable<ProviderStatusInfo>? AvailableProviders { get; set; }
-        private IEnumerable<TargetStatusInfo>? AvailableTargets => AvailableProviders?.SelectMany(provider => provider.Targets);
+        private IEnumerable<TargetStatusInfo>? AvailableTargets => 
+            AvailableProviders
+            ?.SelectMany(provider => provider.Targets)
+            ?.Where(t => t.TargetId != null
+                && (t.TargetId.StartsWith(MicrosoftSimulator) // Of the Microsoft targets, we only want those that start with 'microsoft.simulator'.
+                || AzureExecutionTarget.GetProvider(t.TargetId) != AzureProvider.Microsoft)); // We want all non-Microsoft targets.
```
cesarzc, line 51:
```suggestion
            ?.Where(t => t.TargetId != null ? IsQuantumExecutionTarget(t.TargetId) : false);
```

cesarzc: This suggestion uses the method mentioned in [this comment](https://github.com/microsoft/iqsharp/pull/601/files#r831420468).

cesarzc (approved):
I just left a couple of suggestions on how to make the filtering of Microsoft targets easier to understand. Other than that, looks great to me!

src/AzureClient/AzureClient.cs
ScottCarda-MS: I've added something similar to this suggestion, but I think it is simpler without the ternary operators and to just use regular Boolean operators.

ScottCarda-MS: The `MicrosoftSimulator` string will become even more important in the upcoming PR for the output formatting.

src/AzureClient/AzureClient.cs
ScottCarda-MS: I've added your suggestion but slightly modified. I think using regular Boolean operators is a bit clearer than using the ternary operator. Also the `MicrosoftSimulator` string is still needed elsewhere so that got left in.

cesarzc: This modification looks good :)

PR #602, opened 2022-03-04
Add procps to Dockerfile

guenp:
This resolves the following warning message when importing `qsharp` via Python:
`/bin/bash: free: command not found`

Dependency was introduced in https://github.com/microsoft/iqsharp/pull/528


cgranade (approved):
Thanks for adding this, @guenp. Separately, I'll go on and double-check the performance metrics class to make sure that stderr is properly redirected to avoid accidentally propagating warnings through to the end user.

cgranade:
It looks like the build is blocked on alpha packages having aged out of the feed; I'll go on and update the packages on a separate PR to restore the build, then.

PR #603, opened 2022-03-04
Quantinuum rebrand → main

cgranade:
This PR merges existing work on the Quantinuum rebrand to main.


ricardo-espinoza (changes requested):
Should we take this PR as is with the references to beta packages and then update all the repos to the released build, or should we absorb those changes here first?

cgranade:
> Should we take this PR as is with the references to beta packages and then update all the repos to the released build, or should we absorb those changes here first?

@ricardo-espinoza: I think this should now be resolved by merging main into this PR. If you could, please take another look. Thanks!

cgranade:
Closing as this PR is now redundant.

PR #605, opened 2022-03-04
Improve async discovery of magic commands.

cgranade:
This PR improves async discovery of magic commands, ensuring that cells run in the first ~1s after process start are accurately discovered, and caching redundant lookups to improve future cell execution. Marked in draft as additional testing is needed.


src/Kernel/IQSharpEngine.cs
```diff
@@ -345,12 +365,16 @@ public override async Task<ExecutionResult> ExecuteMundane(string input, IChanne
...
                 {
-                    await this.Initialized;
-                    // Once the engine is initialized, we know that Workspace
+                    // Since this method is only called once this.Initialized
+                    // has completed, we know that Workspace
                     // and Snippets are both not-null.
+                    Debug.Assert(
+                        this.Initialized.IsCompleted,
+                        "Engine was not initialized before call to ExecuteMundane. " +
+                        "This is an internal error; if you observe this message, please file a bug report at https://github.com/microsoft/iqsharp/issues/new."
```
anpaz, line 374: I really like this idea about asking the user to file an issue if they reach an Assert!

src/Kernel/IQSharpEngine.cs
```diff
@@ -178,6 +178,17 @@ private async Task StartAsync()
             this.Workspace = await serviceTasks.Workspace;
             var references = await serviceTasks.References;
 
+            // Start registering magic symbols; we do this in the engine rather
```
anpaz, line 181: Would it make sense to  move this a little earlier?
Not sure this code reqires all other services to be ready.

anpaz (approved):
lgtm.

src/Kernel/IQSharpEngine.cs
cgranade: Good point, I think I can move this above the other awaits, too. Let me go on and do that.

PR #611, opened 2022-03-14
[Perf] Add extra nuget packages to iqsharp-base container

xinyi-joffre:
The %azure.target command currently has poor performance:

Looking at the cell duration telemetry, the 95th percentile for %azure.target commands is 50 seconds, and the 99.5 percentile is 2 min 43 seconds.

After the first time it is executed with poor performance, subsequent cells have very good performance (<1s, even after a kernel restart). It looks like this is due to the packages that need to be downloaded or unpacked from Nuget. Executing the %azure.target command for our three providers adds **55 nuget packages** to the container.

By installing them in the base image first, we can improve the performance of this command, so that it is sub-second even for the first time it is run (assuming the user has not explicitly specified a nuget version not in our nuget cache).

### DIFF
--- a/build/pack.ps1
+++ b/build/pack.ps1
@@ -84,7 +84,7 @@ function Pack-Image() {
         <# Next, we specify any additional NuGet sources to be used. #> `
         --build-arg EXTRA_NUGET_SOURCES="$extraNugetSources" `
         <# Next, we specify any additional NuGet packages that should be imported. #> `
-        --build-arg EXTRA_NUGET_PACKAGES="" `
+        --build-arg EXTRA_NUGET_PACKAGES=$("microsoft.quantum.providers.ionq", "microsoft.quantum.providers.honeywell", "microsoft.quantum.providers.qci") `
         <# Next, we tell Docker what version of IQ# to install. #> `
         --build-arg IQSHARP_VERSION=$Env:NUGET_VERSION `
         <# Finally, we tag the image with the current build number. #> `


xinyi-joffre:
This will need to be updated once the nuget package for honeywell is renamed to quantinuum

build/pack.ps1
```diff
@@ -84,7 +84,7 @@ function Pack-Image() {
         <# Next, we specify any additional NuGet sources to be used. #> `
         --build-arg EXTRA_NUGET_SOURCES="$extraNugetSources" `
         <# Next, we specify any additional NuGet packages that should be imported. #> `
-        --build-arg EXTRA_NUGET_PACKAGES="" `
+        --build-arg EXTRA_NUGET_PACKAGES=$("microsoft.quantum.providers.ionq", "microsoft.quantum.providers.honeywell", "microsoft.quantum.providers.qci") `
```
cgranade, line 87: This package names may be case sensitive? Would also need to check if this gets overwritten in the e2e QDK build.

cgranade: Looking at the internal definition at https://github.com/microsoft/qdk/blob/df63ff5521801cbe16a5fc3b84422f6d9c3308e9/build.vnext/steps-docker.yml#L18, I believe this change may not be effective for the e2e build. I think you've also used PowerShell notation here, but this argument will be processed by Docker as a space-separated list of strings instead?

xinyi-joffre: Thanks! I'll make a PR there as well!

For the argument, I did check that it builds and works locally using the same RUN logic and passing in those parameters exactly--but I can swap to space separated list in strings for consistency, if that's how we use it elsewhere!

xinyi-joffre: Is there still value in adding it here? Or should I abandon this PR and just go with the internal definition?

Issue #594, opened 2022-02-22
Add type signatures to the built-in help command


guenp:
**Is your feature request related to a problem? Please describe.**
Currently, the built-in help command invoked with `?` only returns the Operation or Function name and docstring but not the type signature.

**Describe the solution you'd like**
I'd like to be able to see the type signature in the notebook instead of having to browse the docs to find them.

**Describe alternatives you've considered**
-

**Additional context**
Screenshot:
![MicrosoftTeams-image](https://user-images.githubusercontent.com/4041805/155205549-eeb92361-1818-4b2c-b155-12272a70076f.png)

cgranade:
Thanks for filing this! I think we should do this, but that would take adding an API to the qsharp-compiler repo (or exposing it if it already exists) to reconstruct the declaration syntax from a `CallableDeclarationHeader`.

To drill down a bit, the way `?` works in jupyter-core and hence IQ# is to return a code symbol as a displayable object, and then use display encoders to render that into HTML, plain text, and so forth. IQ# already has a display encoder for Q# symbols, formatting them from the metadata contained in `IQSharpSymbol`. That class includes metadata automatically generated from documentation comments (https://github.com/microsoft/iqsharp/pull/334), but at the moment there's no way to go back to Q# syntax data without using internal-only syntax generation methods.

@bettinaheim, would it be reasonable to expose some of the methods used in documentation generation for reproducing declaration syntax so that we can implement this suggestion?

Short of that, I think we should be able to pull at least some Markdown content from existing attributes and APIs:

<img width="388" alt="image" src="https://user-images.githubusercontent.com/31516/155220135-596a7c66-0aff-480f-b554-81f00b2cc2fe.png">

PR #614, opened 2022-03-15
Fix for #594.

cgranade:
This PR suggests a fix for #594, and is marked in draft to avoid duplication of code in the qsharp-compiler repo. If we want to adopt these changes, I would suggest that a corresponding PR should be made against qsharp-compiler to make the `SyntaxExtensions` class publicly available.

## Example screenshot

image

### DIFF
--- a/src/Jupyter/ConfigurationSource/IConfigurationSource.cs
+++ b/src/Jupyter/ConfigurationSource/IConfigurationSource.cs
@@ -231,5 +231,29 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
         /// </summary>
         public string WorkspaceName =>
             GetOptionOrDefault("azure.quantum.workspace.name", string.Empty);
+
+        // We also want to support some internal-only options that are disabled
+        // in release builds, making it easier to diagnose some internals
+        // during local development. When in release mode, we'll disable setting
+        // these options and always use defaults.
+
+        #if DEBUG
+            public TEnum GetInternalOptionOrDefault<TEnum>(string optionName, TEnum defaultValue) where TEnum : struct =>
+                GetOptionOrDefault("internal." + optionName, defaultValue, Enum.Parse<TEnum>);
+            public string GetInternalOptionOrDefault(string optionName, string defaultValue) =>
+                GetOptionOrDefault("internal." + optionName, defaultValue, e => e);
+            public bool GetInternalOptionOrDefault(string optionName, bool defaultValue) =>
+                GetOptionOrDefault("internal." + optionName, defaultValue, bool.Parse);
+        #else
+            public string GetInternalOptionOrDefault(string optionName, string defaultValue) =>
+                defaultValue;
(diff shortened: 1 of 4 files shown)


cgranade:
@xinyi-joffre: Is this something you would find useful in the portal-hosted notebook experience?

xinyi-joffre:
Yes, this looks like a cool feature!

cgranade:
@xinyi-joffre: Awesome!

@samarsha: Any thoughts on where code that's currently duplicated should live longer-term? Thanks!

cgranade:
@samarsha: Apologies for the hassle, but wanted to ping again for your thoughts on this one. Thanks!

samarsha:
- `ToSyntax` could go in the Q# unparser in the compiler (Transformations/QsharpCodeOutput.cs).
- `InputDeclarations` could maybe go in Core/SyntaxTreeExtensions.fs or some other general utility file. Probably would generalize it to return `string?` or `string option` for the invalid names.
- I guess it makes sense for `TryResolveXref`, `ToLink`, and `ToHtml` to stay in DocumentationGenerator. But you'd need to generalize them to output either Markdown or HTML?

cgranade:
Much obliged, really appreciate your thoughts!

> * `ToSyntax` could go in the Q# unparser in the compiler (Transformations/QsharpCodeOutput.cs).

Sounds good, we may also want to refactor DocumentationGenerator to rely on some of that as well then. At least when DocumentationGenerator was first written, the unparser was missing a few cases and had a fair bit marked as private or internal.

> * `InputDeclarations` could maybe go in Core/SyntaxTreeExtensions.fs or some other general utility file. Probably would generalize it to return `string?` or `string option` for the invalid names.

Makes sense! Would you prefer it in that F# project, then?

> * I guess it makes sense for `TryResolveXref`, `ToLink`, and `ToHtml` to stay in DocumentationGenerator. But you'd need to generalize them to output either Markdown or HTML?

Sounds good, I guess then we could pull in the [Microsoft.Quantum.DocumentationGenerator package](https://www.nuget.org/packages/Microsoft.Quantum.DocumentationGenerator/0.24.203411-beta f) from here to get that extension, would just need to mark it public in that project?

samarsha:
> Sounds good, we may also want to refactor DocumentationGenerator to rely on some of that as well then. At least when DocumentationGenerator was first written, the unparser was missing a few cases and had a fair bit marked as private or internal.

I think the unparser's API could use a lot of improvement, yeah - if you have specific examples of things the documentation generator wants, that would be helpful.

> Makes sense! Would you prefer it in that F# project, then?

Core is probably the best place for it, yeah.

> Sounds good, I guess then we could pull in the [Microsoft.Quantum.DocumentationGenerator package](https://www.nuget.org/packages/Microsoft.Quantum.DocumentationGenerator/0.24.203411-beta f) from here to get that extension, would just need to mark it public in that project?

Yes, there'd be a dependency here on the DocumentationGenerator package.

PR #615, opened 2022-03-16
Improve first cell performance and add performance diagnostics options

cgranade:
This PR improves first cell performance by using asynchronous tasks to load dependencies needed by the C# and Q# compilers in the background at kernel start, reducing the amount that needs to be done when user input is provided. This in particular improves performance when either a magic command is the first cell, or when the user doesn't execute any particular cell immediately.

This PR also introduces new preferences that allow for viewing performance data within a notebook, helping diagnose performance issues that may arise in the future.

#breaking-change: This is a breaking change, since the optimizations in this PR involve making some existing synchronous methods async.

@tcNickolas: for adapting the katas, only a small change is needed (see https://github.com/microsoft/QuantumKatas/pull/765).


xinyi-joffre (approved):

cgranade:
It looks like builds have completed but are not showing up correctly due to the intermittent GitHub outage today. Overriding to merge after manually checking build results.

PR #619, opened 2022-03-18
Merge in from main

ScottCarda-MS:
Merge in changes from main


cgranade (approved):

cesarzc (approved):

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.100",
```
cgranade: This change should be included in the roll-forward key below. In particular, incrementing to `6.0.201` will make it harder for users to use this repo locally with `6.0.100`. I would suggest keeping the rest of this commit, but reverting this line.

ScottCarda-MS: okay, I've put that back at 6.0.100

Issue #606, opened 2022-03-08
Lambda functions and operations cannot be used from IQ#.


cgranade:
**Describe the bug**
New lambda functions and operations (e.g.: `(x) -> x`) cannot currently be compiled using IQ#.

**To Reproduce**

1. From a Python + Q# notebook, run `import qsharp`, and then run the cell:
   ```python
   %%qsharp
   function Foo() : Unit {
       let f = x -> x;
       let y = f(12);
   }
   ```
2. Observe error:
   ```text
   : error IQS002: Unexpected error compiling assembly: Can't generate code for un-lifted lambda..
   ---------------------------------------------------------------------------
   TypeError                                 Traceback (most recent call last)
   ~\AppData\Local\Temp/ipykernel_35408/4074425879.py in <module>
   ----> 1 get_ipython().run_cell_magic('qsharp', '', 'function Foo() : Unit {\n    let f = x -> x;\n    
   let y = f(12);\n}\n')

   ~\Anaconda3\lib\site-packages\IPython\core\interactiveshell.py in run_cell_magic(self, magic_name, line, cell)
      2417             with self.builtin_trap:
      2418                 args = (magic_arg_s, cell)
   -> 2419                 result = fn(*args, **kwargs)
      2420             return result
      2421 

   ~\Anaconda3\lib\site-packages\qsharp\ipython_magic.py in qsharp(magic_args, cell, local_ns)
        26             local_ns[callables._name] = callables
        27         else:
   ---> 28             for qs_callable in callables:
        29                 local_ns[qs_callable._name] = qs_callable
        30 

   TypeError: 'NoneType' object is not iterable
   ```

**Expected behavior**
The above snippet should either compile correctly or fail with a clear and actionable error message.

**Screenshots**
<img width="845" alt="image" src="https://user-images.githubusercontent.com/31516/157292894-8fd1821d-9020-49d8-819f-4c7867994894.png">

**System information**

```text
In []: qsharp.component_versions()
Out[]: {'iqsharp': LooseVersion ('0.23.195983'),
 'Jupyter Core': LooseVersion ('1.5.0.0'),
 '.NET Runtime': LooseVersion ('.NETCoreApp,Version=v3.1'),
 'qsharp': LooseVersion ('0.23.195983')}
```

**Additional context**

[...]

PR #620, opened 2022-03-18
Add regression test for #606.

cgranade:
This PR checks for regression against #606, and is expected to fail until IQ# is updated to use the build from https://github.com/microsoft/qsharp-compiler/pull/1381.


anjbur (approved):

cgranade:
CI should pass now that https://github.com/microsoft/qsharp-compiler/pull/1381 is complete. Re-running now.

PR #621, opened 2022-03-21
Update local build to install .NET 6

anjbur:


cgranade (approved):
LGTM, except that `6.0.100` excludes some necessary security patches. I agree with only requiring 6.0.100, but for our own builds, we may want to use 6.0.201 to include those patches.

ScottCarda-MS (approved):
Thanks @anjbur!

PR #623, opened 2022-03-22
QIR Submission Output

ScottCarda-MS:
Avoids printing a histogram as output when running a Microsoft Simulator target.

### DIFF
--- a/src/AzureClient/AzureClient.cs
+++ b/src/AzureClient/AzureClient.cs
@@ -523,18 +523,6 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             return AvailableTargets.First(target => target.TargetId == ActiveTarget.TargetId).ToExecutionResult();
         }
 
-        private static IEnumerable<string> ReadAllLines(Stream stream)
-        {
-            using (var reader = new StreamReader(stream))
-            {
-                string line = "";
-                while ((line = reader.ReadLine()) != null)
-                {
-                    yield return line;
-                }
-            }
-        }
-
         /// <inheritdoc/>
         public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string jobId, CancellationToken? cancellationToken = default)
         {
@@ -587,23 +575,8 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 using var responseStream = (await request.GetResponseAsync()).GetResponseStream();
                 if (this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
                 {
-                    var outputLines = ReadAllLines(responseStream).ToList();
-                    outputLines = outputLines.Select(line => line.Trim()).ToList();
-                    
-                    // ToDo: This is a hack to separate Messages from the result.
-                    var resultStartLine = outputLines.Count() - 1;
-                    if (outputLines[resultStartLine].EndsWith('"'))
-                    {
-                        while (!outputLines[resultStartLine].StartsWith('"'))
-                        {
-                            resultStartLine -= 1;
-                        }
-                    }
-                    
-                    var messages = String.Join('\n', outputLines.Take(resultStartLine));
+                    var (messages, result) = ParseSimulatorOutput(responseStream);
                     channel?.Stdout(messages);
-                    var result = String.Join(' ', outputLines.Skip(resultStartLine));
-
                     return result.ToExecutionResult();
                 }
                 else
@@ -617,6 +590,36 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                 Logger?.LogError(e, $"Failed to download the job output for the specified Azure Quantum job: {e.Message}");
                 return AzureClientError.JobOutputDownloadFailed.ToExecutionResult();
             }
+        }
+
+        private static (string Messages, string Result) ParseSimulatorOutput(Stream stream)
+        {
+            var outputLines = new List<string>();
+            using (var reader = new StreamReader(stream))
+            {
+                var line = String.Empty;
+                while ((line = reader.ReadLine()) != null)
+                {
+                     outputLines.Add(line.Trim());
+                }
+            }
+
+            // N.B. The current simulator output format is just text and it does not distinguish
+            // between the result of the operation and other kinds of output.
+            // Attempt to parse the output to distinguish the result from the rest of the output
+            // until the simulator output format makes it easy to do so.
+            var resultStartLine = outputLines.Count() - 1;
+            if (outputLines[resultStartLine].EndsWith('"'))
+            {
+                while (!outputLines[resultStartLine].StartsWith('"'))
+                {
+                    resultStartLine -= 1;
+                }
+            }
+
+            var messages = String.Join('\n', outputLines.Take(resultStartLine));
+            var result = String.Join(' ', outputLines.Skip(resultStartLine));
+            return (messages, result);
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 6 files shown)


src/AzureClient/AzureClient.cs
```diff
@@ -572,10 +585,31 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string j
                 //                 cancellation token support.
                 var request = WebRequest.Create(job.OutputDataUri);
                 using var responseStream = (await request.GetResponseAsync()).GetResponseStream();
-                return responseStream.ToHistogram(
-                    Logger, 
-                    isSimulatorOutput: this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
-                .ToExecutionResult();
+                if (this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
+                {
+                    var outputLines = ReadAllLines(responseStream).ToList();
```
cesarzc, line 590: If I remember correctly, the idea was to show the same output as the simulator running locally. This includes all the output produced through the `Message` API, so I believe that trying to separate message from result is not needed.

ScottCarda-MS: There is a subtle difference between how the Messages are displayed and how output is displayed, even when running the simulator locally. Output will be prefaced with a red 'Out[n]:' next to it visually differentiate it from the Messages. I am separating out the Messages from the output here to have the same effect.

src/AzureClient/AzureClient.cs
```diff
@@ -572,10 +585,31 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string j
...
+                        {
+                            resultStartLine -= 1;
+                        }
+                    }
+                    
+                    var messages = String.Join('\n', outputLines.Take(resultStartLine));
+                    channel?.Stdout(messages);
+                    var result = String.Join(' ', outputLines.Skip(resultStartLine));
+
+                    return result.ToExecutionResult();
```
cesarzc, line 607: I think it would be good to encapsulate the code that differentiates messages from result into a single method that returns a tuple:
```csharp
private static (string Messages, string Result) ParseSimulatorOutput(Stream stream)
```

This way the only thing that has to go inside the `if` statement is:
```csharp
var parsedOutput = ParseSimulatorOutput(Stream stream);
channel?.Stdout(parsedOutput.messages);
parsedOutput.Result.ToExecutionResult();
```

src/AzureClient/AzureClient.cs
```diff
@@ -572,10 +585,31 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string j
...
                 var request = WebRequest.Create(job.OutputDataUri);
                 using var responseStream = (await request.GetResponseAsync()).GetResponseStream();
-                return responseStream.ToHistogram(
-                    Logger, 
-                    isSimulatorOutput: this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
-                .ToExecutionResult();
+                if (this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
+                {
+                    var outputLines = ReadAllLines(responseStream).ToList();
+                    outputLines = outputLines.Select(line => line.Trim()).ToList();
```
cesarzc, line 591: Given that you return an `IEnumerable`, I believe you could just concatenate linq calls without having to use `ToList()`:
```csharp
var outputLines = ReadAllLines(responseStream)
    . Select(line => line.Trim())
    .ToList();
```

src/AzureClient/AzureClient.cs
```diff
@@ -522,6 +523,18 @@ public async Task<ExecutionResult> SetActiveTargetAsync(IChannel channel, string
             return AvailableTargets.First(target => target.TargetId == ActiveTarget.TargetId).ToExecutionResult();
         }
 
+        private static IEnumerable<string> ReadAllLines(Stream stream)
+        {
+            using (var reader = new StreamReader(stream))
+            {
+                string line = "";
```
cesarzc, line 530: nit: I usually find `string.Empty` clearer and less error-prone.

src/AzureClient/AzureClient.cs
```diff
@@ -522,6 +523,18 @@ public async Task<ExecutionResult> SetActiveTargetAsync(IChannel channel, string
             return AvailableTargets.First(target => target.TargetId == ActiveTarget.TargetId).ToExecutionResult();
         }
 
+        private static IEnumerable<string> ReadAllLines(Stream stream)
```
cesarzc, line 526: If you are doing encapsulation into `ParseSimulatorOutput` as I suggested in this [comment](https://github.com/microsoft/iqsharp/pull/623/files#r837900742), I would also implement this method inside of `ParseSimulatorOutput` since it is implemented specifically for that and it is short enough.

src/AzureClient/AzureClient.cs
```diff
@@ -572,10 +585,31 @@ public async Task<ExecutionResult> GetJobResultAsync(IChannel? channel, string j
...
-                return responseStream.ToHistogram(
-                    Logger, 
-                    isSimulatorOutput: this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
-                .ToExecutionResult();
+                if (this.ActiveTarget?.TargetId?.StartsWith(MicrosoftSimulator) ?? false)
+                {
+                    var outputLines = ReadAllLines(responseStream).ToList();
+                    outputLines = outputLines.Select(line => line.Trim()).ToList();
+                    
+                    // ToDo: This is a hack to separate Messages from the result.
```
cesarzc, line 593:
```suggestion
                    // N.B. The current simulator output format is just text and it does not distinguish between the result of the operation and other kinds of output.
                    //      Attempt to parse the output to distinguish the result from the rest of the output until the simulator output format makes it easy to do so.
```

cesarzc: Expanded the comment to provide more context on why the custom parsing is needed.

src/AzureClient/AzureClient.cs
ScottCarda-MS: done

src/AzureClient/AzureClient.cs
ScottCarda-MS: I merged the ReadAllLines content to inside of the function and just did the line trimming there.

src/AzureClient/AzureClient.cs
ScottCarda-MS: done

src/AzureClient/AzureClient.cs
ScottCarda-MS: done

src/AzureClient/AzureClient.cs
ScottCarda-MS: I merged the content of that function into `ParseSimulatorOutput` rather than having a local function.

src/Core/References/References.cs
```diff
@@ -184,10 +184,16 @@ public async Task AddPackage(string name, Action<string>? statusCallback = null)
 
         private void Reset()
         {
+            var oldMetadata = _metadata;
```
cesarzc, line 187: Is this a fix to some of the build issues?

cesarzc (approved):
LGTM, thanks for addressing the feedback @ScottCarda-MS!

src/Core/References/References.cs
ScottCarda-MS: Yes this is a fix that @cgranade added to fix the tests running out of memory during builds.

PR #626, opened 2022-03-23
Add "size" (used in array literals) to the list of highlighted keywords

tcNickolas:
Also removed "newtype" since it's mentioned twice (line 22 is the first occurrence)

### DIFF
--- a/src/Kernel/client/syntax.ts
+++ b/src/Kernel/client/syntax.ts
@@ -34,7 +34,7 @@ export function defineQSharpMode() {
         },
         {
             token: "keyword",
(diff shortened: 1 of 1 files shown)


anjbur (approved):

tcNickolas:
In the parallel PR https://github.com/microsoft/qsharp-compiler/pull/1391 we're discussing that size should only be highlighted in certain contexts. I'll try and see if I can make that work, but let's not merge this as-is

raghavthind2005 (approved):

tcNickolas:
I didn't get to this in the past month and likely won't get to in the next month, so I've opened https://github.com/microsoft/iqsharp/issues/662 and will close this PR for now.

PR #627, opened 2022-03-23
Gate compiler caching on optional config flag.

cgranade:
This PR gates some of the improvements made in #615 on an optional configuration option to reduce memory usage by default while still allowing faster compilation times on systems with more memory.

### DIFF
--- a/src/Core/Compiler/CompilerService.cs
+++ b/src/Core/Compiler/CompilerService.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Quantum.IQSharp
 
 
             /// <summary>
-            ///     If <c>true</c>, caches compiler dependencies (e.g.: Roslyn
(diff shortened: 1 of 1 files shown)


anpaz (approved):

src/Core/Compiler/CompilerService.cs
```diff
@@ -53,6 +53,16 @@ public class Settings
             ///     are opened. Aliases can be provided by using <c>=</c>.
             /// </summary>
             public string? AutoOpenNamespaces { get; set; }
+
+
+            /// <summary>
+            ///     If <c>true</c>, caches compiler dependencies (e.g.: Roslyn
```
anpaz, line 59:
```suggestion
            ///     If <c>true</c>, loads and caches compiler dependencies (e.g.: Roslyn
```

PR #637, opened 2022-03-30
Reduce memory usage in overlapping load of references.

cgranade:
This PR reduces memory usage when reference loads overlap (e.g.: in unit tests, or when rapidly adding several packages).

### DIFF
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -5,6 +5,7 @@
     <PlatformTarget>x64</PlatformTarget>
     <RootNamespace>Microsoft.Quantum.IQSharp</RootNamespace>
     <AssemblyName>Microsoft.Quantum.IQSharp.Core</AssemblyName>
+    <LangVersion>10.0</LangVersion>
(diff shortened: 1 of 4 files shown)


src/Core/Core.csproj
```diff
@@ -5,6 +5,7 @@
     <PlatformTarget>x64</PlatformTarget>
     <RootNamespace>Microsoft.Quantum.IQSharp</RootNamespace>
     <AssemblyName>Microsoft.Quantum.IQSharp.Core</AssemblyName>
+    <LangVersion>10.0</LangVersion>
```
ricardo-espinoza, line 8: Question: Why is it required to set the language version to C# 10?

cgranade: The default for `netstandard2.1` is 7.0, IIRC, which doesn't include the `init;` modifier on property access. Added that to make sure that the properties guarded by the new cache aren't accidentally modified post-initialization.

anpaz (approved):

PR #639, opened 2022-03-30
Updating projects and other components to build 0.24.201332

ricardo-espinoza:


src/Tests/NugetPackagesTests.cs
```diff
@@ -28,7 +28,7 @@ public class NugetPackagesTests
         /// Note: Version '0.23.195531-beta' is temporary since it only exists in the Alpha feed.
```
cgranade, line 28: Can this comment be resolved now?

cgranade (approved):

src/Tests/NugetPackagesTests.cs
ricardo-espinoza: oh yes, let me remove it!

PR #642, opened 2022-04-04
Switch to use dotnet restore to cache packages in Docker image

anpaz:
We've noticed that in the qsharp-base image, loading Nuget packages from Python is failing to locally cache some of the dependencies, which is causing a slow startup.
With this change, we create a temp project and add the nuget dependencies via `dotnet`, which we have confirmed does load the dependencies.

### DIFF
--- a/images/iqsharp-base/Dockerfile
+++ b/images/iqsharp-base/Dockerfile
@@ -140,9 +140,10 @@ RUN dotnet iqsharp install --user --path-to-tool="$(which dotnet-iqsharp)"
 # Ensure that the necessary NuGet packages are cached.
 ARG EXTRA_NUGET_PACKAGES=
 RUN echo "Adding standard packages..." && \
-    python -c "import qsharp;" && \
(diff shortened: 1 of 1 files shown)


cgranade (approved):

images/iqsharp-base/Dockerfile
```diff
@@ -140,9 +140,10 @@ RUN dotnet iqsharp install --user --path-to-tool="$(which dotnet-iqsharp)"
 # Ensure that the necessary NuGet packages are cached.
 ARG EXTRA_NUGET_PACKAGES=
 RUN echo "Adding standard packages..." && \
-    python -c "import qsharp;" && \
+    dotnet new console -n "foo" -lang Q# && \
```
ricardo-espinoza, line 143: Do we want to delete the temporary project after adding the packages or are we ok with it remaining as part of the image?

anpaz: it is removed at the end.

PR #663, opened 2022-04-28
Cherry-pick #661 onto April 2022 release`

cgranade:
This PR applies #661 to the April 2022 release.

### DIFF
--- a/.github/workflows/devskim.yml
+++ b/.github/workflows/devskim.yml
@@ -1,0 +1,31 @@
+name: DevSkim
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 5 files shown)


.github/workflows/devskim.yml
```diff
@@ -0,0 +1,31 @@
+name: DevSkim
```
anpaz, line 1: do we really want to pick devskim for the release branch?

anpaz (approved):
lgtm. just wondering if we want to pick devskim for release branch.

.github/workflows/devskim.yml
cgranade: Ah, good catch, I accidentally cherry-picked that as well. I think it should be OK, though, since it only affects GH workflows and not Azure Pipelines; happy to open a new PR with that fixed, though, if you prefer.

PR #668, opened 2022-05-18
Fix image update GitHub Action

anjbur:
- Fixes typo in the path that was blocking the action from running
- Adds the `workflow_dispatch` trigger so the action can be manually run. This is helpful from a compliance standpoint as it lets us rebuild an image with the newest base image, picking up any recent security patches, even when there are no changes to be made in our Dockerfile.
  - Because of this, adds the current timestamp to the image tag. This ensures it's impossible to overwrite an existing tag if the action is manually run without changes to the Dockerfile

### DIFF
--- a/.github/workflows/push-selfcontained.yml
+++ b/.github/workflows/push-selfcontained.yml
@@ -22,7 +22,7 @@ jobs:
       run: az acr login -n qdkimages
     - name: Build and push Docker image
       run: |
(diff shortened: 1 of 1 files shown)


IsraelMiles:
Great catches! Thanks for improving this and making it clear how compliance is improved.

cgranade (approved):

PR #678, opened 2022-06-01
Fix typo in github action definition

anjbur:


cgranade (approved):

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,81 @@
+name: Docker build
```
cgranade, line 1: This is a neat way to organize things!

cgranade (approved):
LGTM. My only comment would be that some comments on both YAML files could help explain why it's organized like this and what docker-build.yml does, esp as that file mostly only makes sense after reading push-selfcontained.yml.

anjbur:
@cgranade Thank you for the feedback! I'll add some comments before merging then.

.github/workflows/docker-build.yml
anjbur: Thank you!

PR #676, opened 2022-06-01
Lock down python package versions for live tests

anpaz:

PR #679, opened 2022-06-02
Lock nbclient and ipython version on qsharp-base image

anpaz:
Similar to what we did for #676, we need to lock the versions of nbclient and ipython on our Docker image, otherwise we are facing deadlocks when running some of the Katas.


cgranade (approved):
LGTM, thanks for doing this!

PR #685, opened 2022-06-17
Update Python to 3.9

anpaz:
This updates Python to 3.9 for building and testing the QDK's qsharp packages.
Among these changes we also:
* Consolidate all PyPI requirements for building/testing the packages into `src/Python/requirements.txt`. 
* These requirements are installed automatically when invoking bootstrap.ps1
* Most versions of these requirements have been locked to prevent spurious build failures.
* CI steps for building conda packages have also been simplified to not be duplicated for different platforms.


build/test-selenium-requirements.txt
```diff
@@ -1,5 +1,2 @@
 pytest
-jupyter_client==6.1.7
-notebook==6.4.10
-pyzmq==19.0.2
-selenium==3.141
```
cgranade, line 5: Oh, it is nice not having all these requirement constraints!

conda-recipes/qsharp/meta.yaml
```diff
@@ -13,28 +13,19 @@ requirements:
...
   run:
     - python
-    - pyzmq
     - iqsharp={{ version }}
 
-  run_constrained:
-    # We forbid SciPy version 1.8.1 due to incompatibilities with some versions
-    # of PyZMQ and GCC, leading to runtime dynamic link failures of the form
-    # "version `GLIBCXX_3.4.30' not found (required by /miniconda/lib/python3.9/site-packages/scipy/fft/_pocketfft/pypocketfft.cpython-39-x86_64-linux-gnu.so)"
-    - scipy!=1.8.1 # [linux]
```
cgranade, line 28: I have run into the SciPy 1.8.1 bug even on versions other than 3.7; it's a problem with the DLL linking on their implementation of the FFT, such that we may want to keep the narrow carve out here.

cgranade (approved):
LGTM! Before taking out of draft, may be good to port `pack-selfcontained.ps1` and the iqsharp-base `Dockerfile`? Thanks for opening this!

build/test-selenium-requirements.txt
anpaz: they have all been consolidatated under src\Python\requirements.txt

conda-recipes/qsharp/meta.yaml
anpaz: I don't see the problem happening anymore, so I'm going to keep it out.

cgranade (approved):

conda-recipes/qsharp/meta.yaml
idavis: @cgranade @anpaz With buster going eol next month, might be worth moving the base image to `python:3.9-slim-bullseye`

cgranade: Fully agreed, @idavis; would be good to have this in for the next release as a result of that.

src/Python/qsharp-core/setup.py
```diff
@@ -61,8 +61,6 @@
         "Operating System :: OS Independent",
     ],
     install_requires=[
-        'jupyter_client<7.0',
```
ricardo-espinoza, line 64: I'm glad that those workarounds are no longer required. :)

ricardo-espinoza (approved):
Looks good. I didn't know this change started almost two months ago. :)

cgranade (approved):

PR #711, opened 2022-08-24
Downgrade severity of MSBuild messages and use heuristics.

cgranade:


xinyi-joffre (approved):

anpaz (approved):

cgranade:
Made one minor change, as it turns out that one of the NuGet APIs that we call was incorrectly marked as being non-nullable, causing the unit test failures above.

cgranade:
Made one minor change, as it turns out that one of the NuGet APIs that we call was incorrectly marked as being non-nullable, causing the unit test failures above.

PR #713, opened 2022-09-01
IQ# - Fullstate Simulator V1 Output Format

ScottCarda-MS:
Pass processor architecture into GenerateBitcode API. Use output format  instead of target to determine output format handling.


src/Core/Compiler/CompilerService.cs
```diff
@@ -414,6 +414,18 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source, ITaskRepo
             var (qsCompilation, assemblyConstants) = this.UpdateCompilation(sources, metadata.QsMetadatas, logger, compileAsExecutable, executionTarget, capability, parent: perfTask);
             logger.WarningCodesToIgnore.Remove(QsCompiler.Diagnostics.WarningCode.EntryPointInLibrary);
 
+            string processorArchitecture = "Any";
```
cgranade, line 417:
```suggestion
            var processorArchitecture = "Any";
```

src/Core/Compiler/CompilerService.cs
```diff
@@ -414,6 +414,18 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source, ITaskRepo
...
+            if (!(executionTarget is null) && executionTarget.ToLower() == "microsoft.simulator.fullstate")
```
cgranade, line 418: Comparing lower-cased strings to a string literal does not give a culture-invariant case-insensitive comparison; you likely want `string.Compare(..., StringComparison.InvariantCultureIgnoreCase)` here to indicate that you want to compare using the invariant culture. In that case, since `string.Compare` has signature `string.Compare(string?, string?, CultureInfo?)`, you should not need to check that `executionTarget` is null before calling `Compare`.

src/Core/Compiler/CompilerService.cs
```diff
@@ -454,7 +466,7 @@ public IEnumerable<QsNamespaceElement> IdentifyElements(string source, ITaskRepo
                     var tempPath = Path.GetTempFileName();
                     var bcFile = CompilationLoader.GeneratedFile(tempPath, Path.GetDirectoryName(tempPath), ".bc", "");
                     var diagnostics = new List<IRewriteStep.Diagnostic>();
-                    CompilationSteps.GenerateBitcode(qsCompilation, capability, bcFile, diagnostics: diagnostics);
+                    CompilationSteps.GenerateBitcode(qsCompilation, assebmlyConstants, bcFile, diagnostics: diagnostics);
```
cgranade, line 469: Is this misspelled?

cgranade (approved):
Some minor comments that would be good to address first, but otherwise looks good to me!

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: done

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: done

src/Core/Compiler/CompilerService.cs
ScottCarda-MS: fixed, and found that there is already an assemblyConstatns from the UpdateCompilation method. I moved the changes to the assemblyConstants to that function to keep it all in one place.

src/AzureClient/AzureClient.cs
```diff
@@ -8,6 +8,7 @@
 using System.IO;
 using System.Net;
 using System.Net.Http;
+using System.Runtime.InteropServices.ComTypes;
```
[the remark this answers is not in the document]
ScottCarda-MS: Not sure what added this, I've removed it.

ScottCarda-MS:
We are no longer pursuing this.

Issue #417, opened 2021-03-03
%azure.output says not complete, when job really failed


xinyi-joffre:
**Describe the bug**

I had a job that failed, and %azure.status and %azure.job correctly show it as failed.

When I use %azure.output, the error messages says that it's not complete:

![image](https://user-images.githubusercontent.com/20933017/109763029-540f2b80-7ba6-11eb-8dae-a29868ff527a.png)

> Job ID b0f2f564-ed44-4b4c-a98e-1415dfde084d has not completed. To check the status, call %azure.status with the job ID.

**Expected behavior**
%azure.output should also say that job is failed, rather than not complete.

**Screenshots**

**System information**
Component	Version
iqsharp	0.15.2101125897
Jupyter Core	1.5.0.0
.NET Runtime	.NETCoreApp,Version=v3.1

PR #715, opened 2022-09-09
Correctly report jobs as failed rather than in progress

anjbur:
It looks like this is a recurrence of #417 from a subsequent PR rolling back part of the fix. This restores that fix.


cgranade (approved):

anpaz (approved):

anjbur:
These changes were brought in as part of #727. Closing this PR.

PR #716, opened 2022-09-09
Pass target capabilities when compiling to QIR.

cgranade:


src/AzureClient/AzureExecutionTarget.cs
```diff
@@ -76,21 +76,58 @@ internal protected AzureExecutionTarget(string? targetId)
...
-        TargetCapabilityModule.Subsumes(this.MaximumCapability, capability);
+        // NB: Duplicates logic at https://github.com/microsoft/qsharp-compiler/blob/7714168fda4379fb7e6a6c616f680ec039c482f4/src/QuantumSdk/DefaultItems/DefaultItems.targets#L78,
+        //     but at the level of providers rather than at the level of resolved processors.
+        (GetProvider(TargetId) switch
+        {
+            AzureProvider.Quantinuum or AzureProvider.Honeywell => new[]
+            {
+                TargetCapabilityModule.AdaptiveExecution,
+                TargetCapabilityModule.BasicMeasurementFeedback,
+                TargetCapabilityModule.BasicExecution
```
bettinaheim, line 87: BasicQuantumFunctionality should also be listed here.

src/AzureClient/AzureExecutionTarget.cs
```diff
@@ -76,21 +76,58 @@ internal protected AzureExecutionTarget(string? targetId)
...
+            {
+                TargetCapabilityModule.BasicExecution
+            },
+            AzureProvider.Microsoft => new[]
+            {
+                TargetCapabilityModule.FullComputation
+            },
+            _ => new[]
+            {
+                TargetCapabilityModule.FullComputation
```
bettinaheim, line 108: All target capabilities should be listed here, i.e. add AdaptiveExecution, BasicExecution, BasicMeasurementFeedback and BasicQuantumFunctionality.

src/Tests/AzureClientTests.cs
```diff
@@ -198,6 +198,7 @@ public void TestManualTargets()
         [DataRow("--clear", "FullComputation", ExecuteStatus.Ok)]
         [DataRow("honeywell.mock", "FullComputation", ExecuteStatus.Error)]
         [DataRow("honeywell.mock", "BasicMeasurementFeedback", ExecuteStatus.Ok)]
+        [DataRow("quantinuum.mock", "AdaptiveExecution", ExecuteStatus.Ok)]
```
bettinaheim, line 201: What exactly is the test coverage that we have with this? I don't think this necessarily checks that QIR is submitted, does it? What test fails without the code changes in AzureClient.cs and AzureExecutionTarget.cs?

bettinaheim (changes requested):
Requesting changes based on the implemented behavior being different than the Sdk, and to get clarity on the testing question.

src/Tests/AzureClientTests.cs
cgranade: These tests check local usage of the IQ# kernel, for checking the QIR that gets submitted, I'd suggest adding to the live tests as well.

src/AzureClient/AzureExecutionTarget.cs
cgranade: These are checked against `Subsumes`, such that those four levels should also work here. Will add explicitly in any case.

cgranade:
@bettinaheim: Taken a pass at your feedback, would greatly appreciate another look. Thank you!

src/Tests/AzureClientTests.cs
bettinaheim: By live tests you mean tests that are not on this repo? Where are these?

bettinaheim (approved):
Approving to unblock the PR, but with the request to add a test to check that the submission works correctly (i.e. the QIR setup is used when either BasicExecution or AdaptiveExecution is the target capability) in the appropriate location.

src/Tests/AzureClientTests.cs
cgranade: They are in this repo, under the `tests.live/` folder, but those tests aren't run with the e2e pipeline — there's a separate pipeline that runs for live tests that pulls those integration tests in. As of 5601aeb4bb58dd47e32ae1cbb1571e69e93940e5, this PR now includes additional integration tests in that folder that ensure that teleportation runs successfully when the target is set to a Quantinuum API validator and when the target capability is `AdaptiveExecution`.

PR #721, opened 2022-09-15
Fix split of IonQ tests into projects.

cgranade:


tests.live/Python/test_live.py
```diff
@@ -79,7 +79,6 @@ def test_ionq_submit(self):
         """
         import time
         import qsharp
-        qsharp.projects.add(path.join(path.dirname(__file__), "qsharp", "ionq", "IonQ.csproj"))
```
cgranade, line 82: This is now injected via test fixtures, and should not be separately included here.

anpaz (approved):

PR #726, opened 2022-09-19
Expose output format

anjbur:
Cherry picked changes from #713 to expose a job's output format.


ScottCarda-MS (approved):
Thanks for migrating these changes out of the other project!

PR #727, opened 2022-09-19
Changes for new estimation target

anjbur:


src/AzureClient/AzureClient.cs
```diff
@@ -129,6 +129,7 @@ public class AzureClient : IAzureClient
                 baseEngine.RegisterDisplayEncoder(new AzureClientErrorToTextEncoder());
                 baseEngine.RegisterDisplayEncoder(new DeviceCodeResultToHtmlEncoder());
                 baseEngine.RegisterDisplayEncoder(new DeviceCodeResultToTextEncoder());
+                baseEngine.RegisterDisplayEncoder(new ResourceEstimationToHtmlEncoder(logger));
             }
```
cgranade, line 133: I'd suggest using ActivatorUtils here to inject dependencies, rather than manually creating a logger. That will make sure the logger has the right metadata as to which service is generating log entries.

src/AzureClient/Visualization/ResourceEstimationEncoder.cs
```diff
@@ -0,0 +1,120 @@
...
+                ? newValue
+                : throw new JsonException($"Malformed JSON. Failed at '{p}' to retrieve value for '{path}'");
+        }
+        return value;
+    }
+}
+
+internal static class ResourceEstimationResultExtensions
+{
+    internal static ResourceEstimationResult ToResourceEstimationResults(this Stream stream)
```
cgranade, line 32: I'd suggest an expression-body method here.

src/Jupyter/Extensions.cs
```diff
@@ -202,6 +202,18 @@ public static bool TryDecodeParameter<T>(this Dictionary<string, string> paramet
                 parameterValue = parameterValue.Substring(1, parameterValue.Length - 2);    // Strip off the enclosing quotes
                 parameterValue = parameterValue.Replace("\\\"", "\"");                      // Don't escape the interior quotes
             }
+
+            // This is a temporary fix in order to send JSON encoded data as strings to the Azure submission client
+            var intermediate = JsonConvert.DeserializeObject<ImmutableDictionary<string, object>>(parameterValue);
+            if (intermediate == null) {
```
cgranade, line 208: Nit, but please use next-line brace style here.

src/Jupyter/Extensions.cs
```diff
@@ -202,6 +202,18 @@ public static bool TryDecodeParameter<T>(this Dictionary<string, string> paramet
                 parameterValue = parameterValue.Substring(1, parameterValue.Length - 2);    // Strip off the enclosing quotes
                 parameterValue = parameterValue.Replace("\\\"", "\"");                      // Don't escape the interior quotes
             }
+
+            // This is a temporary fix in order to send JSON encoded data as strings to the Azure submission client
```
cgranade, line 206: Is there an issue tracking when this temporary workaround can be removed?

src/Jupyter/Extensions.cs
cgranade, line 208: I'd suggest not using `==` to compare to null. The `is` operator is more reliable as it cannot be overriden by operator overloading.

src/Jupyter/Extensions.cs
```diff
@@ -202,6 +202,18 @@ public static bool TryDecodeParameter<T>(this Dictionary<string, string> paramet
...
+            var intermediate = JsonConvert.DeserializeObject<ImmutableDictionary<string, object>>(parameterValue);
+            if (intermediate == null) {
+                return defaultValue;
+            } else {
+                var transformedIntermediate = new Dictionary<string, string>();
+                foreach (var (key, value) in intermediate) {
```
cgranade, line 212: I'd suggest using LINQ's ToDictionary here instead of accumulating manually.

cgranade (approved):
LGTM! Just some minor comments here and there.

src/Jupyter/Extensions.cs
anjbur: @cgranade it looks like `ToDictionary` takes an `Enumerable<T>` and turns it into a dictionary, which won't work here as `intermediate` is already a dictionary. Unless there's another LINQ method I'm overlooking?

cgranade: Dictionaries implement `IEnumerable` over key-value pairs, such that the following should work:
```csharp
var transformed = intermediate
    .ToDictionary(pair => pair.Key, pair => pair.Value.ToString())
    .ToImmutableDictionary();
```
You can even use `ToImmutableDictionary` directly, as it has an overload that takes key and value projectors (https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary.toimmutabledictionary?view=net-6.0#system-collections-immutable-immutabledictionary-toimmutabledictionary-3(system-collections-generic-ienumerable((-0))-system-func((-0-1))-system-func((-0-2)))):
```csharp
var transformed = intermediate
    .ToImmutableDictionary(pair => pair.Key, pair => pair.Value.ToString());
```
(Please forgive typos that may prevent above from running as written; didn't test these suggestions yet.)

src/AzureClient/Visualization/ResourceEstimationEncoder.cs
anjbur: Done!

src/Jupyter/Extensions.cs
anjbur: Done!

src/Jupyter/Extensions.cs
anjbur: Aha I see, thank you for clarifying! I'll get that change up momentarily then.

cgranade (approved):

src/Jupyter/Extensions.cs
anjbur: I just added an issue to track this at #729

PR #734, opened 2022-10-05
Support arrays as arguments for submission to Azure Quantum

cesarzc:
This change adds support in IQ# to use arrays as arguments when submitting programs to Azure Quantum.

### DIFF
--- a/src/AzureClient/EntryPoint/EntryPoint.cs
+++ b/src/AzureClient/EntryPoint/EntryPoint.cs
@@ -115,10 +115,34 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             {
                 return new ArgumentValue.Result(Newtonsoft.Json.JsonConvert.DeserializeObject<Result>(parameterValue)!);
             }
+            else if (parameterType.IsQArray())
+            {
+                var arrayType = parameterType.GenericTypeArguments.FirstOrDefault() ??
+                    throw new ArgumentException($""); // TODO: Set right error message.
+
+                if (arrayType == typeof(bool)) {
+                    
+                }
+                else if (arrayType == typeof(long))
+                {
+                    var values = Newtonsoft.Json.JsonConvert.DeserializeObject<long[]>(parameterValue)
+                        .Select(v => new ArgumentValue.Int(v));
+                    return ArgumentValue.Array.TryCreate(values.ToImmutableArray<ArgumentValue>(), ArgumentType.Int);
+                    //throw new ArgumentException($"cesarzc exception long | Values.Count: {values.Count()}");
+                }
+                
+                throw new ArgumentException($"cesarzc exception | ParameterValue: {parameterValue}, ArrayType: {arrayType} " +
+                $"HasElementType: {parameterType.HasElementType}, GetElementType: {parameterType.GetElementType()} " +
+                $"First: {parameterType.GenericTypeArguments.First()}");
+            }
             // TODO (cesarzc): here's where the change in the IQ# side needs to be done.
             else
             {
-                throw new ArgumentException($"The given type of {parameterType.Name} is not supported."); ;
+                throw new ArgumentException($"The given type of {parameterType.Name} is not supported. " +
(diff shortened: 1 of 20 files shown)


src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -115,9 +115,62 @@ private ArgumentValue GetArgumentValue(System.Reflection.ParameterInfo parameter
             {
                 return new ArgumentValue.Result(Newtonsoft.Json.JsonConvert.DeserializeObject<Result>(parameterValue)!);
             }
+            else if (parameterType.IsQArray())
+            {
+                var arrayType = parameterType.GenericTypeArguments.FirstOrDefault() ??
+                    throw new ArgumentException($"Could not get the type of array.");
+                if (arrayType == typeof(bool)) {
```
cgranade, line 122: Style nit, but next-line brace style is preferred for C# code.
```suggestion
                if (arrayType == typeof(bool))
                {
```

cgranade, line 122: Is there a way to consolidate this into a switch statement or expression?

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -115,9 +115,62 @@ private ArgumentValue GetArgumentValue(System.Reflection.ParameterInfo parameter
...
+                    return ArgumentValue.Array.TryCreate(values, ArgumentType.Bool) ??
+                        throw new ArgumentException($"Could not create array of Bool for {parameter.Name}.");
+                }
+                else if (arrayType == typeof(double))
+                {
+                    var values = Newtonsoft.Json.JsonConvert.DeserializeObject<double[]>(parameterValue)
+                        .Select(v => new ArgumentValue.Double(v))
+                        .ToImmutableArray<ArgumentValue>();
+                    return ArgumentValue.Array.TryCreate(values, ArgumentType.Double) ??
+                        throw new ArgumentException($"Could not create array of Bool for {parameter.Name}.");
```
cgranade, line 135: This seems mostly duplicated with the version for `bool`, is it possible to extract into an extension method or similar?

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -115,9 +115,62 @@ private ArgumentValue GetArgumentValue(System.Reflection.ParameterInfo parameter
...
+                else if (arrayType == typeof(Result))
+                {
+                    var values = Newtonsoft.Json.JsonConvert.DeserializeObject<Result[]>(parameterValue)
+                        .Select(v => new ArgumentValue.Result(v))
+                        .ToImmutableArray<ArgumentValue>();
+                    return ArgumentValue.Array.TryCreate(values, ArgumentType.Result) ??
+                        throw new ArgumentException($"Could not create array of Result for {parameter.Name}.");
+                }
+
+                throw new ArgumentException($"Invalid array type {arrayType}.");
```
cgranade, line 169: This error message seems to imply that an array (for instance) of type `Double[][]` is invalid rather than unsupported for use as a parameter to an entry point. That is, `Double[][]` is a valid type even if not supported in that context.

cgranade (approved):
LGTM, but had some minor comments that may be good to consider before merging.

src/AzureClient/EntryPoint/EntryPoint.cs
cesarzc: That would be ideal, but unfortunately since we make use of `typeof` we can't use a switch expression.

src/AzureClient/EntryPoint/EntryPoint.cs
cesarzc: Change the message to state that it is unsupported rather than invalid.

src/AzureClient/EntryPoint/EntryPoint.cs
cesarzc: It is very similar, but in this case the string is deserialized into an array of `bool` which would cause the extension method to not be exactly the same.

PR #737, opened 2022-10-11
Update resource estimator tooltips to HTML

anjbur:
Adds the tooltip when hovering over any part of the table row.
Screen Shot 2022-10-11 at 3 09 05 PM

### DIFF
--- a/src/AzureClient/Visualization/ResourceEstimationEncoder.cs
+++ b/src/AzureClient/Visualization/ResourceEstimationEncoder.cs
@@ -72,7 +72,7 @@ public record class ResourceEstimationToHtmlEncoder(ILogger? Logger = null) : IR
                             color: #fff;
                             text-align: center;
                             border-radius: 6px;
-                            padding: 5px 0;
+                            padding: 5px;
                             position: absolute;
                             z-index: 1;
                             top: 150%;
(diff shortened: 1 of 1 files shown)


src/AzureClient/Visualization/ResourceEstimationEncoder.cs
```diff
@@ -74,9 +109,9 @@ public record class ResourceEstimationToHtmlEncoder(ILogger? Logger = null) : IR
                     {
                         var value = result.GetValueFromPath(entry.GetValue<string>("path"));
                         sb.AppendLine($@"
-                            <tr title='{entry["description"]}'>
+                            <tr class=aqre-tooltip>
                                 <td>{entry["label"]}</td>
-                                <td>{value}</td>
+                                <td class=aqre-tooltip>{value}<span class=aqre-tooltiptext>{entry["description"]}</span></td>
```
msoeken, line 114: Like this the tooltip is only accessible via the second column. It would be better to have it on the whole row.

msoeken (changes requested):
Thanks, left some comment for discussion

src/AzureClient/Visualization/ResourceEstimationEncoder.cs
anjbur: When I tested this, it was accessible over the whole row!

msoeken: You're right, I saw it wrong.

msoeken (approved):

PR #755, opened 2023-01-23
Updated tests not to exepct errors on hardware capabilities

DmitryVasilevsky:
This corresponds to https://github.com/microsoft/qsharp-compiler/pull/1595


src/Tests/AzureClientTests.cs
```diff
@@ -404,7 +404,7 @@ public async Task TestRuntimeCapabilities(string targetId, AzureClientError? exp
             Assert.IsNotNull(azureWorkspace);
             azureWorkspace?.AddProviders("ionq", "quantinuum");
 
-            // Verify that IonQ job fails to compile (QPRGen0)
+            // Verify that IonQ job should generate warnings (QPRGen0)
```
samarsha, line 407: It's really just verifying that submission succeeds, there's no assertion for warnings.

samarsha (commented):
For VS Code/CLI development, we have the project templates to turn the warnings back into errors for some targets. But for IQ#, it looks like they're always warnings. Are we OK with that difference?

src/Tests/AzureClientTests.cs
DmitryVasilevsky: Updated comment.

DmitryVasilevsky:
> For VS Code/CLI development, we have the project templates to turn the warnings back into errors for some targets. But for IQ#, it looks like they're always warnings. Are we OK with that difference?

I believe this was the goal (after discussion with Cesar).

samarsha (approved):

PR #758, opened 2023-01-25
Updated iQsharp tests to expect non-failures for hardware caps - matching branch

DmitryVasilevsky:


DmitryVasilevsky:
No longer needed.

PR #762, opened 2023-02-06
Update package-lock.json

minestarks:
Run `npm install` to bump the dependencies in package-lock.json.

My main motivation is to update [the dependency on `canvas`](https://github.com/microsoft/iqsharp/blob/23f392d9049edc9bf5d956282237126ccaef50b6/src/Kernel/package.json#L10). The previously pinned version of `canvas` (2.8.0) no longer had a [published binary](https://github.com/Automattic/node-canvas/releases/tag/v2.8.0) that was compatible with the latest build of Node.js. This would trigger a build of that dependency from source, adding all sorts of C++ prerequisites to our build. Bumping `canvas` to the latest compatible version (2.11.0) allows the [prebuilt binaries](https://github.com/Automattic/node-canvas/releases) to be downloaded again, eliminating the need to build from source.

### DIFF
--- a/build/test-selenium.py
+++ b/build/test-selenium.py
@@ -105,7 +105,7 @@ def test_trace_magic():
     assert len(outputs) > 0
 
     # Verify expected text output
-    expected_trace = '|0\u27E9 q0 H H' # \u27E9 is mathematical right angle bracket
+    expected_trace = '|0\u27E9q0HH' # \u27E9 is mathematical right angle bracket
     WebDriverWait(nb.browser, 60).until(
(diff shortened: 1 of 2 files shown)


kuzminrobin (approved):

minestarks:
Note this bumped the `quantum-viz.js` package from 1.0.2 -> 1.0.5 ([changelog](https://github.com/microsoft/quantum-viz.js/compare/v1.0.2...v1.0.5)). This accounts for the Selenium test failure in CI - but it's benign, just whitespace changes in generated HTML.

PR #766, opened 2023-02-10
Deprecating old (C#-based) standalone ResourcesEstimator

DmitryVasilevsky:
Standalone ResourcesEstimator is deprecated in February 2023 release and will be removed in March 2023 release of the QDK. Users are advised to use Azure-based resources estimator.

* Added output message to the %estimate magic command.

### DIFF
--- a/src/Jupyter/Magic/EstimateMagic.cs
+++ b/src/Jupyter/Magic/EstimateMagic.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
             qsim.OnDisplayableDiagnostic += channel.Display;
             qsim.DisableLogToConsole();
 
-            channel.Stdout("The local Resources Estimator will be removed in the March 2023. Use the new Azure Resource Estimator.");
+            channel.Stdout("The local Resources Estimator will be removed in March 2023. The Resources Estimator is now available through Azure Quantum.");
             await symbol.Operation.RunAsync(qsim, inputParameters);
 
             return qsim.Data.ToExecutionResult();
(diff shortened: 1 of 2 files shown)


src/Jupyter/Magic/EstimateMagic.cs
```diff
@@ -114,6 +114,7 @@ public async Task<ExecutionResult> RunAsync(string input, IChannel channel)
             qsim.OnDisplayableDiagnostic += channel.Display;
             qsim.DisableLogToConsole();
 
+            channel.Stdout("The Resources Estimator will be removed in the March 2023 QDK release. Please use the Azure-based version instead.");
```
msoeken, line 117: I suggest to add a link that helps users to know how to use the Azure-based version.

msoeken (approved):

src/Jupyter/Magic/EstimateMagic.cs
DmitryVasilevsky: This is an approved wording. I will merge it "as is" because it is sufficient. There was a concern about putting a link in the message. We will update it if we agree on a better wording (with the link).

ffrachon:
- In the message. Could you add "Learn more at https://aka.ms/AQ/RE".
- Change Resources from plural to singular --> Resource Estimator.

DmitryVasilevsky:
> * In the message. Could you add "Learn more at https://aka.ms/AQ/RE".
> * Change Resources from plural to singular --> Resource Estimator.

@ffrachon, These exact questions were discussed with our PM. As for the link, I was told that putting a link could be a security risk as it appears in the console. As for plural to singular - it is plural in the code and in most of the documentation. There maybe one or two singular instances in the documentation. It is singular in the colloquial speech. As we are deprecating the old local resources estimator, we use plural. The new resource estimator will likely be singular.

PR #720, opened 2022-09-15
Pin pyzmq install version, other dependencies

anpaz:
To workaround: https://github.com/zeromq/pyzmq/issues/1764

PR #771, opened 2023-02-21
Updated Python Dependency Versions

ScottCarda-MS:
Update python dependencies to mitigate py 1.11.0 vulnerability.


kuzminrobin (approved):

src/Python/requirements.txt
```diff
@@ -2,11 +2,11 @@
 ipython==8.10.0      # to avoid high-severity issue.
 jupyter==1.0.0
 nbconvert==7.0.0
-notebook==6.4.12
+notebook==6.5.2
 numpy==1.22.0
 pytest==7.2.0
 pygments==2.12.0    # to avoid high-severity issue.
-pyzmq==23.2.1       # to workaround: https://github.com/zeromq/pyzmq/issues/1764
+pyzmq==25.0.0       # to workaround: https://github.com/zeromq/pyzmq/issues/1764
```
minestarks, line 9: This comment seems obsolete now? Seems like the pin to version 23.x was added [here](https://github.com/microsoft/iqsharp/pull/720) to work around a specific build issue, which sounds like isn't relevant anymore.

ScottCarda-MS: Yes, you are right. This comment seems to be outdated now. Since this PR is already complete, I will remove it in this PR: https://github.com/microsoft/iqsharp/pull/772

PR #773, opened 2023-02-27
Add missing __init__.py file

xinyi-joffre:
Add missing __init__.py file to results folder to fix import errors for `import qsharp.azure`

### DIFF
--- a/src/Python/qsharp-core/setup.py
+++ b/src/Python/qsharp-core/setup.py
@@ -61,6 +61,8 @@ setuptools.setup(
         "Operating System :: OS Independent",
     ],
     install_requires=[
-        'jupyter_client>=7.3.4'
+        'jupyter_client>=7.3.4',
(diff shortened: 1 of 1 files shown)


ScottCarda-MS:
Are we confident the file should be empty? It doesn't need to expose any methods externally?

xinyi-joffre:
> Are we confident the file should be empty? It doesn't need to expose any methods externally?

Yes, the code is not importing from .result directly. It is importing from the file under the result module:

`from .results.resource_estimator import ResourceEstimatorResult, ResourceEstimatorBatchResult`

Exports in the __init__.py file are only needed if you are trying to give shortcut to import from .results directly.

ScottCarda-MS (approved):

minestarks:
@xinyi-joffre @ScottCarda-MS should this go into `main` as well? Just noticed the import errors in my local build.

PR #774, opened 2023-02-27
Bump qutip 4.7.0 -> 4.7.1

minestarks:
Building the repo with Python 3.11 causes a failure since a [prebuilt wheel does not exist for QuTip 4.7.0 x Python 3.11](https://github.com/qutip/qutip/releases/tag/v4.7.0). Instead of trying to debug the failure, bumping the dependency so the prebuilt wheel will be fetched instead.


minestarks:
Got failures; sounds like build agents use 3.9 anyway. Abandoning

PR #775, opened 2023-02-27
Updated doc generated for %estimate

DmitryVasilevsky:
Added deprecation message.

[!IMPORTANT] doesn't seem to be supported so the message won't be in the blue box.

### DIFF
--- a/src/Jupyter/Magic/EstimateMagic.cs
+++ b/src/Jupyter/Magic/EstimateMagic.cs
@@ -47,11 +47,10 @@ namespace Microsoft.Quantum.IQSharp.Jupyter
                     - Arguments for the Q# operation or function must also be specified as `key=value` pairs.
                 ".Dedent(),
                 Remarks = @"
-                    [!IMPORTANT]
                     **QDK warning message**: ""The local Resources Estimator will be removed in March 2023.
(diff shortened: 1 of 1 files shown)


src/Jupyter/Magic/EstimateMagic.cs
```diff
@@ -46,6 +46,14 @@ public class EstimateMagic : AbstractMagic
                     or function name that has been defined either in the notebook or in a Q# file in the same folder.
                     - Arguments for the Q# operation or function must also be specified as `key=value` pairs.
                 ".Dedent(),
+                Remarks = @"
+                    **QDK warning message**: ""The local Resources Estimator will be removed in March 2023.
+                    The Resources Estimator is now available through Azure Quantum.""
+
+                    The `%estimate` magic command is deprecated. To compute physical and logical resource estimation
+                    and runtime, we recommend using the
+                    [Azure Quantum Resource Estimator](https://learn.microsoft.com/en-us/azure/quantum/intro-to-resource-estimation) tool.
```
ScottCarda-MS, line 55: Please remove the /en-us part of the url.

ScottCarda-MS (approved):

PR #776, opened 2023-02-28
Make DisplayableException more serialization-friendly

minestarks:
Fixes https://dev.azure.com/ms-quantum/Quantum%20Program/_workitems/edit/44472

Serializing `DisplayableException` to JSON causes an `OutOfMemoryException` to be thrown. 

JSON tracing shows hundreds of objects referenced through `StackFrame` attempted to be serialized:

```
Newtonsoft.Json Information: 0 : Finished serializing System.RuntimeType. Path 'StackTrace[0].Callable.__Body__.Method.Module.Assembly.DefinedTypes[263]'.
Newtonsoft.Json Information: 0 : Started serializing System.RuntimeType. Path 'StackTrace[0].Callable.__Body__.Method.Module.Assembly.DefinedTypes[263]'.
Newtonsoft.Json Information: 0 : Finished serializing System.RuntimeType. Path 'StackTrace[0].Callable.__Body__.Method.Module.Assembly.DefinedTypes[264]'.
Newtonsoft.Json Information: 0 : Started serializing System.RuntimeType. Path 'StackTrace[0].Callable.__Body__.Method.Module.Assembly.DefinedTypes[264]'.
```

On principle I don't think an object designed to be serialized/displayed should have references to complicated types such as `StackFrame`. So I'm introducing a `DisplayableStackFrame`, a simple struct.

Additionally I'm cleaning up a few areas that I thought looked odd - where we were serializing the same string slightly differently, unnecessary null handling, etc.

Considered, and rejected:
 - Adding a custom JSON converter for `DisplayableException`: designing the object to be simple in the first place seems like a more sensible approach
 - Avoid serializing to JSON in the first place: This turned out to be way trickier than it sounds. Serializing results to JSON seems to be pretty baked into the IQ# engine, and the JSON results are used by the Python wrapper (namely in [`capture_diagnostics`](https://github.com/microsoft/iqsharp/blob/d90578bb70907aa9bdcb4e1234c13f19ea0e27dc/src/Python/qsharp-core/qsharp/clients/iqsharp.py#L225))

### DIFF
--- a/src/Jupyter/Extensions.cs
+++ b/src/Jupyter/Extensions.cs
@@ -103,11 +103,7 @@ public static class Extensions
         simulator.DisableExceptionPrinting();
         simulator.OnException += (exception, stackTrace) =>
         {
-            channel.Display(new DisplayableException
-            {
-                Exception = exception,
-                StackTrace = stackTrace
-            });
+            channel.Display(DisplayableException.Create(exception, stackTrace));
         };
         return simulator;
     }
--- a/src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
+++ b/src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
@@ -2,43 +2,47 @@
 // Licensed under the MIT License.
 #nullable enable
 
-using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Collections.Immutable;
+using Microsoft.Quantum.Simulation.Common;
+using Microsoft.Quantum.Simulation.Core;
 using System.Data;
-using System.Linq;
 using System.Net;
 using System.Text;
 using System.Text.RegularExpressions;
-using Microsoft.Jupyter.Core;
-using Microsoft.Quantum.QsCompiler.SyntaxTree;
-using Microsoft.Quantum.Simulation.Common;
-using Microsoft.Quantum.Simulation.Core;
-using Microsoft.Quantum.Simulation.Simulators;
 
 namespace Microsoft.Quantum.IQSharp
 {
-    internal record struct DisplayableException(Exception Exception, IEnumerable<StackFrame>? StackTrace)
+    internal record struct DisplayableStackFrame(string Callable, string SourceFile, string BestSourceLocation, int LineNumber)
     {
-        public string Header =>
-            $"Unhandled exception. {Exception.GetType().FullName}: {Exception.Message}";
     }
 
-    internal static class StackFrameExtensions
+    internal record struct DisplayableException(string? ExceptionType, string ExceptionMessage, IEnumerable<DisplayableStackFrame> StackTrace)
     {
-        internal static string ToFriendlyName(this ICallable callable)
+        public static DisplayableException Create(Exception exception, IEnumerable<StackFrame> stackTrace)
+        {
+            return new DisplayableException
+            {
+                ExceptionType = exception.GetType().FullName,
+                ExceptionMessage = exception.Message,
+                StackTrace = stackTrace.Select(sf => new DisplayableStackFrame
+                {
+                    Callable = GetCallableFriendlyName(sf.Callable),
+                    SourceFile = sf.SourceFile,
+                    BestSourceLocation = sf.GetBestSourceLocation(),
+                    LineNumber = sf.FailedLineNumber
+                })
+            };
+        }
+
+        public string Header =>
+            $"Unhandled exception. {ExceptionType}: {ExceptionMessage}";
+
+        private static string GetCallableFriendlyName(ICallable callable)
         {
             var fullName = callable.FullName;
             return fullName.StartsWith(Snippets.SNIPPETS_NAMESPACE)
             ? fullName.Substring(Snippets.SNIPPETS_NAMESPACE.Length + 1)
             : fullName;
         }
-
-        internal static string ToSourceLink(this StackFrame frame) =>
-            Regex.Match(frame.SourceFile, "snippet_[0-9]*.qs$").Success
-            ? "(notebook)"
-            : $"<a href=\"{frame.GetBestSourceLocation()}\">{frame.SourceFile}:{frame.FailedLineNumber}</a>";
     }
 
     /// <summary>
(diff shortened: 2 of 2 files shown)


src/Jupyter/Extensions.cs
```diff
@@ -103,11 +103,7 @@ public static T WithStackTraceDisplay<T>(this T simulator, IChannel channel)
         simulator.DisableExceptionPrinting();
         simulator.OnException += (exception, stackTrace) =>
         {
-            channel.Display(new DisplayableException
-            {
-                Exception = exception,
-                StackTrace = stackTrace
-            });
+            channel.Display(DisplayableException.Create(exception, stackTrace));
```
minestarks, line 106: The `DisplayableException` here will end up [being serialized](https://github.com/microsoft/jupyter-core/blob/4d9233bdede5c0b7135d4e0fc8e4ecb24a54ab4d/src/Engines/BaseEngine.cs#L357) into various formats (text, html and JSON) to be included in a message to sent back to the Jupyter frontend.

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -2,27 +2,38 @@
...
 namespace Microsoft.Quantum.IQSharp
 {
-    internal record struct DisplayableException(Exception Exception, IEnumerable<StackFrame>? StackTrace)
+    internal record struct DisplayableStackFrame(string Callable, string SourceFile, string BestSourceLocation, int LineNumber)
     {
+    }
+
+    internal record struct DisplayableException(string? ExceptionType, string ExceptionMessage, IEnumerable<DisplayableStackFrame> StackTrace)
+    {
+        public static DisplayableException Create(Exception exception, IEnumerable<StackFrame> stackTrace)
```
minestarks, line 19: The problematic type is [`StackFrame`](https://github.com/microsoft/qsharp-runtime/blob/09403e3d306567f2cff237881dbb064829b496a5/src/Simulation/Simulators/CommonNativeSimulator/StackTrace.cs#L17), which has a reference to `Callable`, which can have references to `Assembly` objects, etc., etc., which ultimately fan out into hundreds of types to be serialized. We shouldn't have this type in a class that's intended to be serialized.

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -74,10 +90,12 @@ public class DisplayableExceptionToHtmlEncoder : IResultEncoder
                         {String.Join("\n", rows)}
                     </tbody>
                 ";
+
+                // { display: list-item } style is needed for a <summary> tag so that the arrow shows up
```
minestarks, line 94: Jupyter's stylesheets (I think) were causing the disclosure triangle to be hidden, so it wasn't clear you could click on the element to see the callstack.

With this change, it should look like:

image

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#default_style

src/Jupyter/Extensions.cs
minestarks: Jupyter only displays the HTML output, but the Python wrapper uses [JSON results](https://github.com/microsoft/iqsharp/blob/d90578bb70907aa9bdcb4e1234c13f19ea0e27dc/src/Python/qsharp-core/qsharp/clients/iqsharp.py#L388), which is why I can't just simply disable JSON serialization for this type.

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -35,10 +46,10 @@ internal static string ToFriendlyName(this ICallable callable)
             : fullName;
         }
 
-        internal static string ToSourceLink(this StackFrame frame) =>
```
DmitryVasilevsky, line 38: Should this be an extension method or could it be moved to the class?

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -54,15 +65,20 @@ public class DisplayableExceptionToHtmlEncoder : IResultEncoder
...
+                // upon the failure event raised by the simulator
+                System.Diagnostics.Debug.Assert(ex.StackTrace != null);
+
                 var rows = ex
                     .StackTrace
-                    ?.Select(frame => $@"
+                    .Select(frame => $@"
                         <tr>
                             <td>{frame.ToSourceLink()}</td>
-                            <td>{frame.Callable.ToFriendlyName()}</td>
```
DmitryVasilevsky, line 62: Do we still need .ToFriendlyName() extension method on ICallable?

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -74,10 +90,12 @@ public class DisplayableExceptionToHtmlEncoder : IResultEncoder
...
                     </tbody>
                 ";
+
+                // { display: list-item } style is needed for a <summary> tag so that the arrow shows up 
                 var table = $@"
                     <details>
-                        <summary>
-                            Unhandled exception of type {ex.Exception.GetType().FullName}: {ex.Exception.Message}
+                        <summary style=""display:list-item"">
+                            {ex.Header}
```
DmitryVasilevsky, line 98: Should it be HTML Encoded?

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
```diff
@@ -107,11 +125,11 @@ public class DisplayableExceptionToTextEncoder : IResultEncoder
                 var builder = new StringBuilder();
                 builder.AppendLine(ex.Header);
                 var first = true;
-                foreach (var frame in ex.StackTrace ?? Enumerable.Empty<StackFrame>())
+                foreach (var frame in ex.StackTrace ?? Enumerable.Empty<DisplayableStackFrame>())
```
DmitryVasilevsky, line 128: In the code above ex.StackTrace is accessed without the null-coalescing operator (??). Do we need it here?

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
minestarks: Good point, no reason for the extension class anymore

src/Jupyter/Visualization/DisplayableExceptionEncoders.cs
minestarks: No, good point, the assumptions should be the same.

cesarzc (approved):

PR #783, opened 2023-03-15
Add support for passing Target Capabilities in QIR generation and job submission

vxfield:
The [first commit](https://github.com/microsoft/iqsharp/commit/0f592413744ec5eca44b93f7cdebb746702d89ac) fixes the issue when submitting a QIR job and when the `targetCapability` parameter was not being sent to the Azure Quantum service.

The [second commit](https://github.com/microsoft/iqsharp/commit/f9ce9cb4dfc671d13dded89104c08ec3a7d9f4f3) fixes a limitation when generating QIR from a Q# callable, in which it was not possible to pass the `target` or the `target capability`.

It also introduces a `_repr_qir_()` method to the Q# callable, which returns the QIR bitcode (instead of returning the QIR text as in the `as_qir()` method does).
The `_repr_qir_()` will be used by the `azure-quantum` Python package to submit Q# jobs.

The last 2 commits add some telemetry events for QirMagic and EntryPointSubmission.

### DIFF
--- a/src/AzureClient/EntryPoint/EntryPoint.cs
+++ b/src/AzureClient/EntryPoint/EntryPoint.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System.IO;
+using System.Reflection.PortableExecutable;
 using System.Threading;
 using Microsoft.Extensions.Logging;
 using Microsoft.Quantum.QsCompiler;
@@ -13,6 +14,16 @@ using Microsoft.Quantum.Simulation.Core;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
+    public record EntryPointSubmitArgs
+    {
+        public string? MachineName { get; set;  }
+        public string? Target { get; set; }
+        public string? TargetCapability { get; set; }
+        public string? JobId { get; set; }
+    }
+
+    public record EntryPointSubmitEvent : Event<EntryPointSubmitArgs>;
+
     /// <inheritdoc/>
     internal class EntryPoint : IEntryPoint
     {
@@ -24,12 +35,13 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         private Type OutputType { get; }
         private OperationInfo OperationInfo { get; }
         private ILogger? Logger { get; }
+        private IEventService? EventService { get; }
 
         /// <inheritdoc/>
         public Stream? QirStream { get; }
         /// <inheritdoc/>
-        public TargetCapability? TargetCapability { get; } 
- 
+        public TargetCapability? TargetCapability { get; }
+
         /// <summary>
         /// Creates an object used to submit jobs to Azure Quantum.
         /// </summary>
@@ -41,6 +53,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         /// <see cref="EntryPointInfo{I,O}"/> object provided as the <c>entryPointInfo</c> argument.</param>
         /// <param name="operationInfo">Information about the Q# operation to be used as the entry point.</param>
         /// <param name="logger">Logger used to report internal diagnostics.</param>
+        /// <param name="eventService">EventService used to report telemetry.</param>
         /// <param name="qirStream">
         ///     Stream from which QIR bitcode for the entry point can be read.
         /// </param>
@@ -50,8 +63,9 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
         public EntryPoint(object entryPointInfo,
                           Type inputType, 
                           Type outputType, 
-                          OperationInfo operationInfo, 
-                          ILogger? logger, 
+                          OperationInfo operationInfo,
+                          ILogger? logger,
+                          IEventService? eventService,
                           Stream? qirStream = null,
                           TargetCapability? targetCapability = null)
         {
@@ -62,6 +76,7 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
             Logger = logger;
             QirStream = qirStream;
             TargetCapability = targetCapability;
+            EventService = eventService;
         }
 
         private object GetEntryPointInputObject(AzureSubmissionContext submissionContext)
@@ -248,7 +263,19 @@ namespace Microsoft.Quantum.IQSharp.AzureClient
                     && method.GetParameters()[2].ParameterType == typeof(IQuantumMachineSubmissionContext))
                 .MakeGenericMethod(new Type[] { InputType, OutputType });
             var submitParameters = new object[] { EntryPointInfo, entryPointInput, submissionContext };
-            return (Task<IQuantumMachineJob>)submitMethod.Invoke(machine, submitParameters);
+
+            return ((Task<IQuantumMachineJob>)submitMethod.Invoke(machine, submitParameters))
+                .ContinueWith<IQuantumMachineJob>(jobTask =>
+                {
+                    var job = jobTask.Result;
+                    EventService?.Trigger<EntryPointSubmitEvent, EntryPointSubmitArgs>(new EntryPointSubmitArgs()
+                    {
+                        MachineName = machine.GetType().FullName,
+                        JobId = job.Id,
+                        Target = machine.Target
+                    });
+                    return job;
+                });
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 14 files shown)


src/Tool/Telemetry.cs
```diff
@@ -272,6 +281,33 @@ public static EventProperties SetCommonProperties(this EventProperties evt)
             return evt;
         }
 
+        public static EventProperties AsTelemetryEvent(this QirMagicArgs qirMagicArgs)
+        {
+            var evt = new EventProperties() { Name = "QirMagic".WithTelemetryNamespace() };
+            evt.SetProperty("CapabilityName".WithTelemetryNamespace(), qirMagicArgs.CapabilityName);
+            evt.SetProperty("CapabilityInsuficient".WithTelemetryNamespace(), qirMagicArgs.CapabilityInsuficient);
```
xinyi-joffre, line 288:
```suggestion
            evt.SetProperty("CapabilityInsufficient".WithTelemetryNamespace(), qirMagicArgs.CapabilityInsufficient);
```

src/Tool/Telemetry.cs
```diff
@@ -272,6 +281,33 @@ public static EventProperties SetCommonProperties(this EventProperties evt)
...
+            evt.SetProperty("QirStreamSize".WithTelemetryNamespace(), qirMagicArgs.QirStreamSize?.ToString());
+            evt.SetProperty("Capability".WithTelemetryNamespace(), qirMagicArgs.Capability);
+            evt.SetProperty("CapabilityInsuficient".WithTelemetryNamespace(), qirMagicArgs.CapabilityInsuficient);
```
xinyi-joffre, line 291:
```suggestion
            evt.SetProperty("CapabilityInsufficient".WithTelemetryNamespace(), qirMagicArgs.CapabilityInsufficient);
```

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -252,11 +292,27 @@ public Task<IQuantumMachineJob> SubmitAsync(IQirSubmitter submitter, AzureSubmis
 
             var entryPointInput = GetEntryPointInputArguments(submissionContext, allowOptionalParameters);
             Logger?.LogInformation("Submitting job {FriendlyName} with {NShots} shots.", submissionContext.FriendlyName, submissionContext.Shots);
-            var options = SubmissionOptions.Default.With(submissionContext.FriendlyName, submissionContext.Shots, submissionContext.InputParams);
+            var options = SubmissionOptions.Default.With(
+                friendlyName: submissionContext.FriendlyName, 
+                shots: submissionContext.Shots,
+                inputParams: submissionContext.InputParams,
```
xinyi-joffre, line 298: Just for my understanding, inputParams here does not equal inputParams on the job object right? Is this technically for arguments?

Or is this a grab-bag of other params that will get merged with friendlyName, shots, and targetCapability in a later submit command?

xinyi-joffre (approved):

src/AzureClient/EntryPoint/EntryPoint.cs
vxfield: It's a `ImmutableDictionary<string, string>` grab-bag.

From the docs:
> Provider-specific job parameters to be passed to the execution target, expressed as one or more JSON {"key":"value",...} pairs.
> These parameters apply to <c>%azure.execute</c> and <c>%azure.submit</c>. The JSON may not contain separating spaces when used in a Jupyter notebook. 
> NOTE: This property was named "InputParams" (instead of "JobParams") because the closely
   related implementation in microsoft/qsharp-runtime used the name "InputParams"
   (see https://github.com/microsoft/qsharp-runtime/pull/829).
> Please notice the difference between "InputParams" and the preexisting "InputParameters" member below.

cesarzc: `options.InputParams` is related to `inputParams` on the job object but it is adding to the job object (rather than setting `inputParams`).

src/AzureClient/EntryPoint/EntryPoint.cs
```diff
@@ -4,14 +4,26 @@
...
 using System.Threading;
 using Microsoft.Extensions.Logging;
+using Microsoft.Quantum.QsCompiler;
 using Microsoft.Quantum.Runtime;
 using Microsoft.Quantum.Runtime.Submitters;
 using Microsoft.Quantum.Simulation.Core;
 
 namespace Microsoft.Quantum.IQSharp.AzureClient
 {
+    public record EntryPointSubmitArgs
```
cesarzc, line 17: nit: the name can be confused with something that is actually part of the job itself rather than the payload of a telemetry event. Maybe it could be named something like `EntryPointSubmitEventData` or `EntryPointSubmitEventPayload`.

cesarzc:
@vxfield based on the spec review we had earlier in the day, I think the changes that introduce `_repr_qir_()` should not be part of this PR. Since introducing this new function would not result in new functionality, I think it would be better to wait to introduce it (since how we generate QIR for Q# might change considerably). Additionally, if it is introduced, it is significant enough to be its own PR.

src/Kernel/Magic/QirMagic.cs
```diff
@@ -151,76 +225,184 @@ private static unsafe bool TryParseBitcode(string path, out LLVMModuleRef outMod
...
         {
-            return "Internal error: generated entry point does not contain a QIR bitcode stream, but no compilation errors were returned."
-                .ToExecutionResult(ExecuteStatus.Error);
+            this.EventService?.Trigger<QirMagicEvent, QirMagicEventData>(qirMagicArgs);
         }
+    }
 
-        var bitcodeFile = Path.ChangeExtension(Path.GetTempFileName(), ".bc");
-        using (var outStream = File.OpenWrite(bitcodeFile))
+    private ExecutionResult ReturnQIR_IR(IChannel channel, string? outputFilePath, QirOutputFormat outputFormat, Stream qirStream)
```
cesarzc, line 320: It seems like the `outputFormat` argument in these methods is not really needed. It is only used in the `string` for the `LlvmIr` constructor and should be a constant string rather than something that comes from the method arguments.

src/Kernel/Magic/QirMagic.cs
```diff
@@ -56,21 +98,51 @@ public class QirMagic : AbstractMagic
...
+                Defaults to the active Azure Quantum target (which can be set with `%azure.target`).                
+                Otherwise, defaults to a generic target, which may not work when running on a specific target.
+
+                - `{ParameterNameTargetCapability}=<string>`: The capability of the intended execution target.
+                If `{ParameterNameTarget}` is specified or there is an active Azure Quantum target,
+                defaults to the target's maximum capability.
+                Otherwise, defaults to `FullComputation`, which may not be supported when running on a specific target.
+                Possible options are:
+                    * `{TargetCapabilityModule.Top}`
+                    * `{TargetCapabilityModule.Bottom}`
```
cesarzc, line 130: Is this surfaced to the user? What is the intent of `TargetCapabilityModule.Top` and `TargetCapabilityModule.Bottom`.

src/Python/qsharp-core/qsharp/ipython_magic.py
```diff
@@ -24,7 +24,7 @@ def qsharp(magic_args, cell, local_ns=None):
         callables = qs.compile(cell)
         if isinstance(callables, qs.QSharpCallable):
             local_ns[callables._name] = callables
-        else:
+        elif callables:
```
cesarzc, line 27: This is slight change in behavior. 

Before, if `callables` was `None` an error would probably be raised when trying to iterate over it. Now, nothing would happen if `callables` is `None`. The user might not know what is going command silently does nothing.

src/Python/qsharp-core/qsharp/loader.py
```diff
@@ -122,13 +120,77 @@ def simulate_noise(self, **kwargs) -> Any:
         """
         return qsharp.client.simulate_noise(self, **kwargs)
 
-    def as_qir(self) -> bytes:
+    def as_qir(self, metadata: Optional[Dict[str, str]] = None, **kwargs) -> str:
```
cesarzc, line 123: I think it would be good to keep this simple and just get `target` and `target_capability` from `kwargs` instead of also having a `metadata` argument.

src/Python/qsharp-core/qsharp/loader.py
```diff
@@ -122,13 +120,77 @@ def simulate_noise(self, **kwargs) -> Any:
...
+        data = qsharp.client.compile_to_qir(self,
+                                            target=target,
+                                            target_capability=target_capability,
+                                            **kwargs)
+        
+        if "Text" in data:
+            return data['Text']
+        raise IQSharpError([f'Error in generating QIR. {data}'])
+
+    def _repr_qir_(self, metadata: Optional[Dict[str, str]] = None, **kwargs: Any) -> bytes:
```
cesarzc, line 163: Will this be the API used by the `azure-quantum` module for submission?

cesarzc: As in the other comment, I think it would be good to keep this simple and just get `target` and `target_capability` from `kwargs` instead of also having a `metadata` argument.

src/Kernel/Magic/QirMagic.cs
vxfield: As discussed, I removed the list of possible `TargetCapability` from the comments, and moved the comments about the new non-user facing parameters to the developer documentation instead of the `%qir` magic command documentation.

src/Python/qsharp-core/qsharp/ipython_magic.py
vxfield: As discussed, this was actually a bug, and it does not change the behavior of the `%%qsharp` magic when compiling Q# code inline a Python notebook. But it fixes a `null reference` bug when calling another IQ# magic command that does not return a callable like
```
%%qsharp
%version
```

src/Python/qsharp-core/qsharp/loader.py
vxfield: As discussed, I removed the `metadata` parameter in favor of just the `kwargs`.

src/Python/qsharp-core/qsharp/loader.py
vxfield: > Will this be the API used by the azure-quantum module for submission?
Yes.

As discussed, I removed the `metadata` parameter in favor of just the `kwargs`.

